PR #7, opened 2019-02-14
Add ironic-inspector configuration to container

juliakreger:
In order for ironic to trigger ironic-inspector directly
in a workflow where a machine may be manually added and not
discovered, we need configuration in ironic.conf to both enable
the ``inspector`` interface, as well as point ironic to inspector
to enable it's internal workflow for data collection... outside
of discovery operations.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,8 @@ RUN cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig && \
     crudini --set /etc/ironic/ironic.conf DEFAULT enabled_boot_interfaces pxe,ipxe && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_boot_interface ipxe && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_deploy_interface direct && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT enabled_inspect_interfaces inspector && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT default_inspect_interface inspector && \
     crudini --set /etc/ironic/ironic.conf database connection sqlite:///ironic.db && \
     crudini --set /etc/ironic/ironic.conf dhcp dhcp_provider none && \
     crudini --set /etc/ironic/ironic.conf conductor automated_clean false && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -19,13 +19,18 @@ RUN cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig && \
...
+    crudini --set /etc/ironic/ironic.conf DEFAULT enabled_inspect_interfaces inspector && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT default_inspect_interface inspector && \
     crudini --set /etc/ironic/ironic.conf database connection sqlite:///ironic.db && \
     crudini --set /etc/ironic/ironic.conf dhcp dhcp_provider none && \
     crudini --set /etc/ironic/ironic.conf conductor automated_clean false && \
     crudini --set /etc/ironic/ironic.conf conductor api_url http://172.22.0.1:6385 && \
     crudini --set /etc/ironic/ironic.conf deploy http_url http://172.22.0.1 && \
     crudini --set /etc/ironic/ironic.conf deploy http_root /var/www/html/ && \
     crudini --set /etc/ironic/ironic.conf deploy default_boot_option local && \
+    crudini --set /etc/ironic/ironic.conf inspector enabled True && \
```
bfournie: From the .conf this field is deprecated so probably don't need to set

Dockerfile
```diff
@@ -19,13 +19,18 @@ RUN cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig && \
...
+    crudini --set /etc/ironic/ironic.conf inspector auth_type none && \
```
bfournie: auth_type none is the default

juliakreger:
@bfournie Fixed!

Dockerfile
```diff
@@ -19,13 +19,16 @@ RUN cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig && \
     crudini --set /etc/ironic/ironic.conf DEFAULT enabled_boot_interfaces pxe,ipxe && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_boot_interface ipxe && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_deploy_interface direct && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT enabled_inspect_interfaces inspector && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT default_inspect_interface inspector && \
```
dtantsur: Not really needed as long as "enabled" is set

Dockerfile
```diff
@@ -19,13 +19,16 @@ RUN cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig && \
     crudini --set /etc/ironic/ironic.conf DEFAULT enabled_boot_interfaces pxe,ipxe && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_boot_interface ipxe && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_deploy_interface direct && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT enabled_inspect_interfaces inspector && \
```
dtantsur: Strictly speaking, we should keep `no-inspect` in the list. However, I don't think we have any drivers that do not support `inspector`, so it's fine for now.

PR #8, opened 2019-02-18
Switch to JSON RPC instead of Rabbit MQ

dtantsur:
Disclaimer: I haven't tested it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM docker.io/centos:centos7
 
 RUN yum install -y python-requests && \
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - current && \
-    yum install -y openstack-ironic-api openstack-ironic-conductor rabbitmq-server crudini iproute dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs && \
+    yum install -y openstack-ironic-api openstack-ironic-conductor crudini iproute dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs && \
     yum clean all
 
 RUN mkdir /tftpboot && \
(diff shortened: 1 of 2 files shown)


derekhiggins:
Do we need to disable rabbit somewhere? Ironic seems to be trying to connect to it.
2019-02-18 12:02:21.682 8 ERROR oslo.messaging._drivers.impl_rabbit [req-05b55e3e-e83a-43b3-ac53-15497a01ef3e - - - - -] Connection failed: [Errno 111] ECONNREFUSED (retrying in 22.0 seconds): error: [Errno 111] ECONNREFUSED

dtantsur:
@derekhiggins what's the ironic RPM version you end up with?

dtantsur:
@derekhiggins updated to use `current`, should have the right package now.

Dockerfile
```diff
@@ -26,6 +27,7 @@ RUN cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig && \
     crudini --set /etc/ironic/ironic.conf deploy http_root /var/www/html/ && \
     crudini --set /etc/ironic/ironic.conf deploy default_boot_option local && \
     crudini --set /etc/ironic/ironic.conf inspector endpoint_override http://172.22.0.1:5050 && \
+    crudini --set /etc/ironic/ironic.conf json_rpc host_ip 172.22.0.1 && \
```
derekhiggins: Looks like we may need to update this, or change the host we connect to
`   ConnectFailure: Unable to establish connection to http://host11.beaker.tripleo.lab.eng.rdu2.redhat.com:8089: HTTPConnectionPool(host='host11.beaker.tripleo.lab.eng.rdu2.redhat.com', port=8089): Max retries
`
I changed the host_ip to 0.0.0.0 and the patch worked for me, but maybe it would be better to change so that we connect to the 172 address

dtantsur: It will connect to whatever host name the conductor has. Can you check what hostname is used? If it's something `localhost`, then we can probably use `127.0.0.1` (could you try?).

dtantsur: Oh, well, I'm apparently blind. So, `host11.beaker.tripleo.lab.eng.rdu2.redhat.com`. Okay, let's keep it 0.0.0.0 for now, I'll update.

dtantsur:
@derekhiggins updated with `0.0.0.0`.

PR #9, opened 2019-02-26
Add runhealthcheck.sh

derekhiggins:
Add a script that runs inside the container monitoring
the processes that need to be available, if any of them fail then
kill the container. We can then more reliably check the container
status to see if its running or not.

### DIFF
--- a/runhealthcheck.sh
+++ b/runhealthcheck.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+set -ex
+
+trap killcontainer ERR
+function killcontainer(){
+    killall -9 sleep
+}
+
+while true ; do
+    sleep 10
+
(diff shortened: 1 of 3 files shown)


dtantsur:
I don't see `runhealthcheck.sh`, maybe you forgot to `git add` it?

hardys:
lgtm, feel free to merge if @dtantsur agrees :cake:

runhealthcheck.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -ex
+
+trap killcontainer ERR
+function killcontainer(){
+    killall -9 sleep
```
dtantsur: Looks secure :) Maybe we need a PID file?

PR #10, opened 2019-03-04
Switch to using WAL for ironic sqlite database

dtantsur:
The default mode does not work well when several threads and/or processes
are reading and writing to the same database. This change switches
to the WAL mode available since sqlite 3.7.0.

The mode is persistent for a file until changed, so we pre-create
the database file when building a container.


dtantsur:
I think we should (it's a separate repo, right?).

bfournie:
> I think we should (it's a separate repo, right?).

Yes - https://github.com/metalkube/metalkube-ironic-inspector.  I will try it.

PR #11, opened 2019-03-13
Enable idrac interfaces

derekhiggins:
Allow us to use idrac interfaces on Dell HW.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,9 +17,13 @@ RUN cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig && \
     crudini --set /etc/ironic/ironic.conf DEFAULT debug true && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_network_interface noop && \
     crudini --set /etc/ironic/ironic.conf DEFAULT enabled_boot_interfaces pxe,ipxe && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT enabled_power_interfaces ipmitool,idrac && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT enabled_management_interfaces ipmitool,idrac && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT enabled_hardware_types ipmi,idrac && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT enabled_vendor_interfaces ipmitool,no-vendor,idrac && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_boot_interface ipxe && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_deploy_interface direct && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -17,6 +17,9 @@ RUN cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig && \
     crudini --set /etc/ironic/ironic.conf DEFAULT debug true && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_network_interface noop && \
     crudini --set /etc/ironic/ironic.conf DEFAULT enabled_boot_interfaces pxe,ipxe && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT enabled_power_interfaces ipmitool,idrac && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT enabled_management_interfaces ipmitool,idrac && \
+    crudini --set /etc/ironic/ironic.conf DEFAULT enabled_hardware_types ipmi,idrac && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_boot_interface ipxe && \
     crudini --set /etc/ironic/ironic.conf DEFAULT default_deploy_interface direct && \
     crudini --set /etc/ironic/ironic.conf DEFAULT enabled_inspect_interfaces inspector && \
```
dtantsur: You may want to enable their OOB inspection as well.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/199/

Dockerfile
derekhiggins: Looks like the CI didn't test this, I'll update it once the CI fix merges
https://github.com/metalkube/dev-scripts/pull/156

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/206/

juliakreger:
I'm curious why, since idrac lacks a sensor data collection.... I mean in the long term it makes sense, just curious if we're trying to route around something or if we're encountering an issue with ipmi.

derekhiggins:
> I'm curious why, since idrac lacks a sensor data collection.... I mean in the long term it makes sense, just curious if we're trying to route around something or if we're encountering an issue with ipmi.

These hosts I'm working on have a 224G Boot Optimised Storage Solution
(BOSS /dev/sdi), which I guess is where we should be installing the OS, using
the ipmi driver I can't for the life of me get the nodes to boot from this drive, 
every time ironic sets the bootdev to DISK the hard drive order I set in the
bios resets back to the integrated controller and attempts to boot from /dev/sda,
if I switch ironic to us the drac driver it works and I can use the
BOSS drive.

juliakreger:
Ugh, I think I had to define the root for the boss card. Anyway, the raid controller/disk boot is being changed for the next hardware :(

PR #14, opened 2019-03-15
Configure httpd to listen on 5280 instead of 80.

imain:
I just picked the port at random.  Kubernetes definitely has things running on port 80 so we gotta move this. 
 Hopefully this is all that needs changing.  Let me know! :)


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/257/

dtantsur:
We'll also have to update https://github.com/openshift-metalkube/dev-scripts/blob/master/ironic/dnsmasq.conf#L10. And no, I don't know why it's in dev-scripts rather than here :)

dtantsur:
Got it, let's rebase it on top of #15.

imain:
Do you mean apply 15 on top of this one?  That is what I would expect..

dtantsur:
No, the other way around. Otherwise this change is breaking because port 80 is still used in dnsmasq configuration.

imain:
Bob is going to take this on.  Thanks again Bob!!

PR #15, opened 2019-03-15
Add self configuration to Ironic for Openshift Baremetal Operator.

imain:
We're planning to run these containers along side the BMO deployment.
This means we need to configure Ironic at startup, along with any
images needed etc.  We still need the dualboot image in here but
we are much more limited running in a kubernetes pod than with podman
on the local host.

### DIFF
--- a/config/configure_ironic.sh
+++ b/config/configure_ironic.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+set -xe
+
+# Configure dnsmasq and pxe boot parameters.
+
+IRONIC_IP=${IRONIC_IP:-"172.22.0.1"}
+IRONIC_DHCP_RANGE=${IRONIC_DHCP_RANGE:-"172.22.0.10,172.22.0.100"}
+
+cat /config/dnsmasq.conf | sed -e s/IRONIC_IP/$IRONIC_IP/g -e s/IRONIC_DHCP_RANGE/$IRONIC_DHCP_RANGE/g > /etc/dnsmasq.conf
+cat /config/dualboot.ipxe | sed s/IRONIC_IP/$IRONIC_IP/g > /var/www/html/dualboot.ipxe
+cat /config/inspector.ipxe | sed s/IRONIC_IP/$IRONIC_IP/g > /var/www/html/inspector.ipxe
+
+# Add firewall rules to ensure the IPA ramdisk can reach Ironic and Inspector APIs on the host
+for port in 5050 6385 ; do
+    if ! iptables -C INPUT -i provisioning -p tcp -m tcp --dport $port -j ACCEPT > /dev/null 2>&1; then
+        iptables -I INPUT -i provisioning -p tcp -m tcp --dport $port -j ACCEPT
+    fi
+done
+
+# Get the images we need to serve.
+mkdir -p /var/www/html/images
+pushd /var/www/html/images
+
+export RHCOS_IMAGE_URL=${RHCOS_IMAGE_URL:-"https://releases-rhcos.svc.ci.openshift.org/storage/releases/maipo/"}
+export RHCOS_IMAGE_VERSION="${RHCOS_IMAGE_VERSION:-47.284}"
+export RHCOS_IMAGE_NAME="redhat-coreos-maipo-${RHCOS_IMAGE_VERSION}"
+export RHCOS_IMAGE_FILENAME="${RHCOS_IMAGE_NAME}-qemu.qcow2"
+export RHCOS_IMAGE_FILENAME_OPENSTACK="${RHCOS_IMAGE_NAME}-openstack.qcow2"
+
+# This is the one we actually want to use, but I'm not going to assemble it here
+# in a kubernetes managed container.
+export RHCOS_IMAGE_FILENAME_DUALDHCP="${RHCOS_IMAGE_NAME}-dualdhcp.qcow2"
+export RHCOS_IMAGE_FILENAME_LATEST="redhat-coreos-maipo-latest.qcow2"
+
+curl --insecure --compressed -L -o "${RHCOS_IMAGE_FILENAME_OPENSTACK}" "${RHCOS_IMAGE_URL}/${RHCOS_IMAGE_VERSION}/${RHCOS_IMAGE_FILENAME_OPENSTACK}".gz
+curl --insecure --compressed -L https://images.rdoproject.org/master/rdo_trunk/current-tripleo/ironic-python-agent.tar | tar -xf -
+
+if [ ! -e "$RHCOS_IMAGE_FILENAME_OPENSTACK.md5sum" -o \
+     "$RHCOS_IMAGE_FILENAME_OPENSTACK" -nt "$RHCOS_IMAGE_FILENAME_OPENSTACK.md5sum" ] ; then
+    md5sum "$RHCOS_IMAGE_FILENAME_OPENSTACK" | cut -f 1 -d " " > "$RHCOS_IMAGE_FILENAME_OPENSTACK.md5sum"
+fi
+
+ln -sf "$RHCOS_IMAGE_FILENAME_OPENSTACK" "$RHCOS_IMAGE_FILENAME_LATEST"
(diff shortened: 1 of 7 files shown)


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/258/

runironic.sh
```diff
@@ -1,8 +1,15 @@
 #!/usr/bin/bash
+
+set -x
+
+# Optionally configure ironic if the environment variable is set.
+if [ ! -e "$CONFIGURE_IRONIC" ] ; then
```
bfournie: Hopefully we can eventually not need this and use the same config for both the install and BMO Ironic.  Since https://github.com/metalkube/metalkube-ironic/pull/14 sets the port to 5280 without a conditional this would cause a problem if not configuring Ironic here.

configure_ironic.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+set -xe
+
+# Configure dnsmasq and pxe boot parameters.
+
+IRONIC_IP=${IRONIC_IP:-"172.22.0.1"}
+IRONIC_DHCP_RANGE=${IRONIC_DHCP_RANGE:-"172.22.0.10,172.22.0.100"}
+
```
bfournie: Would be useful to also set IRONIC_PORT to 5280 and use a variable in dnsmasq.conf and inspector.ipxe.

config/dnsmasq.conf
```diff
@@ -0,0 +1,24 @@
+interface=provisioning
+except-interface=lo
+bind-dynamic
+dhcp-range=IRONIC_DHCP_RANGE
+enable-tftp
+tftp-root=/tftpboot
+
+dhcp-match=ipxe,175
+# Client is already running iPXE; move to next stage of chainloading
+dhcp-boot=tag:ipxe,http://IRONIC_IP:5280/dualboot.ipxe
```
dtantsur: The `5280` change is from another PR, let's maybe keep it there?

configure_ironic.sh
dtantsur: I'd prefer to separate the port change completely from this one.

configure_ironic.sh
```diff
@@ -0,0 +1,48 @@
...
+cat /config/dnsmasq.conf | sed -e s/IRONIC_IP/$IRONIC_IP/g -e s/IRONIC_DHCP_RANGE/$IRONIC_DHCP_RANGE/g > /etc/dnsmasq.conf
+cat /config/dualboot.ipxe | sed s/IRONIC_IP/$IRONIC_IP/g > /var/www/html/dualboot.ipxe
```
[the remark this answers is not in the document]
dtantsur: Correct.

configure_ironic.sh
```diff
@@ -0,0 +1,48 @@
...
+cat /config/inspector.ipxe | sed s/IRONIC_IP/$IRONIC_IP/g > /var/www/html/inspector.ipxe
+
+# Add firewall rules to ensure the IPA ramdisk can reach Ironic and Inspector APIs on the host
+# I suspect for kubernetes we can just use hostPort with hostNetwork.
+for port in 5050 6385 ; do
```
dtantsur: Note: add `5280` when you switch to it.

configure_ironic.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Get the images we need to serve.
+mkdir -p /var/www/html/images
+pushd /var/www/html/images
+
+export RHCOS_IMAGE_URL=${RHCOS_IMAGE_URL:-"https://releases-rhcos.svc.ci.openshift.org/storage/releases/maipo/"}
+export RHCOS_IMAGE_VERSION="${RHCOS_IMAGE_VERSION:-47.284}"
+export RHCOS_IMAGE_NAME="redhat-coreos-maipo-${RHCOS_IMAGE_VERSION}"
+export RHCOS_IMAGE_FILENAME="${RHCOS_IMAGE_NAME}-qemu.qcow2"
+export RHCOS_IMAGE_FILENAME_OPENSTACK="${RHCOS_IMAGE_NAME}-openstack.qcow2"
```
dtantsur: Could you move these defines to a common shell file that gets sourced in all places where you use it?

dtantsur: Actually, why do you have two files `configure_ironic` in different locations?

runironic.sh
dtantsur: I agree, I think we should configure ironic the same way in dev-scripts and on k8s. Meaning, the dev-scripts part should be removed.

config/configure_ironic.sh
```diff
@@ -0,0 +1,47 @@
...
+export RHCOS_IMAGE_NAME="redhat-coreos-maipo-${RHCOS_IMAGE_VERSION}"
+export RHCOS_IMAGE_FILENAME="${RHCOS_IMAGE_NAME}-qemu.qcow2"
+export RHCOS_IMAGE_FILENAME_OPENSTACK="${RHCOS_IMAGE_NAME}-openstack.qcow2"
+
+# This is the one we actually want to use, but I'm not going to assemble it here
+# in a kubernetes managed container.
+export RHCOS_IMAGE_FILENAME_DUALDHCP="${RHCOS_IMAGE_NAME}-dualdhcp.qcow2"
+export RHCOS_IMAGE_FILENAME_LATEST="redhat-coreos-maipo-latest.qcow2"
+
+curl --insecure --compressed -L -o "${RHCOS_IMAGE_FILENAME_OPENSTACK}" "${RHCOS_IMAGE_URL}/${RHCOS_IMAGE_VERSION}/${RHCOS_IMAGE_FILENAME_OPENSTACK}".gz
```
hardys: We should keep the option of bind mounting in images from the host, otherwise we'll be downloading large images every test run for testing/CI etc - check 04_setup_ironic.sh in dev-scripts where we bind `-v $IRONIC_DATA_DIR/html/images:/var/www/html/images`

Is there any way we can enable similar caching for the k8s hosted case?

hardys: I guess a simple way would be to expose the images on the test host via a webserver then ensure all the image URLs can be overridden?

imain: I think for now I'm going to refer to the httpd running in the pod on the host rather than in the openshift cluster.  That will make this all much more palatable.

config/dnsmasq.conf
imain: basically this would depend on that change.  This file doesn't exist whereas the other PR updates the existing setup to 5280.  I think it'd be best to test apply that first then appy this one.  Also we can't use port 80 in kubernetes so this is a hard requirement.

config/configure_ironic.sh
```diff
@@ -0,0 +1,47 @@
...
+
+IRONIC_IP=${IRONIC_IP:-"172.22.0.1"}
+IRONIC_DHCP_RANGE=${IRONIC_DHCP_RANGE:-"172.22.0.10,172.22.0.100"}
+
+cat /config/dnsmasq.conf | sed -e s/IRONIC_IP/$IRONIC_IP/g -e s/IRONIC_DHCP_RANGE/$IRONIC_DHCP_RANGE/g > /etc/dnsmasq.conf
+cat /config/dualboot.ipxe | sed s/IRONIC_IP/$IRONIC_IP/g > /var/www/html/dualboot.ipxe
+cat /config/inspector.ipxe | sed s/IRONIC_IP/$IRONIC_IP/g > /var/www/html/inspector.ipxe
+
+# Add firewall rules to ensure the IPA ramdisk can reach Ironic and Inspector APIs on the host
+for port in 5050 6385 ; do
```
derekhiggins: Do we need to add the httpd port to this list?

configure_ironic.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
derekhiggins: We have 2 of these files in this patch
./configure_ironic.sh and ./config/configure_ironic.sh

imain: aha!  sorry about that.  I think we should land the 'split containers' patch first and then I'll rework this.

imain:
Thanks guys!

PR #16, opened 2019-03-22
Changes to split up Ironic container

bfournie:
Allows Ironic container to be split between ironic, dnsmasq, and httpd
containers using this same image.

### DIFF
--- a/runhealthcheck.sh
+++ b/runhealthcheck.sh
@@ -7,14 +7,18 @@ function killcontainer(){
 }
 
 while true ; do
-    sleep 10
+    sleep 30
 
-    HTTPDPID=$(pidof -s httpd)
-    fuser 80/tcp |& grep -w "$HTTPDPID"
+    if [ $1 = "httpd" ] ; then
+       HTTPDPID=$(pidof -s httpd)
+       fuser 80/tcp |& grep -w "$HTTPDPID"
 
-    DNSMASQPID=$(pidof dnsmasq)
-    fuser 69/udp |& grep -w "$DNSMASQPID"
+    elif [ $1 = "dnsmasq" ] ; then
+       DNSMASQPID=$(pidof dnsmasq)
+       fuser 67/udp |& grep -w "$DNSMASQPID"
 
-    curl -s http://172.22.0.1:6385 > /dev/null || ( echo "Can't contact ironic-api" && exit 1 )
+    elif [ $1 = "ironic" ] ; then
+       curl -s http://localhost:6385 > /dev/null || ( echo "Can't contact ironic-api" && exit 1 )
+    fi
 
 done
--- a/runhttpd.sh
+++ b/runhttpd.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/bash
+
+rmdir /var/www/html
+ln -s /shared/html/ /var/www/html
+
+/usr/sbin/httpd &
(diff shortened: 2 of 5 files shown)


runhealthcheck.sh
```diff
@@ -9,12 +9,16 @@ function killcontainer(){
...
 
-    DNSMASQPID=$(pidof dnsmasq)
-    fuser 69/udp |& grep -w "$DNSMASQPID"
+    elif [ $1 = "dnsmasq" ] ; then
+       DNSMASQPID=$(pidof dnsmasq)
+       fuser 67/udp |& grep -w "$DNSMASQPID"
 
-    curl -s http://172.22.0.1:6385 > /dev/null || ( echo "Can't contact ironic-api" && exit 1 )
+    elif [ $1 = "ironic" ] ; then
+       curl -s http://172.22.0.1:6385 > /dev/null || ( echo "Can't contact ironic-api" && exit 1 )
```
juliakreger: Is there any way we can just make this such it looks at the IP in the container?

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/348/

runhealthcheck.sh
bfournie: Changed to use localhost for just testing response on this port.

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/354/

runhttpd.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/bash
+
+rmdir /var/www/html
+ln -s /shared/html/ /var/www/html
```
dtantsur: nit: this could probably go to the httpd configuration

runironic.sh
```diff
@@ -1,8 +1,6 @@
 #!/usr/bin/bash
 /usr/bin/python2 /usr/bin/ironic-conductor > /var/log/ironic-conductor.out 2>&1 &
 /usr/bin/python2 /usr/bin/ironic-api > /var/log/ironic-api.out 2>&1 &
```
dtantsur: We need a way to split this out. Or even better, just run it under httpd as a wsgi application (which is anyway the recommended way to run ironic: https://docs.openstack.org/ironic/latest/install/install-rdo.html#configuring-ironic-api-behind-mod-wsgi)

runhttpd.sh
bfournie: Good point, I'll change this in follow-on patch that will change the http port we're using.

runironic.sh
bfournie: Yeah I think using wsgi makes sense which is how we do it with tripleo, and splitting it out will allow one set of logs from this container.  OK to do in a follow-on patch with above changes?

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/371/

hardys:
Doesn't appear to be working - the http request fails downloading the image with Failed connect to 172.22.0.1:80; Connection refused ?

bfournie:
> Doesn't appear to be working - the http request fails downloading the image with Failed connect to 172.22.0.1:80; Connection refused ?

Did you include the dev-scripts changes at the same time (https://github.com/openshift-metalkube/dev-scripts/pull/204/)?  The dev-scripts change is needed to start the httpd container, which sounds like its not happening.

Issue #13, opened 2019-03-15
Ironic container is not very kubernetes friendly


imain:
Hey guys!

So in the process of trying to use this in a pod with the baremetal operator, I noticed a few things which should probably be changed, or at least talked about, with respect to running this container in kubernetes.  My intention here is just to record these so we can work with them.

- We can't bind mount anything in, so I've got a PR to download the images at startup.  Open to suggestions here.
- We shouldn't really be running 4 processes in one container.  It makes debugging much harder.
- We shouldn't be logging to /var/log/.  We can't access that from the host or anywhere else, and when the container dies its not possible to see the issues in the logs.  Ideally we would have 1 process per container logging to stdout so we can get logs from kubernetes.  Even just having the 4 processes log to stdout would be an improvement for debugging.
- httpd is listening on port 80?  hmm, tricky one to work with.  We have to run this as net=host so we can do dhcp/pxe so this interferes with any service on the host (currently hitting this in testing with minishift).
- The healthcheck should be implemented using the kubernetes/docker/podman healthcheck function.

Sorry to just point out a bunch of stuff.. like I say just trying to get it all summarised.

dtantsur:
Hi Ian, great points!

> We can't bind mount anything in, so I've got a PR to download the images at startup.

Is there any k8s-friendly way to push blobs to a container? If not, downloading may be fine.

> We shouldn't really be running 4 processes in one container.

I fully agree. But it was an explicit design request. I'm totally open to going back to how TripleO/Kolla splits the containers. This will need switching to MySQL, but that's on our roadmap anyway. Also note that some containers will need a shared volume (the same way TripleO does it).

>  Ideally we would have 1 process per container logging to stdout so we can get logs from kubernetes.

Yeah, I think using `/var/log` is a side-effect from having 1 container. Also note that the current containers are more or less hacked together to Just Work.

> httpd is listening on port 80?

We can change to anything.

imain:
> Hi Ian, great points!
> 
> > We can't bind mount anything in, so I've got a PR to download the images at startup.
> 
> Is there any k8s-friendly way to push blobs to a container? If not, downloading may be fine.

We may be able to use a volume in the future, but I think our devel setup atm does not have any storage options to back the volumes.  I think for now downloading make sense.

> > We shouldn't really be running 4 processes in one container.
> 
[...]

PR #18, opened 2019-03-26
Change out sqlite to mariadb

juliakreger:
Turns out we need a real database for the level of parallel
access. Changes out the database to an instance of mariadb
that is configured upon start-up if necessary with a random
password and unrelated database rights are removed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,14 +2,11 @@ FROM docker.io/centos:centos7
 
 RUN yum install -y python-requests && \
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/5609d2e3aee35578e914bcbfac92a46c759c0a31/tripleo_repos/main.py | python - current && \
-    yum install -y openstack-ironic-api openstack-ironic-conductor crudini iproute dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs psmisc sysvinit-tools && \
+    yum install -y openstack-ironic-api openstack-ironic-conductor crudini iproute dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs psmisc sysvinit-tools mariadb-server python-PyMySQL python2-chardet && \
     yum clean all
 
 RUN mkdir /tftpboot && \
     cp /usr/share/ipxe/undionly.kpxe /usr/share/ipxe/ipxe.efi /tftpboot/
-
-RUN mkdir -p /var/lib/ironic && \
-    sqlite3 /var/lib/ironic/ironic.db "pragma journal_mode=wal"
 
 RUN cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig && \
     crudini --set /etc/ironic/ironic.conf DEFAULT auth_strategy noauth && \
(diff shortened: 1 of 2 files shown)


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/375/

runironic.sh
```diff
@@ -1,4 +1,35 @@
...
+DATADIR="/var/lib/mysql"
+if [ ! -f "$DATADIR/ironic-initalized" ]; then
+    crudini --set /etc/my.conf mysqld max_connects 64
+    crudini --set /etc/my.conf mysqld max_heap_table_size 1M
+    crudini --set /etc/my.conf mysqld innodb_buffer_pool_size 5M
+    crudini --set /etc/my.conf mysqld innodb_log_buffer_size 512K
+    mysql_install_db --datadir=$DATADIR
+    chown -R mysql /var/log/mariadb
+    chown -R mysql $DATADIR
+    cd /usr
```
yprokule: Why do we need this?

juliakreger: 1) setup initial db otherwise it does not start
2) Without tyring to interactively execute mysql_secure_installation, we need to ensure that mysql is the owner of the logs and data directory as line 9 was executed as root.
3) The change of the working directory is from the manual instructions on how to start mysql_safe.

juliakreger:
@dtantsur  dnsmasq and httpd are being moved out, and I think we should keep mysql with the two base services as otherwise we're going to have to do password orchestration and handling.

dtantsur:
You mean, have separate mariadb processes for ironic and ironic-inspector? How does it play with the request to have one service per container in #13?

runironic.sh
yprokule: >     3. The change of the working directory is from the manual instructions on how to start mysql_safe.

That's what I was wondering about. Thanks @juliakreger

juliakreger:
Well, if we split to one service per container, We're just going to have to pre-bake credentials in or use none... Which seems like a security issue to me.

juliakreger:
Bottom line, I was on a call where people expressed they were borderlined blocked on hitting the locking issues with sqllite, so if we take the approach for perfection at  the moment, we're going to continue to block their ability to deliver.

On a plus side, I just got my local environment up with the container built from this change and I do have a working ironic \o/

runironic.sh
```diff
@@ -1,4 +1,35 @@
...
+CREATE USER 'ironic'@'localhost' identified by '$ironic_password';
+GRANT ALL on *.* TO 'ironic'@'localhost' WITH GRANT OPTION;
+DROP DATABASE IF EXISTS test;
+CREATE DATABASE ironic;
+FLUSH PRIVILEGES;
+EOF
+    mysql -u root -p$ironic_password -h 127.0.0.1 < /tmp/configure-mysql.sql
+    ps auxf
+    crudini --set /etc/ironic/ironic.conf database connection mysql+pymysql://ironic:$ironic_password@localhost/ironic?charset=utf8
+    ironic-dbsync --config-file /etc/ironic/ironic.conf create_schema
```
dtantsur: nit: we can use the normal `upgrade` command with mysql and run it every time

juliakreger:
@dtantsur  I think we should only need this. The activity on inspector should be minimal.... should be.

dtantsur:
Should be :)

elfosardo:
since #16 has been merged, you can separate the db service and use a runmariadb.sh script instead of integrating all in runironic.sh
just not sure if we want to use the same image, but we can split them later if needed

Dockerfile
```diff
@@ -2,15 +2,12 @@ FROM docker.io/centos:centos7
 
 RUN yum install -y python-requests && \
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - current && \
-    yum install -y openstack-ironic-api openstack-ironic-conductor crudini iproute dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs psmisc sysvinit-tools && \
+    yum install -y openstack-ironic-api openstack-ironic-conductor crudini iproute dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs psmisc sysvinit-tools mariadb-server python-PyMySQL  python2-requests python2-chardet && \
```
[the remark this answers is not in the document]
juliakreger: So interesting thing about that, the stock reqeusts package gets used and I got compatability errors when testing the container build.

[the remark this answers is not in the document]
elfosardo: I guess we don't have to worry too much since python2-requests gets installed anyway as dependency and replaces python-requests

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/414/

derekhiggins:
merging, CI failed but it had gotten passed the ironic part so this patch should be ok.

PR #20, opened 2019-03-28
Pin tripleo-repos to avoid bug

hardys:
Until https://review.openstack.org/#/c/648358/ lands master
tripleo-repos is broken


hardys:
Ok fast-merging to unblock broken quay.io build

derekhiggins:
+1

PR #22, opened 2019-03-29
Provide overrides of Ironic provisioning IP, http port, and dnsmasq r…

bfournie:
…ange

Allow for overrides using environment variables of Ironic configuration.
The folowing environment variables are accepted:
  IP - provisioning interface IP address (default 172.22.0.1)
  HTTP_PORT - port used by http server (default 80)
  DHCP_RANGE - dhcp range to use for provisioning
       (default 172.22.0.10-172.22.0.100

### DIFF
--- a/dnsmasq.conf
+++ b/dnsmasq.conf
@@ -1,0 +1,24 @@
+interface=INTERFACE
+except-interface=lo
+bind-dynamic
+dhcp-range=DHCP_RANGE
+enable-tftp
+tftp-root=/shared/tftpboot
+
+dhcp-match=ipxe,175
+# Client is already running iPXE; move to next stage of chainloading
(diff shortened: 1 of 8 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/428/

dnsmasq.conf
```diff
@@ -0,0 +1,24 @@
+interface=provisioning
```
yprokule: Hi @bfournie interface _provisioning_ doesn't exist on master nodes, can it be made parameterized?
/cc @hardys @mcornea

dtantsur: I think it should, yes

bfournie: Good point, added INTERFACE env variable to allow configuration.

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/438/

PR #23, opened 2019-03-29
Split mariadb and ironic containers

elfosardo:
Splitting mariadb container from ironic
Relates to:
https://github.com/openshift-metalkube/dev-scripts/pull/241

### DIFF
--- a/runironic.sh
+++ b/runironic.sh
@@ -1,53 +1,53 @@
 #!/usr/bin/bash
 
-# mysql configuration
-PATH=$PATH:/usr/sbin/
-DATADIR="/var/lib/mysql"
-if [ ! -f "$DATADIR/ironic-initalized" ]; then
-    crudini --set /etc/my.conf mysqld max_connects 64
-    crudini --set /etc/my.conf mysqld max_heap_table_size 1M
-    crudini --set /etc/my.conf mysqld innodb_buffer_pool_size 5M
-    crudini --set /etc/my.conf mysqld innodb_log_buffer_size 512K
-    mysql_install_db --datadir=$DATADIR
-    chown -R mysql /var/log/mariadb
-    chown -R mysql $DATADIR
-    cd /usr
-    mysqld_safe --datadir=$DATADIR --user mysql &
-    sleep 1
-    ironic_password=$(echo $(date;hostname)|sha256sum |cut -c-20)
-    mysqladmin -u root password $ironic_password
-    cat > /tmp/configure-mysql.sql << EOF
-TRUNCATE mysql.user;
-CREATE USER 'ironic'@'localhost' identified by '$ironic_password';
-GRANT ALL on *.* TO 'ironic'@'localhost' WITH GRANT OPTION;
-DROP DATABASE IF EXISTS test;
-CREATE DATABASE ironic;
-FLUSH PRIVILEGES;
-EOF
-    mysql -u root -p$ironic_password -h 127.0.0.1 < /tmp/configure-mysql.sql
-    ps auxf
-    crudini --set /etc/ironic/ironic.conf database connection mysql+pymysql://ironic:$ironic_password@localhost/ironic?charset=utf8
-    ironic-dbsync --config-file /etc/ironic/ironic.conf create_schema
-    touch $DATADIR/ironic-initalized
-else
-    mysqld_safe --datadir=$DATADIR --user mysql &
-fi
-
 # Get environment settings and update ironic.conf
 IP=${IP:-"172.22.0.1"}
 HTTP_PORT=${HTTP_PORT:-"80"}
 INTERFACE=${INTERFACE:-"provisioning"}
-
-sed -i -e s/IRONIC_IP/$IP/g -e s/HTTP_PORT/$HTTP_PORT/g /etc/ironic/ironic.conf 
+MARIADB_PASSWORD=${MARIADB_PASSWORD:-"change_me"}
 
 # Allow access to Ironic
-if ! iptables -C INPUT -i $INTERFACE -p tcp -m tcp --dport 5050 -j ACCEPT > /dev/null 2>&1; then
-    iptables -I INPUT -i $INTERFACE -p tcp -m tcp --dport 5050 -j ACCEPT
+if ! iptables -C INPUT -i $INTERFACE -p tcp -m tcp --dport 6385 -j ACCEPT > /dev/null 2>&1; then
+    iptables -I INPUT -i $INTERFACE -p tcp -m tcp --dport 6385 -j ACCEPT
 fi
 
-/usr/bin/python2 /usr/bin/ironic-conductor > /var/log/ironic-conductor.out 2>&1 &
+cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig
 
+crudini --set /etc/ironic/ironic.conf DEFAULT auth_strategy noauth
+crudini --set /etc/ironic/ironic.conf DEFAULT my_ip ${IP}
+crudini --set /etc/ironic/ironic.conf DEFAULT debug true
+crudini --set /etc/ironic/ironic.conf DEFAULT default_network_interface noop
+crudini --set /etc/ironic/ironic.conf DEFAULT enabled_boot_interfaces pxe,ipxe
+crudini --set /etc/ironic/ironic.conf DEFAULT enabled_power_interfaces ipmitool,idrac
+crudini --set /etc/ironic/ironic.conf DEFAULT enabled_management_interfaces ipmitool,idrac
+crudini --set /etc/ironic/ironic.conf DEFAULT enabled_hardware_types ipmi,idrac
+crudini --set /etc/ironic/ironic.conf DEFAULT enabled_vendor_interfaces ipmitool,no-vendor,idrac
+crudini --set /etc/ironic/ironic.conf DEFAULT default_boot_interface ipxe
+crudini --set /etc/ironic/ironic.conf DEFAULT default_deploy_interface direct
+crudini --set /etc/ironic/ironic.conf DEFAULT enabled_inspect_interfaces inspector,idrac
+crudini --set /etc/ironic/ironic.conf DEFAULT default_inspect_interface inspector
+crudini --set /etc/ironic/ironic.conf DEFAULT rpc_transport json-rpc
+crudini --set /etc/ironic/ironic.conf dhcp dhcp_provider none
+crudini --set /etc/ironic/ironic.conf conductor automated_clean false
+crudini --set /etc/ironic/ironic.conf conductor api_url http://${IP}:6385
+crudini --set /etc/ironic/ironic.conf database connection mysql+pymysql://ironic:${MARIADB_PASSWORD}@localhost/ironic?charset=utf8
+crudini --set /etc/ironic/ironic.conf deploy http_url http://${IP}:${HTTP_PORT}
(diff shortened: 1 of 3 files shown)


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/432/

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/437/

runironic.sh
```diff
@@ -1,53 +1,53 @@
...
-    mysqld_safe --datadir=$DATADIR --user mysql &
-fi
-
 # Get environment settings and update ironic.conf
 IP=${IP:-"172.22.0.1"}
 HTTP_PORT=${HTTP_PORT:-"80"}
 INTERFACE=${INTERFACE:-"provisioning"}
-
-sed -i -e s/IRONIC_IP/$IP/g -e s/HTTP_PORT/$HTTP_PORT/g /etc/ironic/ironic.conf 
+MARIADB_PASSWORD=${MARIADB_PASSWORD:-"change_me"}
```
juliakreger: Lets do something to make this dynamic. I'd rather create it in a shared folder across container and just skip as we go.

runmariadb.sh
```diff
@@ -0,0 +1,27 @@
...
+crudini --set /etc/my.conf mysqld innodb_log_buffer_size 512K
+
+mysql_install_db --datadir="$DATADIR"
+
+chown -R mysql /var/log/mariadb
+chown -R mysql "$DATADIR"
+
+cat > /tmp/configure-mysql.sql <<-EOSQL
+    DELETE FROM mysql.user ;
+    CREATE USER 'ironic'@'localhost' identified by '${MARIADB_PASSWORD}' ;
```
juliakreger: MARIADB_PASSWORD is undefined in this script.

runmariadb.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/bash
+PATH=$PATH:/usr/sbin/
+DATADIR="/var/lib/mysql"
+
+crudini --set /etc/my.conf mysqld max_connects 64
+crudini --set /etc/my.conf mysqld max_heap_table_size 1M
+crudini --set /etc/my.conf mysqld innodb_buffer_pool_size 5M
+crudini --set /etc/my.conf mysqld innodb_log_buffer_size 512K
+
+mysql_install_db --datadir="$DATADIR"
```
juliakreger: So what if a container is restarted? Do we really want to do this again?  Same with setting the password below, do we really want to in that case?

runironic.sh
elfosardo: I'm not sure having a shared folder just for this would be the best solution.
The idea is to override this using --env during podman run, as it shows in https://github.com/openshift-metalkube/dev-scripts/pull/241, where it's dynamically created

runmariadb.sh
elfosardo: true, missed this as using it in conjunction with https://github.com/openshift-metalkube/dev-scripts/pull/241
I will fix it

runmariadb.sh
elfosardo: I will add a logic for the restart

elfosardo:
> I think we need to iterate once more on this, or discuss if we should cove some of the cases like the container restarting. I'm a little worried about coding in a static password, as well as the default is missing for runmariadb which seems like it is going to cause the change to fail.

the password is dynamically generated in dev-scripts and can be overridden during podman run using --env
I agree on the restart case, didn't think that through

runironic.sh
juliakreger: Ahh! I didn't see that. I think that is perfect then!

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/440/

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/442/

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/459/

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/467/

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/500/

runmariadb.sh
```diff
@@ -0,0 +1,33 @@
...
+CREATE DATABASE IF NOT EXISTS  ironic ;
+FLUSH PRIVILEGES ;
+EOSQL
+
+    mysqld_safe --datadir=$DATADIR --user mysql --init-file /tmp/configure-mysql.sql
+else
+    mysqld_safe --datadir=$DATADIR --user mysql
+fi
+
+sleep infinity
```
celebdor: rather than sleep infinity. I think it would be better to let this main container process become mysqld_safe by changing lines 27 and 29 to start with `exec`

celebdor: In runironic.sh I didn't propose it because it seems that spawns several processes.

elfosardo: I was thinking about it, I agree would be better, I'll do a quick test with that

runmariadb.sh
```diff
@@ -0,0 +1,30 @@
...
+    crudini --set /etc/my.conf mysqld innodb_buffer_pool_size 5M
+    crudini --set /etc/my.conf mysqld innodb_log_buffer_size 512K
+
+    mysql_install_db --datadir="$DATADIR"
+
+    chown -R mysql /var/log/mariadb
+    chown -R mysql "$DATADIR"
+
+    cat > /tmp/configure-mysql.sql <<-EOSQL
+        DELETE FROM mysql.user ;
```
bfournie: I think the last change may be causing a problem. When I test this now the container exits and I see:
$ sudo podman logs mariadb
/bin/runmariadb: line 30: warning: here-document at line 17 delimited by end-of-file (wanted `EOSQL')
/bin/runmariadb: line 31: syntax error: unexpected end of file

I think the indenting for these lines has to go back to the first column, the last change introduced the intending.

elfosardo: you're right, I don't know why this changed got actually reverted during my last commit
it should be back to normal now

bfournie:
The containers starts fine now, however I'm getting this error in /var/log/mariadb/mariadb.log when run with https://github.com/openshift-metalkube/dev-scripts/pull/241:
2019-04-05 19:26:02 0 [Note] /usr/libexec/mysqld: ready for connections.
Version: '10.3.10-MariaDB'  socket: '/var/lib/mysql/mysql.sock'  port: 3306  MariaDB Server
2019-04-05 19:26:05 9 [Warning] Aborted connection 9 to db: 'ironic' user: 'ironic' host: 'localhost' (Got an error reading communication packets)

runironic.sh
```diff
@@ -1,53 +1,53 @@
...
+crudini --set /etc/ironic/ironic.conf DEFAULT default_boot_interface ipxe
+crudini --set /etc/ironic/ironic.conf DEFAULT default_deploy_interface direct
+crudini --set /etc/ironic/ironic.conf DEFAULT enabled_inspect_interfaces inspector,idrac
+crudini --set /etc/ironic/ironic.conf DEFAULT default_inspect_interface inspector
+crudini --set /etc/ironic/ironic.conf DEFAULT rpc_transport json-rpc
+crudini --set /etc/ironic/ironic.conf dhcp dhcp_provider none
+crudini --set /etc/ironic/ironic.conf conductor automated_clean false
+crudini --set /etc/ironic/ironic.conf conductor api_url http://${IP}:${HTTP_PORT}
+crudini --set /etc/ironic/ironic.conf database connection mysql+pymysql://ironic:${MARIADB_PASSWORD}@localhost/ironic?charset=utf8
+crudini --set /etc/ironic/ironic.conf deploy http_url http://${IP}
```
bfournie: http_url should be http://${IP}:${HTTP_PORT}

runironic.sh
```diff
@@ -1,53 +1,53 @@
...
+crudini --set /etc/ironic/ironic.conf DEFAULT enabled_hardware_types ipmi,idrac
+crudini --set /etc/ironic/ironic.conf DEFAULT enabled_vendor_interfaces ipmitool,no-vendor,idrac
+crudini --set /etc/ironic/ironic.conf DEFAULT default_boot_interface ipxe
+crudini --set /etc/ironic/ironic.conf DEFAULT default_deploy_interface direct
+crudini --set /etc/ironic/ironic.conf DEFAULT enabled_inspect_interfaces inspector,idrac
+crudini --set /etc/ironic/ironic.conf DEFAULT default_inspect_interface inspector
+crudini --set /etc/ironic/ironic.conf DEFAULT rpc_transport json-rpc
+crudini --set /etc/ironic/ironic.conf dhcp dhcp_provider none
+crudini --set /etc/ironic/ironic.conf conductor automated_clean false
+crudini --set /etc/ironic/ironic.conf conductor api_url http://${IP}:${HTTP_PORT}
```
bfournie: I think a couple typos in transition from Dockerfile
api_url should be http://${IP}:6385

elfosardo:
> The containers starts fine now, however I'm getting this error in /var/log/mariadb/mariadb.log when run with [openshift-metalkube/dev-scripts#241](https://github.com/openshift-metalkube/dev-scripts/pull/241):
> 2019-04-05 19:26:02 0 [Note] /usr/libexec/mysqld: ready for connections.
> Version: '10.3.10-MariaDB' socket: '/var/lib/mysql/mysql.sock' port: 3306 MariaDB Server
> 2019-04-05 19:26:05 9 [Warning] Aborted connection 9 to db: 'ironic' user: 'ironic' host: 'localhost' (Got an error reading communication packets)
> 
> [Update] Note though I see that same warning message when running with mariadb in the Ironic container prior to this patch so I don't think its significant as the database works fine.

This is just a warning and can be safely ignored. The reason why we're seeing this now it's because in this version of MariaDB (and since version 10.2) the default value for log_warnings went from 1 (warnings not logged) to 2 (warnings logged).
To prevent spamming, we can set log_warnings to 1.
There are some parameters that can be used to try and adjust the database response, but this is not a destructive or blocking issue.

runironic.sh
elfosardo: copy/paste is just too hard for me :/
fixed

runironic.sh
elfosardo: fixed

runironic.sh
bfournie: Thanks, looks good!

runironic.sh
```diff
@@ -1,53 +1,53 @@
...
 IP=${IP:-"172.22.0.1"}
 HTTP_PORT=${HTTP_PORT:-"80"}
 INTERFACE=${INTERFACE:-"provisioning"}
-
-sed -i -e s/IRONIC_IP/$IP/g -e s/HTTP_PORT/$HTTP_PORT/g /etc/ironic/ironic.conf 
+MARIADB_PASSWORD=${MARIADB_PASSWORD:-"change_me"}
 
 # Allow access to Ironic
 if ! iptables -C INPUT -i $INTERFACE -p tcp -m tcp --dport 5050 -j ACCEPT > /dev/null 2>&1; then
     iptables -I INPUT -i $INTERFACE -p tcp -m tcp --dport 5050 -j ACCEPT
```
bfournie: I noticed this copy/paste issue from an earlier change I made - should be 6385 here.  Would be good if you have a chance to change, otherwise we'll pick it up in follow-on.

elfosardo: changed it

PR #24, opened 2019-03-29
When using an empty shared volume, these directories don't exist yet.

imain:


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/452/

PR #25, opened 2019-04-01
Allow to exclude listen interfaces for dnsmasq.

yprokule:
When running dnsmasq on master nodes we don't want to run DHCP
server on a 'production' network with an external DHCP server.
To specify interfaces that have to be excluded specify them as
a comma separated list, for e.g.:
  EXCEPT_INTERFACE="lo,em1,enp12s0"

### DIFF
--- a/dnsmasq.conf
+++ b/dnsmasq.conf
@@ -1,5 +1,4 @@
 interface=INTERFACE
-except-interface=lo
 bind-dynamic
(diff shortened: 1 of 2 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/454/

PR #29, opened 2019-04-15
Store Ironic container logs in shared mount

bfournie:
Store the logs from the Ironic containers in shared area to facilitate log rollup.
This also enables storing of the ironic python agent deploy logs from each node's
deployment.

### DIFF
--- a/runhttpd.sh
+++ b/runhttpd.sh
@@ -18,6 +18,17 @@ sed -i -e 's|\(^[[:space:]]*\)\(DocumentRoot\)\(.*\)|\1\2 "/shared/html"|' \
     -e 's|<Directory "/var/www/html">|<Directory "/shared/html">|' \
     -e 's|<Directory "/var/www">|<Directory "/shared">|' /etc/httpd/conf/httpd.conf
 
+# Remove log files from last deployment
+rm -rf /shared/log/httpd
+   
+mkdir -p /shared/log/httpd
+
+# Make logs available in shared mount
+touch /shared/log/httpd/access_log
+ln -s /shared/log/httpd/access_log /var/log/httpd/access_log
+touch /shared/log/httpd/error_log
+ln -s /shared/log/httpd/error_log /var/log/httpd/error_log
+
 # Allow external access
 if ! iptables -C INPUT -p tcp --dport $HTTP_PORT -j ACCEPT 2>/dev/null ; then
     iptables -I INPUT -p tcp --dport $HTTP_PORT -j ACCEPT
--- a/runironic.sh
+++ b/runironic.sh
@@ -15,7 +15,7 @@ cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig
 
 crudini --set /etc/ironic/ironic.conf DEFAULT auth_strategy noauth
 crudini --set /etc/ironic/ironic.conf DEFAULT my_ip ${IP}
-crudini --set /etc/ironic/ironic.conf DEFAULT debug true
+crudini --set /etc/ironic/ironic.conf DEFAULT debug true 
 crudini --set /etc/ironic/ironic.conf DEFAULT default_network_interface noop
 crudini --set /etc/ironic/ironic.conf DEFAULT enabled_boot_interfaces pxe,ipxe,fake
 crudini --set /etc/ironic/ironic.conf DEFAULT enabled_power_interfaces ipmitool,idrac,fake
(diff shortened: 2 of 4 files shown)


runhttpd.sh
```diff
@@ -18,6 +18,17 @@ sed -i -e 's|\(^[[:space:]]*\)\(DocumentRoot\)\(.*\)|\1\2 "/shared/html"|' \
     -e 's|<Directory "/var/www/html">|<Directory "/shared/html">|' \
     -e 's|<Directory "/var/www">|<Directory "/shared">|' /etc/httpd/conf/httpd.conf
 
+# Remove log files from last deployment
+if [ -d "/shared/log/httpd" ]; then
```
dtantsur: This check is not needed, `-f` flag does it for you. You can do just
```bash
rm -rf /shared/log/httpd
mkdir -p /shared/log/httpd
```

runhttpd.sh
```diff
@@ -18,6 +18,17 @@ sed -i -e 's|\(^[[:space:]]*\)\(DocumentRoot\)\(.*\)|\1\2 "/shared/html"|' \
...
+   rm -rf /shared/log/httpd/*
+else
+   mkdir -p /shared/log/httpd
+fi
+
+# Make logs available in shared mount
+ln -s /shared/log/httpd/access_log /var/log/httpd/access_log
```
dtantsur: I *think* `ln -s` works on non-existing files, but I'd prefer we touched them anyway.

runironic.sh
```diff
@@ -42,12 +42,26 @@ crudini --set /etc/ironic/ironic.conf pxe tftp_master_path /shared/tftpboot
...
 
 ironic-dbsync --config-file /etc/ironic/ironic.conf upgrade
 
 mkdir -p /shared/html
-/usr/bin/python2 /usr/bin/ironic-conductor > /var/log/ironic-conductor.out 2>&1 &
-/usr/bin/python2 /usr/bin/ironic-api > /var/log/ironic-api.out 2>&1 &
+
+# Remove log files from last deployment
+if [ -d "/shared/log/ironic" ]; then
+   rm -rf /shared/log/ironic/*
```
dtantsur: ditto re using two commands

runironic.sh
```diff
@@ -42,12 +42,26 @@ crudini --set /etc/ironic/ironic.conf pxe tftp_master_path /shared/tftpboot
...
+else
+   mkdir -p /shared/log/ironic
+fi
+
+# Make logs available in shared mount
+ln -s /shared/log/ironic/ironic-conductor.log /var/log/ironic/ironic-conductor.log
```
dtantsur: ditto re `ln -s`. also why not link the whole directory instead?

runironic.sh
```diff
@@ -42,12 +42,26 @@ crudini --set /etc/ironic/ironic.conf pxe tftp_master_path /shared/tftpboot
...
+ln -s /shared/log/ironic/ironic-api.log /var/log/ironic/ironic-api.log
+
+/usr/bin/python2 /usr/bin/ironic-conductor 2>&1 | tee /shared/log/ironic/ironic-conductor.out &
```
dtantsur: I think you should use something like `--log-file /shared/log/ironic/ironic-conductor.log` (let's not use `out` extension, it's for `a.out` files).

runmariadb.sh
```diff
@@ -23,8 +25,8 @@ CREATE DATABASE IF NOT EXISTS  ironic ;
 FLUSH PRIVILEGES ;
 EOSQL
 
-    exec mysqld_safe --init-file /tmp/configure-mysql.sql
+    exec mysqld_safe --init-file /tmp/configure-mysql.sql | tee /shared/log/mariadb/mariadb.out
```
dtantsur: Ditto re `out` extension

runhttpd.sh
bfournie: Fixed.

runhttpd.sh
bfournie: I found it does work but added touch for consistency.

runironic.sh
bfournie: Fixed.

runironic.sh
bfournie: Using --log-file is a better way to go, changed to use that.

runironic.sh
bfournie: Used --log-file instead.

runmariadb.sh
```diff
@@ -23,8 +25,8 @@ CREATE DATABASE IF NOT EXISTS  ironic ;
 FLUSH PRIVILEGES ;
 EOSQL
 
-    exec mysqld_safe --init-file /tmp/configure-mysql.sql
+    exec mysqld_safe --init-file /tmp/configure-mysql.sql | tee /shared/log/mariadb/mariadb.log
```
elfosardo: you could use --log-error option or set that up in the config
also may be useful to log general (--general-log) and slow query logs (--slow-query-log)

bfournie: Updated config and added general-log. Needed to set permissions to log properly.

elfosardo:
looks good, needs a rebase

PR #31, opened 2019-04-16
Support ironic_prometheus_exporter

iurygregory:
- Install the oslo messaging notifier Driver
- Configures ironic to use the new Driver
- Run a Flask Application under uWSGI to expose the metrics to Prometheus (by default it will run on port 5001 and this port need to be accessible for the Prometheus Container)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,23 @@
 FROM docker.io/centos:centos7
 
+# TODO(iurygregory): remove epel-release and  other packages necessary for the ironic_prometheus_exporter when we have a package for it
 RUN yum install -y python-requests && \
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - current-tripleo && \
-    yum install -y openstack-ironic-api openstack-ironic-conductor crudini iproute dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs psmisc sysvinit-tools mariadb-server python-PyMySQL python2-chardet && \
+    yum install -y epel-release python-pip python-devel gcc openstack-ironic-api openstack-ironic-conductor crudini iproute dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs psmisc sysvinit-tools mariadb-server python-PyMySQL python2-chardet && \
+    yum install -y python-configparser python2-prometheus_client && \
     yum clean all
 
 RUN mkdir /tftpboot && \
     cp /usr/share/ipxe/undionly.kpxe /usr/share/ipxe/ipxe.efi /tftpboot/
 
+COPY ./installexporter.sh /bin/installexporter
+RUN /bin/installexporter
 COPY ./runironic.sh /bin/runironic
 COPY ./rundnsmasq.sh /bin/rundnsmasq
 COPY ./runhttpd.sh /bin/runhttpd
 COPY ./runmariadb.sh /bin/runmariadb
 COPY ./runhealthcheck.sh /bin/runhealthcheck
+COPY ./runexporterapp.sh /bin/runexporterapp
 
 COPY ./dnsmasq.conf /etc/dnsmasq.conf
 COPY ./inspector.ipxe /tmp/inspector.ipxe
--- a/installexporter.sh
+++ b/installexporter.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/bash
+
+git clone https://github.com/metal3-io/ironic-prometheus-exporter.git
+cd ironic-prometheus-exporter/
+python setup.py install
--- a/runexporterapp.sh
+++ b/runexporterapp.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/bash
+
+cd /ironic-prometheus-exporter
+export IRONIC_CONFIG=/etc/ironic/ironic.conf
+export FLASK_RUN_HOST=0.0.0.0
+export FLASK_RUN_PORT=5001
+uwsgi --socket ${FLASK_RUN_HOST}:${FLASK_RUN_PORT} --protocol=http -w ironic_prometheus_exporter.app.wsgi
--- a/runironic.sh
+++ b/runironic.sh
@@ -31,6 +31,10 @@ crudini --set /etc/ironic/ironic.conf DEFAULT default_deploy_interface direct
 crudini --set /etc/ironic/ironic.conf DEFAULT enabled_inspect_interfaces inspector,idrac,fake
 crudini --set /etc/ironic/ironic.conf DEFAULT default_inspect_interface inspector
 crudini --set /etc/ironic/ironic.conf DEFAULT rpc_transport json-rpc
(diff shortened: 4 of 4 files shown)


runexporterapp.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/bash
+
+cd ironic-prometheus-exporter/
+export IRONIC_CONFIG=/etc/ironic/ironic.conf
+export FLASK_APP=ironic_prometheus_exporter/app/exporter.py
+export FLASK_RUN_HOST=$HOST_IP
+export FLASK_RUN_PORT=5001
+python -m flask run &
```
dtantsur: Let us use a real WSGI runner, like uWSGI or gunicorn. This is not intended for production and will probably behave poorly.

installexporter.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/bash
+
+git clone https://github.com/metalkube/ironic-prometheus-exporter/
```
dtantsur: Any plans to package this for RDO?

iurygregory: We haven't consider this in any meetings or stories, maybe in the next sprint or something I would say.

runexporterapp.sh
iurygregory: Something like http://flask.pocoo.org/docs/1.0/deploying/wsgi-standalone/#gunicorn right?

installexporter.sh
dtantsur: Let's add a story for it. We cannot use `git clone` forever.

runexporterapp.sh
dtantsur: Yeah, if `gunicorn` is in CentOS (I know that `uwsgi` is).

iurygregory: https://centos.pkgs.org/7/centos-extras-x86_64/python-gunicorn-18.0-2.el7.noarch.rpm.html seems we can have it installed =)

Dockerfile
```diff
@@ -3,20 +3,23 @@ FROM docker.io/centos:centos7
 RUN yum install -y python-requests && \
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/5609d2e3aee35578e914bcbfac92a46c759c0a31/tripleo_repos/main.py | python - current && \
     yum install -y openstack-ironic-api openstack-ironic-conductor crudini iproute dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs psmisc sysvinit-tools mariadb-server python-PyMySQL python2-chardet && \
+    yum install -y epel-release python-pip python-devel gcc && \
```
elfosardo: maybe we can add this to the yum command on line 5 instead

Dockerfile
```diff
@@ -3,20 +3,23 @@ FROM docker.io/centos:centos7
...
+    yum install -y python-configparser python2-prometheus_client && \
     yum clean all
 
 RUN mkdir /tftpboot && \
     cp /usr/share/ipxe/undionly.kpxe /usr/share/ipxe/ipxe.efi /tftpboot/
 
+
```
elfosardo: nit: blank line not needed

runexporterapp.sh
```diff
@@ -1,8 +1,8 @@
...
 
-cd ironic-prometheus-exporter/
+cd /ironic-prometheus-exporter
 export IRONIC_CONFIG=/etc/ironic/ironic.conf
 export FLASK_APP=ironic_prometheus_exporter/app/exporter.py
-export FLASK_RUN_HOST=$HOST_IP
+export FLASK_RUN_HOST=0.0.0.0
 export FLASK_RUN_PORT=5001
-python -m flask run &
+uwsgi --socket $FLASK_RUN_HOST:$FLASK_RUN_PORT --protocol=http -w ironic_prometheus_exporter.app.wsgi &
```
elfosardo: let's wrap the bash variables in braces

Dockerfile
iurygregory: Done

Dockerfile
iurygregory: Done

runexporterapp.sh
iurygregory: Done

Dockerfile
dtantsur: Can we also leave a TODO to remove epel-release? We cannot use it in the final product.

runexporterapp.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/bash
+
+cd /ironic-prometheus-exporter
+export IRONIC_CONFIG=/etc/ironic/ironic.conf
+export FLASK_APP=ironic_prometheus_exporter/app/exporter.py
```
dtantsur: Is this actually needed?

iurygregory: I removed in the newest commit

Dockerfile
iurygregory: @dtantsur something like: `# TODO(iurygregory): remove epel-release and  other packages necessary for the ironic_prometheus_exporter when we have a package for it` ?

iurygregory:
@dtantsur when you have sometime can you review again? Thanks!

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/670/

hardys:
Sorry this now needs a rebase, also I think you need to rebase your branch and force-push to avoid the merge commit, and maybe squash some commits e.g "Simple fixes" should perhaps be squashed into the feature commit?

iurygregory:
@hardys Thanks! I will do this

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/682/

runexporterapp.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/bash
+
+cd /ironic-prometheus-exporter
+export IRONIC_CONFIG=/etc/ironic/ironic.conf
+export FLASK_RUN_HOST=0.0.0.0
+export FLASK_RUN_PORT=5001
+uwsgi --socket ${FLASK_RUN_HOST}:${FLASK_RUN_PORT} --protocol=http -w ironic_prometheus_exporter.app.wsgi &
```
dtantsur: I think we should not use & here since it's used in runironic.sh

iurygregory: Done

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/683/

derekhiggins:
lgtm, Just wondering if there is a reason its not a separate container as thats the way we've been going with other services?

iurygregory:
@derekhiggins the data from the nodes are stored in the ironic container (the location is defined in the `ironic.conf`) and the Flask App only knows the files location because it reads the `ironic.conf` in this initial implementation.

derekhiggins:
> @derekhiggins the data from the nodes are stored in the ironic container (the location is defined in the `ironic.conf`) and the Flask App only knows the files location because it reads the `ironic.conf` in this initial implementation.

ok, thanks for the extra info.

PR #43, opened 2019-04-24
Revert "Pin tripleo-repos to avoid bug"

derekhiggins:
This reverts commit 32d19dcf797429eba0c3394d74b8a4e952ae0b48.

This fixup is no longer needed so move back to using master
as the pin may stop working at some stage.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM docker.io/centos:centos7
 
 RUN yum install -y python-requests && \
-    curl https://raw.githubusercontent.com/openstack/tripleo-repos/5609d2e3aee35578e914bcbfac92a46c759c0a31/tripleo_repos/main.py | python - current-tripleo && \
+    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - current-tripleo && \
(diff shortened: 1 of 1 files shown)


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/599/

Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM docker.io/centos:centos7
 
 RUN yum install -y python-requests && \
-    curl https://raw.githubusercontent.com/openstack/tripleo-repos/5609d2e3aee35578e914bcbfac92a46c759c0a31/tripleo_repos/main.py | python - current && \
+    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - current && \
```
dtantsur: maybe it's time to add `-b stein`? or do we plan to keep using master?

derekhiggins: Not sure,I guess we need to keep at master if we want to start using the mdns work (or any other new features) when they are ready. 

Either way this patch is just reversing a workaround we needed at one stage,

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/643/

PR #45, opened 2019-04-29
Switch from "current" to "current-tripleo-rdo"

derekhiggins:
Moving to use a repository that has passed the RDO
ci should protect us a bit better from regressions.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/608/

PR #47, opened 2019-05-16
Set uefi_pxe_config_template to use the ipxe template

derekhiggins:
Our dnsmasq config is setup to send "ipxe.efi" incases
where the host is UEFI booting. But by default ironic
sends a grub2 config when using UEFI boot. Switch
Ironic to use ipxe so they both match up and booting
UEFI works.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/638/

PR #48, opened 2019-05-24
Add more configuration options to containers.

imain:
Add the ability to configure the containers based on interface
rather than specific IPs.  We don't know the IP address the pod
will be run on so we need a way to do this dynamically.

### DIFF
--- a/dnsmasq.conf
+++ b/dnsmasq.conf
@@ -1,4 +1,4 @@
-interface=INTERFACE
+interface=PROVISIONING_INTERFACE
 bind-dynamic
 dhcp-range=DHCP_RANGE
 enable-tftp
--- a/rundnsmasq.sh
+++ b/rundnsmasq.sh
@@ -1,28 +1,33 @@
 #!/usr/bin/bash
 
-IP=${IP:-"172.22.0.1"}
+PROVISIONING_INTERFACE=${PROVISIONING_INTERFACE:-"provisioning"}
+
 HTTP_PORT=${HTTP_PORT:-"80"}
 DHCP_RANGE=${DHCP_RANGE:-"172.22.0.10,172.22.0.100"}
-INTERFACE=${INTERFACE:-"provisioning"}
-EXCEPT_INTERFACE=${EXCEPT_INTERFACE:-"lo"}
+DNSMASQ_EXCEPT_INTERFACE=${DNSMASQ_EXCEPT_INTERFACE:-"lo"}
+
+PROVISIONING_IP=$(ip -4 address show dev "$PROVISIONING_INTERFACE" | grep -oP '(?<=inet\s)\d+(\.\d+){3}' | head -n 1)
 
 mkdir -p /shared/tftpboot
+mkdir -p /shared/html/images
+mkdir -p /shared/html/pxelinux.cfg
 mkdir -p /shared/log/dnsmasq
 
 # Copy files to shared mount
 cp /usr/share/ipxe/undionly.kpxe /usr/share/ipxe/ipxe.efi /shared/tftpboot
 
 # Use configured values
-sed -i -e s/IRONIC_IP/$IP/g -e s/HTTP_PORT/$HTTP_PORT/g \
-       -e s/DHCP_RANGE/$DHCP_RANGE/g -e s/INTERFACE/$INTERFACE/g /etc/dnsmasq.conf
-for iface in $( echo $EXCEPT_INTERFACE | tr ',' ' '); do
-    sed -i -e "/^interface=.*/ a\except-interface=$iface" /etc/dnsmasq.conf
+sed -i -e s/IRONIC_IP/${PROVISIONING_IP}/g -e s/HTTP_PORT/${HTTP_PORT}/g \
+       -e s/DHCP_RANGE/${DHCP_RANGE}/g -e s/PROVISIONING_INTERFACE/${PROVISIONING_INTERFACE}/g \
+       /etc/dnsmasq.conf
+for iface in $( echo "$DNSMASQ_EXCEPT_INTERFACE" | tr ',' ' '); do
+    sed -i -e "/^interface=.*/ a\except-interface=${iface}" /etc/dnsmasq.conf
 done
 
 # Allow access to dhcp and tftp server for pxeboot
 for port in 67 69 ; do
-    if ! iptables -C INPUT -i $INTERFACE -p udp --dport $port -j ACCEPT 2>/dev/null ; then
-        iptables -I INPUT -i $INTERFACE -p udp --dport $port -j ACCEPT
+    if ! iptables -C INPUT -i "$PROVISIONING_INTERFACE" -p udp --dport "$port" -j ACCEPT 2>/dev/null ; then
+        iptables -I INPUT -i "$PROVISIONING_INTERFACE" -p udp --dport "$port" -j ACCEPT
     fi
 done
 
--- a/runhttpd.sh
+++ b/runhttpd.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/bash
 
-IP=${IP:-"172.22.0.1"}
+PROVISIONING_INTERFACE=${PROVISIONING_INTERFACE:-"provisioning"}
 HTTP_PORT=${HTTP_PORT:-"80"}
+HTTP_IP=$(ip -4 address show dev "$PROVISIONING_INTERFACE" | grep -oP '(?<=inet\s)\d+(\.\d+){3}' | head -n 1)
 
 mkdir -p /shared/html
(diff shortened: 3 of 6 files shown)


rundnsmasq.sh
```diff
@@ -1,28 +1,33 @@
...
-IP=${IP:-"172.22.0.1"}
+PROVISIONING_INTERFACE=${PROVISIONING_INTERFACE:-"provisioning"}
+
 HTTP_PORT=${HTTP_PORT:-"80"}
 DHCP_RANGE=${DHCP_RANGE:-"172.22.0.10,172.22.0.100"}
-INTERFACE=${INTERFACE:-"provisioning"}
-EXCEPT_INTERFACE=${EXCEPT_INTERFACE:-"lo"}
+DNSMASQ_EXCEPT_INTERFACE=${DNSMASQ_EXCEPT_INTERFACE:-"lo"}
+
+PROVISIONING_IP=`ip -4 address show dev $PROVISIONING_INTERFACE | grep -oP '(?<=inet\s)\d+(\.\d+){3}' | head -n 1`
```
elfosardo: please use $() instead of backquotes to wrap the command
wrap $PROVISIONING_INTERFACE in double-quotes "
also I would be careful with perl-regex in grep as it's not universally portable, but it's probably fine

rundnsmasq.sh
```diff
@@ -1,28 +1,33 @@
...
 mkdir -p /shared/log/dnsmasq
 
 # Copy files to shared mount
 cp /usr/share/ipxe/undionly.kpxe /usr/share/ipxe/ipxe.efi /shared/tftpboot
 
 # Use configured values
-sed -i -e s/IRONIC_IP/$IP/g -e s/HTTP_PORT/$HTTP_PORT/g \
-       -e s/DHCP_RANGE/$DHCP_RANGE/g -e s/INTERFACE/$INTERFACE/g /etc/dnsmasq.conf
-for iface in $( echo $EXCEPT_INTERFACE | tr ',' ' '); do
+sed -i -e s/IRONIC_IP/$PROVISIONING_IP/g -e s/HTTP_PORT/$HTTP_PORT/g \
```
elfosardo: since we're here, I would wrap all the variables in braces ${}

rundnsmasq.sh
```diff
@@ -1,28 +1,33 @@
...
+       -e s/DHCP_RANGE/$DHCP_RANGE/g -e s/PROVISIONING_INTERFACE/$PROVISIONING_INTERFACE/g \
+       /etc/dnsmasq.conf
+for iface in $( echo $DNSMASQ_EXCEPT_INTERFACE | tr ',' ' '); do
```
elfosardo: let's wrap $DNSMASQ_EXCEPT_INTERFACE in double-quotes "

rundnsmasq.sh
```diff
@@ -1,28 +1,33 @@
...
     sed -i -e "/^interface=.*/ a\except-interface=$iface" /etc/dnsmasq.conf
 done
 
 # Allow access to dhcp and tftp server for pxeboot
 for port in 67 69 ; do
-    if ! iptables -C INPUT -i $INTERFACE -p udp --dport $port -j ACCEPT 2>/dev/null ; then
-        iptables -I INPUT -i $INTERFACE -p udp --dport $port -j ACCEPT
+    if ! iptables -C INPUT -i $PROVISIONING_INTERFACE -p udp --dport $port -j ACCEPT 2>/dev/null ; then
```
elfosardo: ditto double-quotes "

rundnsmasq.sh
```diff
@@ -1,28 +1,33 @@
...
+        iptables -I INPUT -i $PROVISIONING_INTERFACE -p udp --dport $port -j ACCEPT
```
elfosardo: ditto

runhttpd.sh
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/bash
 
-IP=${IP:-"172.22.0.1"}
+PROVISIONING_INTERFACE=${PROVISIONING_INTERFACE:-"provisioning"}
 HTTP_PORT=${HTTP_PORT:-"80"}
+HTTP_IP=`ip -4 address show dev $PROVISIONING_INTERFACE | grep -oP '(?<=inet\s)\d+(\.\d+){3}' | head -n 1`
```
elfosardo: please use $() instead of backquotes to wrap the command
let's wrap $PROVISIONING_INTERFACE in double-quotes "

runhttpd.sh
```diff
@@ -11,7 +12,7 @@ cp /tmp/inspector.ipxe /shared/html/inspector.ipxe
 cp /tmp/dualboot.ipxe /shared/html/dualboot.ipxe
 
 # Use configured values
-sed -i -e s/IRONIC_IP/$IP/g -e s/HTTP_PORT/$HTTP_PORT/g /shared/html/inspector.ipxe
+sed -i -e s/IRONIC_IP/$HTTP_IP/g -e s/HTTP_PORT/$HTTP_PORT/g /shared/html/inspector.ipxe
```
elfosardo: use braces here to wrap variables ${}

runironic.sh
```diff
@@ -1,24 +1,24 @@
 #!/usr/bin/bash
 
 # Get environment settings and update ironic.conf
-IP=${IP:-"172.22.0.1"}
+PROVISIONING_INTERFACE=${PROVISIONING_INTERFACE:-"provisioning"}
+IRONIC_IP=`ip -4 address show dev $PROVISIONING_INTERFACE | grep -oP '(?<=inet\s)\d+(\.\d+){3}' | head -n 1`
```
elfosardo: please use $() instead of backquotes to wrap the command
wrap $PROVISIONING_INTERFACE in double-quotes "

runironic.sh
```diff
@@ -1,24 +1,24 @@
...
-INTERFACE=${INTERFACE:-"provisioning"}
 MARIADB_PASSWORD=${MARIADB_PASSWORD:-"change_me"}
 NUMPROC=$(cat /proc/cpuinfo  | grep "^processor" | wc -l)
 NUMWORKERS=$(( NUMPROC < 12 ? NUMPROC : 12 ))
 
 
 # Allow access to Ironic
-if ! iptables -C INPUT -i $INTERFACE -p tcp -m tcp --dport 6385 -j ACCEPT > /dev/null 2>&1; then
-    iptables -I INPUT -i $INTERFACE -p tcp -m tcp --dport 6385 -j ACCEPT
+if ! iptables -C INPUT -i $PROVISIONING_INTERFACE -p tcp -m tcp --dport 6385 -j ACCEPT > /dev/null 2>&1; then
```
elfosardo: ditto double-quotes "

runironic.sh
```diff
@@ -1,24 +1,24 @@
...
+    iptables -I INPUT -i $PROVISIONING_INTERFACE -p tcp -m tcp --dport 6385 -j ACCEPT
```
elfosardo: ditto double-quotes "

runironic.sh
```diff
@@ -1,24 +1,24 @@
...
 fi
 
 cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig
 
 crudini --set /etc/ironic/ironic.conf DEFAULT auth_strategy noauth
-crudini --set /etc/ironic/ironic.conf DEFAULT my_ip ${IP}
-crudini --set /etc/ironic/ironic.conf DEFAULT debug true 
+crudini --set /etc/ironic/ironic.conf DEFAULT my_ip ${IRONIC_IP}
```
elfosardo: no need for braces here but better wrap in double-quotes "

dnsmasq.conf
```diff
@@ -1,7 +1,7 @@
-interface=INTERFACE
+interface=PROVISIONING_INTERFACE
 bind-dynamic
 dhcp-range=DHCP_RANGE
-enable-tftp
+enable-tftp=PROVISIONING_INTERFACE
```
bfournie: Not sure we need this (man page - "Without an argument, the TFTP service is provided to the same set of interfaces as DHCP service"), but it doesn't hurt.

rundnsmasq.sh
bfournie: One concern is that this must change in multiple places if/when we support IPv6.

imain: Yes agreed.  We'll have to keep that in mind.  I don't see a way around this for now.

dnsmasq.conf
imain: yeah this was a leftover.

imain:
@elfosardo I think I got them all :).

imain:
Is there something I need to do here?  Would like to get this in..

PR #50, opened 2019-06-05
Publish ironic endpoint via mDNS

dtantsur:
Requires https://github.com/metal3-io/ironic-inspector-image/pull/21 and a recent IPA image.


dtantsur:
Added another patch that makes ironic-inspector actually use mdns. Requires pretty recent ironic-python-agent from RDO.

dtantsur:
Same answer: this is intended because they're processed differently on the IPA side.

PR #53, opened 2019-06-11
Do not use hard coded Python 2

### LINKED ISSUE #19: Container is hard coded to use python2.

elfosardo:
Fix #19

### DIFF
--- a/runironic.sh
+++ b/runironic.sh
@@ -74,8 +74,8 @@ rm -rf /shared/log/ironic
 
 mkdir -p /shared/log/ironic
 
-/usr/bin/python2 /usr/bin/ironic-conductor --log-file /shared/log/ironic/ironic-conductor.log &
-/usr/bin/python2 /usr/bin/ironic-api --log-file  /shared/log/ironic/ironic-api.log &
+/usr/bin/ironic-conductor --log-file /shared/log/ironic/ironic-conductor.log &
+/usr/bin/ironic-api --log-file  /shared/log/ironic/ironic-api.log &
 
 /bin/runhealthcheck "ironic" &>/dev/null &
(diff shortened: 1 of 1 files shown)


runironic.sh
```diff
@@ -74,8 +74,8 @@ rm -rf /shared/log/ironic
 
 mkdir -p /shared/log/ironic
 
-/usr/bin/python2 /usr/bin/ironic-conductor --log-file /shared/log/ironic/ironic-conductor.log &
-/usr/bin/python2 /usr/bin/ironic-api --log-file  /shared/log/ironic/ironic-api.log &
+/usr/bin/python /usr/bin/ironic-conductor --log-file /shared/log/ironic/ironic-conductor.log &
+/usr/bin/python /usr/bin/ironic-api --log-file  /shared/log/ironic/ironic-api.log &
```
juliakreger: so we shouldn't need to directly invoke with python on the path as PBR generates the script it includes a "#!/usr/bin/python". Ultimately we should want to shift to python3, but I'm guessing this is one of the areas current support in centos is problematic to build the images :\

elfosardo: I actually did a test with Python 3.x but it was not very sucessfull :/
and yes, you're right, I should just remove the python binary invocation

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/733/

dtantsur:
@elfosardo could you rename your fork repo to ironic-image? The CI is pretty picky about names.

elfosardo:
@dtantsur sure thing! done :)

dtantsur:
Passed http://10.8.144.11:8080/job/dev-tools/734/

PR #56, opened 2019-06-13
WIP Set host_ip for ironic API bind address

hardys:
This ensures we only listen on the provisioning network, not
0.0.0.0

Fixes: #55

### DIFF
--- a/runironic.sh
+++ b/runironic.sh
@@ -64,6 +64,7 @@ crudini --set /etc/ironic/ironic.conf pxe uefi_pxe_config_template \$pybasedir/d
 crudini --set /etc/ironic/ironic.conf agent deploy_logs_collect always
 crudini --set /etc/ironic/ironic.conf agent deploy_logs_local_path /shared/log/ironic/deploy
 crudini --set /etc/ironic/ironic.conf api api_workers "$NUMWORKERS"
+crudini --set /etc/ironic/ironic.conf api host_ip ${IRONIC_IP}
 
 ironic-dbsync --config-file /etc/ironic/ironic.conf upgrade
 


hardys:
> The change looks good. Do we need to change the instructions for accessing ironic API using openstackclient?

Hmm good point, dev-scripts says `export OS_URL=http://localhost:6385/` but pretty soon I'm going to move ironic to the bootstrap VM, so that was going to change to the API VIP, which won't work with this new setup.

Hmm.

hardys:
I think having this traffic internal and not accessible via the public network is correct from a security perspective, so perhaps we'll just set up some port forwarding via SSH in dev-scripts to make the move to the bootstrap VM transparent, and in the interim we could setup an iptables rule to forward from localhost to the provisioning nic?

hardys:
Marking this WIP until we figure out the best way to not break existing deployments/testing.

hardys:
Actually there is some complexity here;

- On the bootstrap VM we need to access ironic from the installer, e.g via the public/baremetal nic, and the deployed nodes need to reach ironic via the provisioning network

- On the cluster we really don't want to bind to all interfaces, or tenant users connecting to the k8s API will be able to see it, modulo firewall rules which I've not yet checked.

I think the safest option is to bind to a specific network in the container config, then when needed (e.g on the bootstrap VM) set up a forwarding rule to enable the necessary access.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/740/

derekhiggins:
> I think having this traffic internal and not accessible via the public network is correct from a security perspective, so perhaps we'll just set up some port forwarding via SSH in dev-scripts to make the move to the bootstrap VM transparent, and in the interim we could setup an iptables rule to forward from localhost to the provisioning nic?

An easier path might be to add an IP to the provisioning bridge on the virt host and then set a route for provisioning traffic to use it.

imain:
Yeah I've noticed that the baremetal operator uses localhost for the ironic and ironic-inspector endpoints as well.

hardys:
Removed the lifecycle/stale as this is still and issue which should be fixed IMO, we just need to figure out how to do it without breaking things that depend on the current behavior

Issue #49, opened 2019-05-31
Support vendor hardware interface(Fujitsu iRMC driver)


furukawa3:

dhellmann:
I think this issue is completed by #74. If not, please reopen this ticket.

PR #61, opened 2019-06-24
Allows hardware types and interfaces to be customizable

tiendc:
Currently hardware types and interfaces are hard-code,
this patch allows they to be able to change by operator.

#49 
Signed-off-by: Dao Cong Tien <tiendc@vn.fujitsu.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,3 +26,6 @@ The following environment variables can be passed in to customize run-time funct
 - HTTP_PORT - port used by http server (default 80)
 - DHCP_RANGE - dhcp range to use for provisioning (default 172.22.0.10-172.22.0.100)
 - MARIADB_PASSWORD - The database password
+- ENABLED_HARDWARE_TYPES - hardware types to enable in ironic
+- ENABLED_\<HWIF\>_INTERFACES - hardware interfaces to enable in ironic, with \<HWIF\> is the name of the hardware interface
+- DEFAULT_\<HWIF\>_INTERFACE - default value for a hardware interface, with \<HWIF\> is the name of the hardware interface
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -13,6 +13,9 @@ RUN yum install -y python-requests && \
 RUN mkdir /tftpboot && \
     cp /usr/share/ipxe/undionly.kpxe /usr/share/ipxe/ipxe.efi /tftpboot/
 
+# Required by vendor driver Fujitsu iRMC
+RUN pip install python-scciclient
```
dtantsur: Please no pip, we're removing EPEL from here. scciclient should be available from the repository we use.

runironic.sh
```diff
@@ -8,6 +8,17 @@ MARIADB_PASSWORD=${MARIADB_PASSWORD:-"change_me"}
 NUMPROC=$(cat /proc/cpuinfo  | grep "^processor" | wc -l)
 NUMWORKERS=$(( NUMPROC < 12 ? NUMPROC : 12 ))
 
+# Configuration for hardware types and interfaces
+ENABLED_HARDWARE_TYPES=${ENABLED_HARDWARE_TYPES:-"ipmi,idrac,fake-hardware"}
```
dtantsur: Why not add irmc here since you're installing scciclient?

tiendc: iRMC currently does not support iPXE boot interface, if I add irmc to this configuration, then with default_boot_interface is ipxe, ironic-conductor will fail to start with error IncompatibleInterface. That's why if operator needs to use irmc, it should be passed through env variables.

dtantsur: I think it has been fixed on master.

tiendc: Thanks. I've checked and seen that irmc now support ipxe. Also, installing openstack-ironic-conductor already includes python-scciclient. So I leave this patch's purpose is to allow to customize hardware types and interfaces. Adding support for irmc in ironic.conf will be done in a separated pull request.

PR #62, opened 2019-06-27
Allow running ironic API, conductor and exporter separately

dtantsur:
We want to stop running several processes in one container, since it
violates container good practices and makes collecting logs and
health checking much harder.

### DIFF
--- a/runironic-exporter.sh
+++ b/runironic-exporter.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/bash
+
+. /bin/configure-ironic.sh
+
+# TODO(dtantsur): merge these two scripts when/if we no longer support
+# all-in-one container
+/bin/runexporterapp
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,9 @@ COPY ./runhttpd.sh /bin/runhttpd
 COPY ./runmariadb.sh /bin/runmariadb
 COPY ./configure-ironic.sh /bin/configure-ironic.sh
 
-# TODO(dtantsur): remove this if/when we stop supporting all-in-one container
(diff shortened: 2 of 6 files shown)


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/818/

runironic-exporter.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/bash
+
+. /bin/configure-ironic.sh
+
+# TODO(dtantsur): merge these two scripts when/if we no longer support
+# all-in-one container
```
bfournie: Its not quite clear to me what this means, is this referring to when we use a different image for ironic-prometheus-exporter?

hardys: At this point I'd prefer to avoid adding any new images - we already have two critical tasks (machine-api-opearator integration of the baremetal operator, and cluster hosted ironic) blocked on getting forks of the existing images into openshift, adding any more right at this moment is likely to further delay things, which we really, really do need to avoid :)

That said we can start planning for new images to be used in future, but just not right now when those critical tasks are still pending.

dtantsur:
I didn't really mean splitting the image here, just the eventual removal of the all-in-one ironic entry point.

Note, however, that there is certain risk that openshift folks won't accept the way we build images, but that's a different story.

-------- Original Message --------

dtantsur:
I'd like to merge this change this change since the CI works on it, and it alone is backward compatible with how things are run right now. Opinions?

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/841/

PR #65, opened 2019-07-08
Limit dbsync to only the Conductor entry point

dtantsur:
Ironic API will happily start without database connection, while
Conductor requires database on start-up. Running dbsync in two
places is potentially racy, so limit it to Conductor only.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/849/

PR #70, opened 2019-07-09
Open httpd firewall ports for all interfaces

hardys:
Httpd is configured to listen on all interfaces.  I'd like to use
the external interface to download the cached images so we open
up all interfaces for listening.

### DIFF
--- a/runhttpd.sh
+++ b/runhttpd.sh
@@ -31,8 +31,8 @@ touch /shared/log/httpd/error_log
 ln -s /shared/log/httpd/error_log /var/log/httpd/error_log
 
 # Allow external access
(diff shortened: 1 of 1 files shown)


hardys:
@imain this seems fine, but we do already open this port in dev-scripts ref 02_configure_host.sh - and I think we actually only need this in the case where dev-scripts runs httpd on the host for the purpose of caching the image?

imain:
That's right Steve.  I'll test again today but yesterday I'm pretty sure I had to run this command to allow the host, so something is still in place.

You're right though, I should just do this in dev-scripts as that's where I need this change.  Or do we ever forsee needing this elsewhere?

imain:
Closing this as we don't need it.  Opened the port in dev-scripts instead.

PR #73, opened 2019-07-11
Limit mDNS to the provisioning interface

dtantsur:
This still causes mdns queries responded on the interface the bridge is attached to.
I'm not sure if it's a problem or an implementation detail.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/879/

PR #76, opened 2019-07-15
Switch to gunicorn

iurygregory:
Switching to gunicorn, since uwsgi is not available in the OSP packages.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/894/

PR #80, opened 2019-07-25
Add set -e to all the entrypoint scripts

hardys:
I noticed that it's possible for the dbsync in the conductor script
to fail (if mariadb is not yet up), which is silently ignored and
results in a broken Ironic.

It's probably one of many possible failures, so lets set -e so containers
exit when an error happens, which allows for automated restart e.g
via k8s or systemd.

### DIFF
--- a/rundnsmasq.sh
+++ b/rundnsmasq.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/bash
 
+set -e
+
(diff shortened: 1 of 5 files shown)


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/941/

PR #83, opened 2019-07-30
Revert "Add set -e to all the entrypoint scripts"

derekhiggins:
(updated to revert)
We have always been ignoring errors from the iptables
commands(on master nodes). Revert this while we figure
out a way to eliminate them altogether.

This reverts commit 7e7eb9a.

### DIFF
--- a/rundnsmasq.sh
+++ b/rundnsmasq.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/bash
 
-set -e
-
 PROVISIONING_INTERFACE=${PROVISIONING_INTERFACE:-"provisioning"}
 
 HTTP_PORT=${HTTP_PORT:-"80"}
--- a/runhttpd.sh
+++ b/runhttpd.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/bash
 
-set -e
-
 PROVISIONING_INTERFACE=${PROVISIONING_INTERFACE:-"provisioning"}
 HTTP_PORT=${HTTP_PORT:-"80"}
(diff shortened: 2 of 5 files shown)


dtantsur:
Shouldn't we just revert the change to add `set -e`?

russellb:
I actually think iptables commands should be removed. They aren’t needed in the cluster. These should just be documented as network requirements for this image instead.

russellb:
I’m Ok merging this as a quick fix first but would like to follow up on removal

derekhiggins:
> I’m Ok merging this as a quick fix first but would like to follow up on removal

ya, i suspect we dont need them at all also but want to make sure @hardys doesn't rely on them in his patch to move ironic to the bootstrap node before removing them.

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/963/

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/964/

russellb:
> Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/964/

Ironic had to work on the provisioning host at least to get this far

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/967/

hardys:
Note we do need some iptables rules for both the bootstrap vm case (I set some in the startironic.sh script but removing these may mean more rules are needed there cc @stbenjam )

Then on the cluster we need something to set rules that enable IPA access the the ironic/inspector API so that will need to happen somewhere else if we remove from the container images, cc @imain )

hardys:
Also note we should consider re adding the set -e for the conductor start script, or explicitly check to ensure the dB sync worked, otherwise the systemd service on the bootstrap VM looks active, but really failed

PR #86, opened 2019-08-09
[WIP] Yet another build from source

elfosardo:
This patch builds ironic and its dependencies (and the dependencies of the dependencies and... well you get it) from source.
This is done to reduce the usage of rpms, unless strictly necessary and only if the rpms are available in standard base repos, so to simulate a possible process where external repos are not accessible.
We download all the source packages in a common dir and build and install them from there.
The build process works as-is, but some deps are probably missing, so a run test is necessary to determine what we need to add in terms of python libs from rpms.
In a common internal close building infrastructure it might happen that the pypy repo is not available but there are ther ways to get the source packages in that case, I'm not entirely crazy :)


Dockerfile
```diff
@@ -3,14 +3,21 @@ FROM docker.io/centos:centos7
 RUN yum install -y python-requests && \
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - -b stein current-tripleo && \
     yum update -y && \
-    yum install -y openstack-ironic-api openstack-ironic-conductor crudini \
-        iproute iptables dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs psmisc sysvinit-tools \
-        mariadb-server python-PyMySQL python2-chardet && \
+    yum install -y crudini iproute iptables dnsmasq httpd qemu-img-ev \
+        iscsi-initiator-utils parted gdisk ipxe-bootimgs psmisc \
+        sysvinit-tools mariadb-server python-PyMySQL python2-chardet \
+        python-pip git gcc python-devel && \
```
dtantsur: ipmitool?

dtantsur: also pip requires epel, we cannot just install it..

Dockerfile
```diff
@@ -3,14 +3,21 @@ FROM docker.io/centos:centos7
...
+
+RUN chmod +x /bin/installer.sh && /bin/installer.sh
+
 RUN mkdir /tftpboot && \
     cp /usr/share/ipxe/undionly.kpxe /usr/share/ipxe/ipxe.efi /tftpboot/
 
+RUN mkdir -p /etc/ironic
 COPY ./ironic.conf /tmp/ironic.conf
 RUN crudini --merge /etc/ironic/ironic.conf < /tmp/ironic.conf && \
     rm /tmp/ironic.conf
```
dtantsur: remove gcc and other build dependencies at some point

elfosardo: yep, true

Dockerfile
elfosardo: I think pip doesn't require epel downstream, so we should be good

elfosardo:
it's passing CI \o/
need to hack some stuff

Dockerfile
dtantsur: I'm pretty sure RHEL did not have pip the last time I checked. Maybe it has changed..

dtantsur: Hmm, you're right, apparently it was added in RHEL 8. CentOS 7 is still an issue, I wonder if tripleo-repos enabled EPEL.

Dockerfile
```diff
@@ -3,14 +3,25 @@ FROM docker.io/centos:centos7
 RUN yum install -y python-requests && \
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - -b stein current-tripleo && \
```
dtantsur: We have to stop doing this.

elfosardo: can't we keep doing it for metal3?
downstream we can access stable packages anyway, if they're not available in rhel then we probably have to find an alternative, but for now we have access to osp pkgs

Dockerfile
elfosardo: done

Dockerfile
dtantsur: isn't the goal here to stop using RDO/OSP packages?

Dockerfile
juliakreger: It would not surprise me if tripleo-repos enables EPEL since it is the only place I could find some packages when I was looking at a partial rpm + source approach.

Dockerfile
juliakreger: Kind of is the goal. Where can we find crudini ?

actual-reqs.txt
```diff
@@ -0,0 +1,128 @@
...
+dogpile-cache
+eventlet
+extras
+fasteners
+futurist
+greenlet
+idna
+#should we install this from git instead ?
+#becomes complicated downstream though
+ironic
```
juliakreger: Would we pickup last release in that case downstream? I suspect that might be okay although I suspect we need some more flexibility upstream than we do downstream with brand new features.

installer.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+SOURCE_DIR="/var/tmp/sources-install/"
+REQUIREMENTS_FILE="/var/tmp/actual-reqs.txt"
+
+mkdir -p $SOURCE_DIR
+
+pip install --upgrade pip
+
+pip download -r $REQUIREMENTS_FILE -d $SOURCE_DIR
```
juliakreger: I guess this could actually work really well if we had a downstream pip mirror.....

elfosardo: working on that, or something similar ;)
but there's a way to get the tarballs downstream without using pip

Dockerfile
elfosardo: the goal is to stop using anything coming from OSP/RDO, but for some packages if they're available in old version that are accessible and usable I don't see why not keeping doing that until we find an alternative

elfosardo: also there are alternatives to crudini if we really can't find a way to use it downstream

actual-reqs.txt
elfosardo: we can specify the version we need, upstream is already doable as we're using pip
downstream is early WIP

Dockerfile
dtantsur: > Kind of is the goal. Where can we find crudini

crudini can be also installed from pip?

> for some packages if they're available in old version that are accessible and usable I don't see why not keeping doing that until we find an alternative

right, but it's pretty urgent

installer.sh
dtantsur: can we also mirror requirements and use upper-constraints?

elfosardo: that's the plan :)

yazug:
Might be able to use the Builder copy in model to avoid having to install python-devel and other devel packages and then remove them.  But this is an optimization only

elfosardo:
I started running tests with the actual container, it seems working with the updated list.

russellb:
This PR shouldn't be phrased in terms of upstream and downstream, since that's not applicable to metal3-io.  Can you rephrase this in terms of the upstream benefit only?  Otherwise, this should be worked on in OpenShift only.

Dockerfile
yazug: tripleo repos If I remember correctly do not enable epel but instead rely on RDO's RPM builds of packages for CI.

Dockerfile
yazug: crudini is available via pypi last time I checked.  https://pypi.org/project/crudini/

elfosardo:
@russellb That's right, I will rephrase the PR message.
Although one of the upstream benefits (metal3-io or Openshift) is that this is reproducible in Openshift downstream, and it's done exactly to reproduce a possible downstream experience with its limitations, for example the non-accessibility to external repos.
Especially for the building process, we shouldn't keep too many differences between metal3-io and Openshift, so IMHO it makes sense to apply the change first here and then to Openshift.

Dockerfile
elfosardo: @yazug  thanks! I will update the list :)

elfosardo:
@dhellmann if we're going for this solution, and at the moment we're still evaluating other paths, we're probably end up doing that

dtantsur:
> Have you looked at just copying the source for all of the dependencies into this repository, and treating ironic like we do go programs where the developers are responsible for managing the dependencies?

The "developers" are not doing it for ironic (in this repository our role is "packagers", not "developers"). In Go projects it is upstream that is responsible for vendoring (which, for the record, is a horrible idea). Here we're becoming responsible for it without any support from upstream.

elfosardo:
We probably won't need this anymore, leaving this open but on hold until further notice

elfosardo:
going to close this since we're back to RPMs

PR #87, opened 2019-08-14
Wheel vendoring

dhellmann:
This  is an alternative approach to #86 which relies on the developer to manage the source dependencies as we do with go apps. The build script is going to need some work, as is the dependency list, but as a proof-of-concept it's ready for review.


tools/update-deps.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -xe
+
+source $(dirname $0)/common.sh
+
+update_dep_from_pypi alembic 1.0.11
```
hardys: Interested to know why we can't use pip here, e.g clone the project repo, then do `pip install .`

I know it's not in RHEL by default, but for upstream use that would be OK and better aligned with typical OpenStack install-from-source usage e.g devstack?

dhellmann: The idea is to treat both images the same, and only vary them based on the inputs to the source code where we actually need to. That's what we do for go.

tools/common.sh
```diff
@@ -0,0 +1,81 @@
...
+
+    # Remove any versioned copies of the dep, in case we're switching
+    # from PyPI to git.
+    rm -rf $UNPACK/${dep}-[0-9]*
+
+    # Remove any unversioned copies of the dep
+    rm -rf $UNPACK/${dep}
+
+    git clone ${repo} $UNPACK/$dep
+    (cd $UNPACK/${dep} && git checkout ${ref})
```
dhellmann: Instead of including the whole git tree, we probably want to check it out to /tmp and build an sdist that we can then unpack with the previous function. That would make the treatment of both sets of sources consistent, and give us a real version number.

tools/common.sh
```diff
@@ -0,0 +1,81 @@
...
+    if [ ! -f $expected_file ]
+    then
+        # Remove any previous versions that might be hanging around
+        rm -f $DOWNLOAD/${dep}-[0-9]*
+
+        pip download \
+            -d $DOWNLOAD \
+            --no-deps \
+            --no-binary :all: \
+            ${dep}===${version}
```
dhellmann: This creates a ${dep}-${version} directory. We should probably rename that to just ${dep} to make cleaning simpler and to make it easier to track changes in the code as dependencies are updated.

dhellmann:
I updated this to be more consistent and only check in the unrolled sdist, even when producing that from a git repo ourselves.

elfosardo:
@dhellmann this solution has been already evaluated but excluded because of different reasons, for example the complexity of managing so much code and tracking versions, the size of the repo, security and legal implications (especially for Openshift), sharing the source code with other image repos.
I understand something similar is done with go, but python looks very different in terms of dependencies and number of third party libraries.
Please also consider that at the moment we're evaluating other solutions that are not source based, we'll have more details in the next days.

dtantsur:
I think we're keeping the packaging-based approach for the time being.

PR #100, opened 2019-10-07
Start using Centos 8 as base

elfosardo:
Start testing images with Centos 8 as base.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,17 +28,16 @@ RUN dd bs=1024 count=2880 if=/dev/zero of=esp.img && \
       mcopy -i esp.img -v /tmp/grubx64.efi ::EFI/BOOT && \
       mdir -i esp.img ::EFI/BOOT
 
+FROM docker.io/centos:centos8
 
-FROM docker.io/centos:centos7
-
-RUN yum install -y python-requests && \
-    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - -b train current-tripleo && \
-    yum update -y && \
-    yum install -y python-gunicorn openstack-ironic-api openstack-ironic-conductor crudini \
-        iproute dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk psmisc \
-        sysvinit-tools mariadb-server genisoimage python-ironic-prometheus-exporter \
-        python-jinja2 python-sushy-oem-idrac && \
-    yum clean all && \
+RUN dnf install -y python3 python3-requests && \
+    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b train current && \
+    dnf update -y && \
+    dnf install -y python3-gunicorn openstack-ironic-api openstack-ironic-conductor crudini \
+        iproute dnsmasq httpd qemu-img iscsi-initiator-utils parted gdisk psmisc \
+        mariadb-server genisoimage python3-ironic-prometheus-exporter \
+        python3-jinja2 python3-sushy-oem-idrac && \
+    dnf clean all && \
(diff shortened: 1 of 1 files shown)


dtantsur:
There seem to be CentOS 8 packages for Train now: https://trunk.rdoproject.org/centos8-train/status_report.html. Time to continue?

Dockerfile
```diff
@@ -1,12 +1,13 @@
-FROM docker.io/centos:centos7
+FROM docker.io/centos:centos8
 
-RUN yum install -y python-requests && \
-    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - -b stein current-tripleo && \
+RUN yum install -y python3 python3-requests && \
```
dtantsur: Do you really need to install python3 explicitly?

Dockerfile
```diff
@@ -1,12 +1,13 @@
...
+    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b train current-tripleo && \
     yum update -y && \
     yum install -y openstack-ironic-api openstack-ironic-conductor crudini \
         iproute iptables dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs psmisc sysvinit-tools \
         mariadb-server python2-chardet genisoimage && \
```
dtantsur: python3-chardet?

Dockerfile
elfosardo: it was necessary until last time :/

Dockerfile
elfosardo: this has been already corrected in a different patch

elfosardo:
@dtantsur python interpreter is not installed by default in Centos 8 base images, so we need to explicitly install it

Dockerfile
```diff
@@ -28,17 +28,16 @@ RUN dd bs=1024 count=2880 if=/dev/zero of=esp.img && \
...
-
-RUN yum install -y python-requests && \
-    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - -b train current-tripleo && \
-    yum update -y && \
-    yum install -y python-gunicorn openstack-ironic-api openstack-ironic-conductor crudini \
-        iproute dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk psmisc \
-        sysvinit-tools mariadb-server genisoimage python-ironic-prometheus-exporter \
-        python-jinja2 && \
-    yum clean all && \
+RUN dnf install -y python3 python3-requests && \
```
dtantsur: Installing python3-requests is supposed to pull in python 3, I wonder why it's not the case..

elfosardo: this is "fun", I was also expecting that but apparently python3-requests doesn't care about python3!
The only dependencies are python3-chardet, python3-idna, python3-pysocks, python3-urllib3

elfosardo:
@dtantsur  I tested the builder conversion to centos 8 but there is an issue when building the ESP image:
```
grub2-mkimage: error: cannot open `/usr/lib/grub/x86_64-efi/linuxefi.mod': No such file or directory
```
despite having the grub2-efi-x64-modules package.
I did a brief reserach and I can see that the package has /usr/lib/grub/x86_64-efi/linux.mod but not the efi one.
So for now I'm keeping centos7 as base for the builder.

elfosardo:
@dtantsur  I tested this locally and it seems working fine
I did a depper research about migrating the builder and I can confirm we can remove linuxefi as it's not part of grub2 upstream, it's just distribution specific, and I don't really think we need it.
I have the final change ready, if we want to remove it, but I'd like also @juliakreger  opinion on this before.

PR #101, opened 2019-10-09
Re-adding Train features

### LINKED ISSUE #102: Flask host and port are static in exporter conf
The change #101 re-introduced the prometheus-exporter app in the ironic image.
This app is based on Flask and in the original configuration its host and running port are statically defined in the startup script runexporterapp.sh
These two parameters should be made customizable and controllable from outside the container.

elfosardo:
This patch adds back the features that were removed because we
were on Stain branch:
- continuous introspection
- ironic-prometheus-exporter
- mdns support

Also fixes #102 

Please merge this and https://github.com/metal3-io/ironic-inspector-image/pull/38 at the same time!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,13 @@ FROM docker.io/centos:centos7
 RUN yum install -y python-requests && \
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - -b train current-tripleo && \
     yum update -y && \
-    yum install -y openstack-ironic-api openstack-ironic-conductor crudini \
-        iproute iptables dnsmasq httpd qemu-img-ev iscsi-initiator-utils parted gdisk ipxe-bootimgs psmisc sysvinit-tools \
+    yum install -y python-gunicorn openstack-ironic-api openstack-ironic-conductor crudini \
+        iproute iptables dnsmasq httpd qemu-img-ev iscsi-initiator-utils \
+        parted gdisk ipxe-bootimgs psmisc sysvinit-tools \
         mariadb-server python2-chardet genisoimage && \
-    yum clean all && rm -rf /var/cache/{yum,dnf}/*
+    yum install -y python-ironic-prometheus-exporter && \
+    yum clean all && \
+    rm -rf /var/cache/{yum,dnf}/*
 
 RUN mkdir /tftpboot && \
     cp /usr/share/ipxe/undionly.kpxe /usr/share/ipxe/ipxe.efi /tftpboot/
@@ -17,6 +20,7 @@ RUN crudini --merge /etc/ironic/ironic.conf < /tmp/ironic.conf && \
 
 COPY ./runironic-api.sh /bin/runironic-api
 COPY ./runironic-conductor.sh /bin/runironic-conductor
+COPY ./runironic-exporter.sh /bin/runironic-exporter
 COPY ./rundnsmasq.sh /bin/rundnsmasq
 COPY ./runhttpd.sh /bin/runhttpd
 COPY ./runmariadb.sh /bin/runmariadb
--- a/runironic-exporter.sh
+++ b/runironic-exporter.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/bash
+
+. /bin/configure-ironic.sh
+
+FLASK_RUN_HOST=${FLASK_RUN_HOST:-"0.0.0.0"}
+FLASK_RUN_PORT=${FLASK_RUN_PORT:-"9608"}
+
+export IRONIC_CONFIG="/etc/ironic/ironic.conf"
+
+exec gunicorn -b ${FLASK_RUN_HOST}:${FLASK_RUN_PORT} -w 4 \
+    ironic_prometheus_exporter.app.wsgi:application
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -39,3 +39,4 @@ interfaces = $IRONIC_IP
 EOF
 
 mkdir -p /shared/html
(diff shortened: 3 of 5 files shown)


Dockerfile
```diff
@@ -17,13 +20,16 @@ RUN crudini --merge /etc/ironic/ironic.conf < /tmp/ironic.conf && \
...
 COPY ./runhttpd.sh /bin/runhttpd
 COPY ./runmariadb.sh /bin/runmariadb
 COPY ./configure-ironic.sh /bin/configure-ironic.sh
 
-# TODO(dtantsur): remove these 2 scripts if we decide to
-# stop supporting running all 2 processes via one entry point.
+# TODO(dtantsur): remove these 3 scripts and squash runexporterapp into
+# runironic-exporter if we decide to stop supporting running all 3 processes
+# via one entry point.
+COPY ./runexporterapp.sh /bin/runexporterapp
```
dtantsur: Let's not re-add this, it was needed for backward compatibility.

runironic-exporter.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
dtantsur: Let's not re-add this file either.

runexporterapp.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+export IRONIC_CONFIG=/etc/ironic/ironic.conf
```
dtantsur: I wonder if this is still needed

runironic.sh
```diff
@@ -26,6 +26,7 @@ mkdir -p /shared/log/ironic
 /usr/bin/ironic-api --log-file  /shared/log/ironic/ironic-api.log &
 
 /bin/runhealthcheck "ironic" &>/dev/null &
+/bin/runexporterapp &
```
dtantsur: Let's not re-add this as well.

runexporterapp.sh
iurygregory: It's necessary to have the variable set, not mandatory to be here.
The Flask app needs to know the `location` that is in the `ironic.conf` see [1]

[1] https://github.com/openstack/ironic-prometheus-exporter/blob/master/ironic_prometheus_exporter/app/exporter.py

elfosardo: I would leave it here, just to be sure it's loaded by the exporter

Dockerfile
elfosardo: actually, shouldn't we leave this ?

runironic.sh
elfosardo: and this, otherwise how do we run the app ?

runexporterapp.sh
```diff
@@ -0,0 +1,6 @@
...
+export FLASK_RUN_HOST=0.0.0.0
```
dhellmann: We need to be able to control this from the outside of the container, so that ironic only binds to the provisioning network interface.

runexporterapp.sh
```diff
@@ -0,0 +1,6 @@
...
+export FLASK_RUN_PORT=9608
```
dhellmann: I expect we'll want to be able to control this with a variable, too.

runexporterapp.sh
elfosardo: can this be done in a follow-up?

runexporterapp.sh
elfosardo: can this be done in a follow-up?

runexporterapp.sh
dhellmann: Sure. Would you open an issue to track that?

russellb: Isn’t this just a trivial one line change to accept an env var?

runexporterapp.sh
russellb: Why? Isn’t it a trivial one line change to accept an env var?

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1221/

runexporterapp.sh
dtantsur: If so, we need to run configure-ironic, otherwise the file may not be present.

runexporterapp.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
dtantsur: Let's merge the contents of this file into runironic-exporter.sh to avoid confusion?

runexporterapp.sh
```diff
@@ -0,0 +1,9 @@
...
+
+FLASK_RUN_HOST=${FLASK_RUN_HOST:-"0.0.0.0"}
+FLASK_RUN_PORT=${FLASK_RUN_PORT:-"9608"}
+
+export IRONIC_CONFIG="/etc/ironic/ironic.conf"
+
+gunicorn -b ${FLASK_RUN_HOST}:${FLASK_RUN_PORT} -w 4 ironic_prometheus_exporter.app.wsgi:application
```
dtantsur: Do we need `exec` here to remove the `bash` process from the picture?

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1231/

runexporterapp.sh
elfosardo: yes, good point

runexporterapp.sh
elfosardo: sure, merged

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1236/

derekhiggins:
> Looks good. Has anyone tried it?

CI has, the failure is unrelated and after the ironic parts,

PR #118, opened 2019-12-04
Adds idrac-redfish interface variations as options

juliakreger:
The idrac hardware type has several interfaces that inherit
the base redfish interface. This provides a mechanism compatability
layer for consumers, and while it ultimatley allows users to pick
and choose, we only need these first two interfaces for now.

Later on we can add idrac-redfish-virtual-media as a boot interface
however right now the Train RPMs used for container images do no thave
that driver so the conductor start-up will fail.


juliakreger:
Can I get a lgtm?

elfosardo:
> Can I get a lgtm?

not enough powers :/

ironic.conf
```diff
@@ -9,8 +9,8 @@ enabled_boot_interfaces = pxe,ipxe,fake,redfish-virtual-media
 enabled_deploy_interfaces = direct,fake
 enabled_hardware_types = ipmi,idrac,irmc,fake-hardware,redfish
 enabled_inspect_interfaces = inspector,idrac,irmc,fake,redfish
```
dtantsur: Ideally, inspection as well.

PR #119, opened 2019-12-04
Add the idrac-redfish-virtual-media boot interface

juliakreger:
Note: This boot interface requires an addtiaional sushy-oem-dell
package which is not yet available as an RPM.


juliakreger:
So what we're missing here is the sushy-oem-dell package. The change to RDO mirrored, although seems not to be packaged yet. I guess we could git clone and install locally initially?

@derekhiggins Seems like an easy compromise to just manually install it from git for now?

elfosardo:
My advice is to install from pypi if the version there is what we need.
If we need a more recent commit, then use pip to install from the git repo, no need for clone.

dtantsur:
The first package will be there once somebody from RDO approves https://review.rdoproject.org/r/#/c/24057/. You can even use the packages from the CI job: https://logs.rdoproject.org/57/24057/9/check/DLRN-rpmbuild/b26763b/buildset/centos/current/

derekhiggins:
> So what we're missing here is the sushy-oem-dell package. The change to RDO mirrored, although seems not to be packaged yet. I guess we could git clone and install locally initially?
> 
> @derekhiggins Seems like an easy compromise to just manually install it from git for now?

Looks like the package is now available in the RDO train repository, so you should be able to just install it
https://trunk.rdoproject.org/train/centos7/current/

derekhiggins:
Looks like we need to either move to rdo/master or back port https://review.opendev.org/#/c/672498/ , trying this out I got an error about no entry point for idrac-redfish-virtual-media

juliakreger:
@derekhiggins I believe we've already backported, it is just not going to make it out into the packaging pipeline for weeks from what I understood from the packaging/release team.

derekhiggins:
> @derekhiggins I believe we've already backported, it is just not going to make it out into the packaging pipeline for weeks from what I understood from the packaging/release team.

Do you mean backported downstream? I don't see it in upstream ironic stable/train which is what this image is using.

dtantsur:
The package is ready in RDO. However, we cannot really backport that feature upstream to Train. Any suggestions? Should we switch to patching code in-place (SIGH)?

elfosardo:
we will have to start using master again then
I would avoid patching code in-place

dtantsur:
I don't think switching to master is currently on the radar, especially because of the troubles it would cause downstream..

elfosardo:
@dtantsur  @juliakreger  @derekhiggins  #125 should add the dependency

dtantsur:
We'll also need https://github.com/metal3-io/ironic-image/pull/100 to be able to switch to master.

elfosardo:
We should be ready to move forward with this now

juliakreger:
It can't be tried until it merges, well aside from the work @derekhiggins did.

ironic.conf
```diff
@@ -5,7 +5,7 @@ default_boot_interface = ipxe
 default_deploy_interface = direct
 default_inspect_interface = inspector
 default_network_interface = noop
-enabled_boot_interfaces = pxe,ipxe,fake,redfish-virtual-media
+enabled_boot_interfaces = pxe,ipxe,fake,redfish-virtual-media,idrac-redfish-virtual-media
 enabled_deploy_interfaces = direct,fake
 enabled_hardware_types = ipmi,idrac,irmc,fake-hardware,redfish
```
dtantsur: This needs `idrac-redfish`, no?

juliakreger: was already present, looks like I need to rebase.

PR #120, opened 2019-12-16
ironic image virtual media fixups

derekhiggins:


Dockerfile
```diff
@@ -41,7 +41,7 @@ COPY --from=builder /tmp/ipxe/src/bin/undionly.kpxe /tftpboot
 COPY --from=builder /tmp/ipxe/src/bin-x86_64-efi/snponly.efi /tftpboot
 COPY --from=builder /tmp/ipxe/src/bin-x86_64-efi/ipxe.efi /tftpboot
 
-COPY --from=builder /tmp/esp.img /httpboot/uefi_esp.img
+COPY --from=builder /tmp/esp.img /tmp/uefi_esp.img
```
elfosardo: could we copy this directly to /shared/html/uefi_esp.img ?

derekhiggins: No, /shared is a shard volume that is shared between all of the ironic containers, it only gets created when the containers are started.

ironic.conf
```diff
@@ -51,3 +50,6 @@ pxe_config_template = $pybasedir/drivers/modules/ipxe_config.template
 tftp_master_path = /shared/tftpboot
 tftp_root = /shared/tftpboot
 uefi_pxe_config_template = $pybasedir/drivers/modules/ipxe_config.template
+
+[redfish]
+use_swift = false
```
juliakreger: oh fun. :(

Dockerfile
```diff
@@ -16,12 +16,16 @@ RUN git clone http://git.ipxe.org/ipxe.git && \
...
 ## be copied in, however that requires more advanced scripting beyond
 ## an MVP.
+## NOTE(derekh): We need to build our own grub image because the one
+## that gets installed by grub2-efi-x64 (/boot/efi/EFI/centos/grubx64.efi)
+## looks for grub.cnf in /EFI/centos, ironic puts it in /boot/grub
 RUN dd bs=1024 count=2880 if=/dev/zero of=esp.img && \
       mkfs.msdos -F 12 -n 'ESP_IMAGE' ./esp.img && \
       mmd -i esp.img EFI && \
       mmd -i esp.img EFI/BOOT && \
+      grub2-mkimage -C xz -O x86_64-efi -p /boot/grub -o /tmp/grubx64.efi boot linux search normal configfile part_gpt btrfs ext2 fat iso9660 loopback test keystatus gfxmenu regexp probe efi_gop efi_uga all_video gfxterm font scsi echo read ls cat png jpeg halt reboot linuxefi && \
```
juliakreger: So the problem doing this is we strip all signing by running grub2-mkimage, at least that is my understanding. @bfournie  can confirm.

derekhiggins: > So the problem doing this is we strip all signing by running grub2-mkimage, at least that is my understanding. @bfournie can confirm.

Ya, looking at the spec for the grub2 package, after running mkimage it run pesign on the file so we are definitely missing that step. I don't know how(or if) this effects secure boot.

juliakreger: I mean, if this basically works now, lets roll with it.

PR #122, opened 2020-01-06
WIP : Wait for expected number of drivers starting API

hardys:
Currently the API will start even if the conductor is still
starting up, which can lead to failures if you try to deploy
nodes as soon as the API is accessible.

### DIFF
--- a/runironic-api.sh
+++ b/runironic-api.sh
@@ -2,4 +2,14 @@
 
 . /bin/configure-ironic.sh
 
+# Wait for ironic to load all expected drivers
+# the DB query returns a string like ["fake-hardware", "idrac", "ipmi"]
+DB_DRIVERS=$(mysql --user=ironic --password=${MARIADB_PASSWORD} --protocol=tcp -r -s -e "use ironic; select drivers from conductors;" | grep -oE '"[^"]+"' | wc -l)
+CONF_DRIVERS=$(crudini --get /etc/ironic/ironic.conf DEFAULT enabled_hardware_types | tr ',' '\n' | wc -l)
+while [ $DB_DRIVERS -lt $CONF_DRIVERS ]; do
+  echo "Waiting for expected drivers $CONF_DRIVERS from conductor"
+  sleep 5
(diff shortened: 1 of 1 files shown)


runironic-api.sh
```diff
@@ -2,4 +2,13 @@
 
 . /bin/configure-ironic.sh
 
+# Wait for ironic to load all expected drivers
+# the DB query returns a string like ["fake-hardware", "idrac", "ipmi"]
+DB_DRIVERS=$(mysql --user=ironic --password=${MARIADB_PASSWORD} --protocol=tcp -r -s -e "use ironic; select drivers from conductors;" | jq '.|length')
```
derekhiggins: Not sure if jq pull in much dependencies but if its an issue we could 
mysql -N -u ironic -p${MARIADB_PASSWORD} -h 127.0.0.1 -e 'select drivers from conductors' ironic | grep -oE '"[^"]+"' | wc -l

runironic-api.sh
```diff
@@ -2,4 +2,13 @@
...
+CONF_DRIVERS=$(crudini --get ironic.conf DEFAULT enabled_hardware_types tr ',' '\n' | wc -l)
```
derekhiggins: we're need to cd to the /etc/ironic directory here or specificy the full path to ironic.conf

also missing a "|" before the tr

runironic-api.sh
hardys: Ack, I don't mind which approach tbh, jq looks to only pull in one additional package beyond glibc:

```
$ sudo yum deplist jq | grep provider | grep -v i686 | sort | uniq
Repodata is over 2 weeks old. Install yum-cron? Or run: yum makecache fast
   provider: glibc.x86_64 2.17-292.el7
   provider: oniguruma.x86_64 5.9.5-3.el7
```

hardys: Went with the grep and removed the jq dependency, thanks!

runironic-api.sh
```diff
@@ -2,4 +2,13 @@
...
+while [ $DB_DRIVERS -lt $CONF_DRIVERS ]; do
+  sleep 5
```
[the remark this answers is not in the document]
hardys: Good point, done

runironic-api.sh
hardys: Thanks, done

derekhiggins:
lgtm

PR #123, opened 2020-01-07
Fix bootloader copy

maelk:
The bootloader was copied to the html folder in runironic-conductor
while some other scripts are used to start ironic. Moving the copy
to runhttpd as the file is served by httpd.


maelk:
The bug was introduced in https://github.com/metal3-io/ironic-image/commit/fb42b35008d7c44007a347ced203016154391a27 . The copy function is in runironic-conductor.sh while we use runironic.sh in metal3-dev-env

PR #125, opened 2020-01-08
Add python-sushy-oem-idrac pkg for vmedia boot interface

elfosardo:
This is necessary for #119


elfosardo:
@dtantsur  that's ok, I'd rather merge this first and then centos 8, also because the CI in dev-env is not working at the moment :/

PR #126, opened 2020-01-09
Use env in shebang for bash scripts

elfosardo:
Standardize bash command location relying on env to detect the
default system binary instead of giving its path as assumed.

### DIFF
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/usr/bin/env bash
 
 . /bin/ironic-common.sh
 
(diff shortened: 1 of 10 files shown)


dtantsur:
What are you trying to achieve by that? Would be good to mention in the commit message.

elfosardo:
@dtantsur  I added a description and a message to the commit

dtantsur:
This is useful when you expect many entities of the same binary. I don't think we're going to have many bashes?

elfosardo:
@dtantsur  I hope we won't, this should help using a default system one
another solution could be choose one binary path and stick with that in all scripts
I do prefer the env solution, but I don't mind the second option, as far as we're consistent with the choice

hardys:
This is only cosmetic, since we know the bash path on the base image/os we expect to support.

I'm fine with it, but tbh I'd see far greater value in a change like make sure `set -e` is specified consistently in all entrypoint scripts?

PR #134, opened 2020-01-23
Use master for tripleo repo

elfosardo:
this is needed for #119

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ RUN dd bs=1024 count=2880 if=/dev/zero of=esp.img && \
 FROM docker.io/centos:centos8
 
 RUN dnf install -y python3 python3-requests && \
-    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b train current && \
+    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current && \
     dnf update -y && \
     dnf install -y python3-gunicorn openstack-ironic-api openstack-ironic-conductor crudini \
         iproute dnsmasq httpd qemu-img iscsi-initiator-utils parted gdisk psmisc \


hardys:
Doesn't this mean we're using something which essentially hasn't passed any of the promotion jobs at all?

https://dashboards.rdoproject.org/rdo-dev

derekhiggins:
> Doesn't this mean we're using something which essentially hasn't passed any of the promotion jobs at all?
> 
> https://dashboards.rdoproject.org/rdo-dev

Yup, we should probably switch current to current-tripleo

hardys:
> Yup, we should probably switch current to current-tripleo

IIUC we're using current atm because current-tripleo doesn't exist for centos8 yet, but if we move to master instead of current that's not tested at all, right?

elfosardo:
@derekhiggins @hardys AFAIK current-tripleo doesn't exist for CentOS 8 yet

hardys:
> @derekhiggins @hardys AFAIK current-tripleo doesn't exist for CentOS 8 yet

Sure so we can use current in the meantime (which IIRC has passed some non-TripleO CI), my concern is (if I'm remembering the hierarchy of RDO pins correctly) that master is the least tested option, so basically we're switching to an untested build which may not work, as a workaround for the fact the RDO promotion is a few days behind trunk?

derekhiggins:
> so basically we're switching to an untested build which may not work, as a workaround for the fact the RDO promotion is a few days behind trunk?

I don't think that is the motivation here, we are switching from train to master for a feature that wont be backported to train. At least that my read on whats going on.

elfosardo:
@hardys  my understanding is that the rpms are tested, you can actually see the results here:
https://trunk.rdoproject.org/centos8-master/status_report.html
The only thing with current is that it contains the latest packages that can be built, even if some other packages are failing to build.
That means that the promotion has still not happened (otherwise we should have current-tripleo).
There's also to consider the Openstack CI behind all this.

hardys:
> > so basically we're switching to an untested build which may not work, as a workaround for the fact the RDO promotion is a few days behind trunk?
> 
> I don't think that is the motivation here, we are switching from train to master for a feature that wont be backported to train. At least that my read on whats going on.

Ah yes you're right, thanks - looking at the comments in https://github.com/metal3-io/ironic-image/pull/119 makes that clear.

I know we discussed doing this some months ago for some previous Ironic features, and ultimately decided against it, seems like something we should probably raise on the community ML before proceeding?

elfosardo:
@hardys I actually added a post in the ML around 30 minutes ago :)

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1441/

derekhiggins:
> Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1441/

Looking the conductor logs, looks like the nodes went to "clean failed' for some reason

elfosardo:
if we don't want to use current, we can use alternative repos with packages that passed CI as shown in #135

maelk:
I'm not too familiar with RDO. If we were to use master, that is untested, so we could end up with a broken image, while the risks are reduced if we use the alternative repo, is this correct ? What is the downside of using the other repo ? @elfosardo

PR #141, opened 2020-02-21
Use tested packages

elfosardo:
The current-tripleo repo is now available for centos8 and we
should use it.


elfosardo:
> I think `current` is also tested, but via a different CI.

the current build is whatever latest can be built as far as the build doesn't fail
current-tripleo goes through the actual tripleo ci
https://docs.openstack.org/tripleo-docs/latest/ci/stages-overview.html

PR #142, opened 2020-03-03
Reduce distro specific packages

Xenwar:
As we have it now, ironic image uses centos-8 as a base image.
Using a different base image does require making a lot of changes.

The purpose of this commit is reducing the dependency on a any specific distribution.
With respect to ironic, it installs it from the repository. It also removes distro specific commands.

Ideally, the image build process can be made distro agnostic by installing all packages using pip3. However, the purpose of this commit is on reducing the dependency.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,15 +30,28 @@ RUN dd bs=1024 count=2880 if=/dev/zero of=esp.img && \
 
 FROM docker.io/centos:centos8
 
-RUN dnf install -y python3 python3-requests && \
+RUN dnf install -y gcc python3 python3-requests python3-devel && \
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current && \
     dnf update -y && \
-    dnf install -y python3-gunicorn openstack-ironic-api openstack-ironic-conductor crudini \
+    dnf install -y python3-gunicorn crudini OpenIPMI ipmitool \
         iproute dnsmasq httpd qemu-img iscsi-initiator-utils parted gdisk psmisc \
-        mariadb-server genisoimage python3-ironic-prometheus-exporter \
+        mariadb-server genisoimage \
         python3-jinja2 python3-sushy-oem-idrac && \
     dnf clean all && \
     rm -rf /var/cache/{yum,dnf}/*
+
+RUN /usr/bin/pip3 install pymysql
+RUN /usr/bin/pip3 install git+https://opendev.org/openstack/ironic.git@master
+
+RUN /usr/bin/pip3 install ironic-prometheus-exporter
+
+# Create soft links in order to avoid changing multiple files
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -30,23 +30,40 @@ RUN dd bs=1024 count=2880 if=/dev/zero of=esp.img && \
...
-        mariadb-server genisoimage python3-ironic-prometheus-exporter \
+        mariadb-server genisoimage \
         python3-jinja2 python3-sushy-oem-idrac && \
     dnf clean all && \
     rm -rf /var/cache/{yum,dnf}/*
 
+RUN /usr/bin/pip3 install pymysql
+RUN /usr/bin/pip3 install git+https://opendev.org/openstack/ironic.git@master
+
+RUN /usr/bin/pip3 install ironic-prometheus-exporter
```
dtantsur: Please no pip install by default. Pip install is a terrible idea most of the time, this pulls in an absolutely untested combination of source packages and their dependencies.

Dockerfile
```diff
@@ -30,23 +30,40 @@ RUN dd bs=1024 count=2880 if=/dev/zero of=esp.img && \
 
 FROM docker.io/centos:centos8
 
-RUN dnf install -y python3 python3-requests && \
+RUN dnf install -y gcc python3 python3-requests python3-devel && \
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current && \
     dnf update -y && \
-    dnf install -y python3-gunicorn openstack-ironic-api openstack-ironic-conductor crudini \
+    dnf install -y python3-gunicorn crudini OpenIPMI ipmitool \
```
dtantsur: You're missing a lot of dependencies here.

Dockerfile
```diff
@@ -60,12 +77,14 @@ COPY ./runmariadb.sh /bin/runmariadb
 COPY ./configure-ironic.sh /bin/configure-ironic.sh
 COPY ./ironic-common.sh /bin/ironic-common.sh
 
-# TODO(dtantsur): remove this script when we stop supporting running both
-# API and conductor processes via one entry point.
+# TODO(dtantsur): remove these 2 scripts if we decide to
+# stop supporting running all 2 processes via one entry point.
+COPY ./runhealthcheck.sh /bin/runhealthcheck
```
dtantsur: Please don't reintroduce it.

dhellmann:
I'm not sure why we care about supporting different base images, but if we do perhaps we should use separate Dockerfiles for each of them?

dtantsur:
Yeah, I can imagine something like `Dockerfile.source` with packages from pip. But I'd keep the main Dockerfile with packages.

Xenwar:
Thank you all for your comments regarding the reason for the current setup.
Based on your inputs, I am closing this PR.

PR #149, opened 2020-04-21
 Use upgrade instead of deprecated update

elfosardo:
Also convert yum to dnf in the builder.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 ## Note: we are pinning to a specific commit for reproducible builds.
 ## Updated as needed.
 FROM docker.io/centos:centos8 AS builder
-RUN yum install -y gcc git make genisoimage xz-devel grub2 grub2-efi-x64-modules shim dosfstools mtools
+RUN dnf install -y gcc git make genisoimage xz-devel grub2 grub2-efi-x64-modules shim dosfstools mtools
 WORKDIR /tmp
 COPY . .
 RUN git clone http://git.ipxe.org/ipxe.git && \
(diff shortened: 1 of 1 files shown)


elfosardo:
@dtantsur the base image (the FROM field) should already contain the latest packages with all the security fix, if it doesn't the issue is bigger than just for our containers.
Using any type of general update/upgrade is not recommended when building containers, to avoid security issues ourselves, we should upgrade specific packages with specific versions, but again that is a bigger problem at the origin.
The main reason to not use a generic update/upgrade is to avoid ending up every time with a different image, also there could be a potential discrepancy between different containers in the same environment, so for example the ironic image could contain different packages from the ironic-inspector image, and so on.

P.S. Since we use dnf, the 'update' command is deprecated in favor of 'upgrade'.

dtantsur:
>  the base image (the FROM field) should already contain the latest packages with all the security fix, if it doesn't the issue is bigger than just for our containers.

Well, the last time I checked it did not. And with running update worse case we'll get the same result. Particularly,

> to avoid ending up every time with a different image

is impossible if 

>  the base image (the FROM field) should already contain the latest packages

elfosardo:
@dtantsur I'm not sure I understand what you mean with "it's impossible"
It is possible if you build an image today for one service and another image tomorrow for a different service. If there are updates in packages during the night, the second image will take the new packages, while the first one will stay with the old ones, ending up with two different images in the same environment.
If the base images we're using do not get the latest security fixes then the problem is not just on us, it's in the origin of the base image.
If we want to install security fixes or upgrade specific packages, then we should choose which packages to install/upgrade, not just upgrade everything indiscriminately.

dtantsur:
> If the base images we're using do not get the latest security fixes then the problem is not just on us

Yes, it is. We're shipping them.

> If we want to install security fixes or upgrade specific packages, then we should choose which packages to install/upgrade, not just upgrade everything indiscriminately.

I still don't understand why. I don't see any problem in situation when

> so for example the ironic image could contain different packages from the ironic-inspector image, and so on.

PR #153, opened 2020-05-04
Allow any IPv6 address to be configured for provisioning, but link local scope

mvazquezc:
This PR fixes an issue when using routable IPv6 addresses for the provisioning network. The previous grep command only matched local unicast addresses, this new grep will match any IPv6 address, but the link local scope.

The httpd container keeps waiting for "PROV_IFACE" to be configured otherwise.

### DIFF
--- a/ironic-common.sh
+++ b/ironic-common.sh
@@ -14,7 +14,7 @@ function wait_for_interface_or_ip() {
   else
     until [ ! -z "${IRONIC_IP}" ]; do
       echo "Waiting for ${PROVISIONING_INTERFACE} interface to be configured"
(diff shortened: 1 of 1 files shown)


celebdor:
Looks good. Let's wait for the CI

jparrill:
@dhellmann seems that the integration tests passes, could you please merge this one? we're blocked

maelk:
This patch fell in the CI gap, the tests ran without using the image built. The gap should be fixed now. The fix for the bug introduced here is https://github.com/metal3-io/ironic-image/issues/156

jparrill:
@maelk do you know if the new image was generated and uploaded? or we need to wait for the new merge?

PR #156, opened 2020-05-07
Fix the IP address parsing in ironic-common.sh

### LINKED ISSUE #155: IP fetched with prefix when configuring ironic
After this change : https://github.com/metal3-io/ironic-image/commit/e4df69305733355e9e77c78c79c0f84c65c5c3db when runnning 
```
[root@ci-test-vm-20200507050740 /]# ip -br add show scope global up dev "provisioning" | awk '{print $3; exit}'
172.22.0.1/24
```
The prefix should be stripped. This prevents ironic from starting properly

maelk:
The prefix was captured with the IP address. This change strips it

Fixes #155


Jaakko-Os:
looks good

PR #157, opened 2020-05-07
Allows users to configure any non-link-local IPV6 address for the provisioning interface

mvazquezc:
This fix introduces the fix created for metal3-io ironic-image https://github.com/metal3-io/ironic-image/pull/156/

Thank @maelk for fixing our previous fix!


mvazquezc:
I meant to send this to other repo. My bad.

PR #159, opened 2020-05-19
Add ibmc driver

IamFive:
enable ibmc hardware type

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,7 @@ RUN dnf install -y python3 python3-requests && \
     dnf install -y python3-gunicorn openstack-ironic-api openstack-ironic-conductor crudini \
         iproute dnsmasq httpd qemu-img iscsi-initiator-utils parted gdisk psmisc \
         mariadb-server genisoimage python3-ironic-prometheus-exporter \
-        python3-jinja2 python3-sushy-oem-idrac && \
+        python3-jinja2 python3-sushy-oem-idrac python3-ibmcclient && \
     dnf clean all && \
     rm -rf /var/cache/{yum,dnf}/*
 
--- a/ironic.conf
+++ b/ironic.conf
@@ -7,12 +7,12 @@ default_inspect_interface = inspector
 default_network_interface = noop
 enabled_boot_interfaces = pxe,ipxe,fake,redfish-virtual-media,idrac-redfish-virtual-media
 enabled_deploy_interfaces = direct,fake
-enabled_hardware_types = ipmi,idrac,irmc,fake-hardware,redfish
+enabled_hardware_types = ipmi,idrac,irmc,fake-hardware,redfish,ibmc
 enabled_inspect_interfaces = inspector,idrac,irmc,fake,redfish
-enabled_management_interfaces = ipmitool,idrac,irmc,fake,redfish,idrac-redfish
-enabled_power_interfaces = ipmitool,idrac,irmc,fake,redfish,idrac-redfish
+enabled_management_interfaces = ipmitool,idrac,irmc,fake,redfish,idrac-redfish,ibmc
+enabled_power_interfaces = ipmitool,idrac,irmc,fake,redfish,idrac-redfish,ibmc
 enabled_raid_interfaces = no-raid,irmc,agent,fake
-enabled_vendor_interfaces = ipmitool,no-vendor,idrac,fake
+enabled_vendor_interfaces = ipmitool,no-vendor,idrac,fake,ibmc
 rpc_transport = json-rpc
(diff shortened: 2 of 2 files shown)


maelk:
looks good to me

maelk:
quay.io is down, so no tests are passing today

zaneb:
Given that the iBMC driver is [deprecated in Ironic](https://review.opendev.org/703465), I'm not sure that this makes sense for us?

juliakreger:
@zaneb See: https://review.opendev.org/#/c/723225/ for it being un-deprecated.

zaneb:
@juliakreger ah, thanks. That commit only reverts 1/4 of the `supported = False` things and, crucially, fails to remove the deprecation notice from the [docs](https://docs.openstack.org/ironic/latest/admin/drivers/ibmc.html). Here is a fix: https://review.opendev.org/729355

juliakreger:
@zaneb Doh! Thanks!

IamFive:
@maelk I got a question. Will `driver-requirements.txt` auto installed? Should we update docker file, and install vendor driver dependent python libs?

maelk:
I am not sure which file you are mentioning, do you have a pointer ? For now we install ironic through a package manager, dnf, so the dependencies should be taken care of, afaik

IamFive:
The driver-requirements.txt i mentioned is : https://github.com/openstack/ironic/blob/master/driver-requirements.txt .

The vendor dependencies is not auto installed in the docker image. So, I update the docker file, https://github.com/metal3-io/ironic-image/pull/159/files#diff-3254677a7917c6c01f55212f86c57fbfR43-R45.

Dockerfile
```diff
@@ -40,6 +40,9 @@ RUN dnf install -y python3 python3-requests && \
     dnf clean all && \
     rm -rf /var/cache/{yum,dnf}/*
 
+# install vendor driver libs
+RUN /usr/bin/pip3 install python-ibmcclient>=0.2.3
```
elfosardo, line 44: please we should not install anything using pip but only rely on packages as for other requirements

IamFive: hi @elfosardo , `python-ibmcclient` is available now at: https://trunk.rdoproject.org/centos8/component/baremetal/45/eb/45ebd10ea96a436f0ba6bd06dd393df26d333606_168399e5/

IamFive:
hi @maelk , python-ibmcclient lib has been added to rdopackages, may we continue integration steps now.

Dockerfile
```diff
@@ -40,6 +40,9 @@ RUN dnf install -y python3 python3-requests && \
     dnf clean all && \
     rm -rf /var/cache/{yum,dnf}/*
 
+# install vendor driver libs
+RUN dnf install python3-ibmcclient
+
```
maelk, line 45: please include this as part of the previous RUN command. It is good practice to limit the number of RUN and in addition you would benefit from the cache cleanup.

maelk:
In addition, Travis failed with :
```
No match for argument: python3-ibmcclient

Error: Unable to find a match: python3-ibmcclient

The command '/bin/sh -c dnf install python3-ibmcclient' returned a non-zero code: 1

Makefile:8: recipe for target 'docker' failed

make: *** [docker] Error 1

The command "make $TARGET" exited with 2.
```

IamFive:
hi @maelk , I am not sure how rdopackage repo is added. `python-ibmcclient` lib was published yestoday(2020-06-01), you can find it here: https://trunk.rdoproject.org/centos8/component/baremetal/45/eb/45ebd10ea96a436f0ba6bd06dd393df26d333606_168399e5/.

To include new added libs, this repo(http://trunk.rdoproject.org/centos8-master/current/delorean.repo) should be replaced under /etc/yum.repos.d.

Hi, @dtantsur , may u help give a comment about this issue.

IamFive:
sorry, my bad, did not see the `dnf clean all` thing. It should be ok now.

elfosardo:
@IamFive we switched from current to current-tripleo repositories, that's why the package can't be found
The package needs to be promoted to current-tripleo before being accessible by the image

IamFive:
hi all, python3-ibmcclient is available on [current-tripleo](https://trunk.rdoproject.org/centos8-master/component/baremetal/current-tripleo/) now.  May we start merge this PR.

IamFive:
@zaneb  also need an approved

zaneb:
> @zaneb also need an approved

I *know* I'm not cool enough for that because there's a [list](https://github.com/metal3-io/ironic-image/blob/master/OWNERS). Also AIUI the whole idea is to get two different people.

PR #165, opened 2020-06-22
Add ability to set arbitrary ironic configs

derekhiggins:
As more use cases are using this image and may need to tweak
various config options. Add the method to set arbitary options
rather then having to add a new option for each case.

Potential solution (the ironic part) to https://github.com/metal3-io/ironic-image/issues/143

Another alternative method might be to mount a partial ironic.conf to be merged into /etc/ironic/ironic.conf , I've leaned in the direction of the env variable as its more consistent
with that we currently have

Marking as WIP until we discuss whats best.

### DIFF
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -43,5 +43,9 @@ endpoint_override = http://${IRONIC_URL_HOST}:5050
 endpoint_override = http://${IRONIC_URL_HOST}:6385
 EOF
 
+# oslo.config also supports Config Opts From Environment, log them
+echo '# Options set from Environment variables' | tee /etc/ironic/ironic.extra
+env | grep "^OS_" | tee -a /etc/ironic/ironic.extra
+
 mkdir -p /shared/html
 mkdir -p /shared/ironic_prometheus_exporter
--- a/README.md
+++ b/README.md
@@ -35,3 +35,4 @@ The following environment variables can be passed in to customize run-time funct
 - HTTP_PORT - port used by http server (default 80)
 - DHCP_RANGE - dhcp range to use for provisioning (default 172.22.0.10-172.22.0.100)
 - MARIADB_PASSWORD - The database password
+- OS_<section>_\_<name>=<value> - This format can be used to set arbitary ironic config options


dtantsur:
Are you aware of https://specs.openstack.org/openstack/oslo-specs/specs/rocky/config-from-environment.html? Maybe it can be reused?

derekhiggins:
> Are you aware of https://specs.openstack.org/openstack/oslo-specs/specs/rocky/config-from-environment.html? Maybe it can be reused?

thanks, I wasn't aware of this, I tried it out and it works, the only down side I can see is that somebody debugging the image might look at ironic.conf and not realise that it doesn't represent
all the config options, so I propose we also log any env options to the file system so that they would
be seen. something like this in configure-ironic.sh
`env | grep "^OS_" > /etc/ironic/ironic.env  `

dhellmann:
> > Are you aware of https://specs.openstack.org/openstack/oslo-specs/specs/rocky/config-from-environment.html? Maybe it can be reused?
> 
> thanks, I wasn't aware of this, I tried it out and it works, the only down side I can see is that somebody debugging the image might look at ironic.conf and not realise that it doesn't represent
> all the config options, so I propose we also log any env options to the file system so that they would
> be seen. something like this in configure-ironic.sh
> `env | grep "^OS_" > /etc/ironic/ironic.env`

oslo.config can be made to log all of the known configuration options and their values. Apps typically do that by invoking `log_opt_values()` on startup. Does Ironic do that?

derekhiggins:
> > > Are you aware of https://specs.openstack.org/openstack/oslo-specs/specs/rocky/config-from-environment.html? Maybe it can be reused?
> > 
> > 
> > thanks, I wasn't aware of this, I tried it out and it works, the only down side I can see is that somebody debugging the image might look at ironic.conf and not realise that it doesn't represent
> > all the config options, so I propose we also log any env options to the file system so that they would
> > be seen. something like this in configure-ironic.sh
> > `env | grep "^OS_" > /etc/ironic/ironic.env`
> 
> oslo.config can be made to log all of the known configuration options and their values. Apps typically do that by invoking `log_opt_values()` on startup. Does Ironic do that?

Yes it does, but sometimes in a long running ironic container the head of the log may be gone, also I typically exec into the container and usually look around at the setup.

configure-ironic.sh
```diff
@@ -43,5 +43,9 @@ endpoint_override = http://${IRONIC_URL_HOST}:5050
 endpoint_override = http://${IRONIC_URL_HOST}:6385
 EOF
 
+# oslo.config also supports Config Opts From Environment, log them
+echo '# Options set from Environment variables' | tee /etc/ironic/ironic.conf
+env | grep "^OS_" | tee -a /etc/ironic/ironic.conf
```
dhellmann, line 48: This is adding shell environment variables directly to the config file. Could that cause parse issues? Should they be added as comments?

derekhiggins: thanks, that was a typo, I meant to add them to a different file.

dhellmann: Ah, cool, I assumed you wanted everything in one place for debugging. A separate file seems even safer, though.

PR #167, opened 2020-06-25
Make ironic timeouts customizable

furkatgofurov7:
**What this PR does / why we need it:**
This commit adds some parameters for the ironic configuration

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,3 +36,9 @@ The following environment variables can be passed in to customize run-time funct
 - DHCP_RANGE - dhcp range to use for provisioning (default 172.22.0.10-172.22.0.100)
 - MARIADB_PASSWORD - The database password
 - OS_<section>_\_<name>=<value> - This format can be used to set arbitary ironic config options
+
+The ironic configuration can be overridden by various environment variables. The following can serve as an example:
+- OS_CONDUCTOR__DEPLOY_CALLBACK_TIMEOUT=4800 - timeout (seconds) to wait for a callback from a deploy ramdisk
+- OS_CONDUCTOR__INSPECT_TIMEOUT=1800 - timeout (seconds) for waiting for node inspection
+- OS_CONDUCTOR__CLEAN_CALLBACK_TIMEOUT=1800 - timeout (seconds) to wait for a callback from the ramdisk doing the cleaning
+- OS_PXE__BOOT_RETRY_TIMEOUT=1200 - timeout (seconds) to enable boot retries.


maelk:
I didn't see that it was set this way in ironic.conf, sorry, discard what I said, since we anyways set our own settings

configure-ironic.sh
```diff
@@ -13,6 +13,15 @@ IRONIC_FAST_TRACK=${IRONIC_FAST_TRACK:-true}
 # Whether cleaning disks before and after deployment
 IRONIC_AUTOMATED_CLEAN=${IRONIC_AUTOMATED_CLEAN:-true}
 
+# Timeout (seconds) to wait for a callback from a deploy ramdisk
+IRONIC_DEPLOY_CALLBACK_TIMEOUT=${IRONIC_DEPLOY_CALLBACK_TIMEOUT:-"4800"}
```
maelk, line 17: Considering the comment you pointed out, we might also want to put the pxe deploy timeout customized here with the same note about its dependency

furkatgofurov7: @maelk this is the comment I was referring to in the [ironic.conf file](https://github.com/metal3-io/ironic-image/blob/e1d39417015f5a7799a42a91a36f7005d872b562/ironic.conf#L55) and I have made it customizable as you suggested.

configure-ironic.sh
```diff
@@ -13,6 +13,20 @@ IRONIC_FAST_TRACK=${IRONIC_FAST_TRACK:-true}
...
```
elfosardo, line 17: the options look good, just the values seem a bit higher than expected, how did you get them?

furkatgofurov7: @elfosardo I was checking [ironic.conf file](https://github.com/metal3-io/ironic-image/blob/e1d39417015f5a7799a42a91a36f7005d872b562/ironic.conf#L31) and the values were set in that file. I guess it sets this value in purpose to make sure [boot_retry_timeout](https://github.com/metal3-io/ironic-image/blob/e1d39417015f5a7799a42a91a36f7005d872b562/ironic.conf#L56) is 3 times less than that variable as described in the comments. Maybe, we can set it to **"3600"** if it is higher than expected?

elfosardo: I see, now I remember why we set the value like that. That's fine then, thanks!

furkatgofurov7: Thank you for reviewing.

dhellmann:
I feel a bit like we're reimplementing confd. :-)

These are oslo.config settings, right? In other places we used the automatically checked environment variables to pass values right to the daemon. Would that approach work here, too? Could we document the relevant variables names instead of modifying the config file?

maelk:
That's a very good point. This change could instead be a readme explaining how to override the settings in the ironic config.

derekhiggins:
> That's a very good point. This change could instead be a readme explaining how to override the settings in the ironic config.

Something like this could work
https://github.com/metal3-io/ironic-image/pull/165

README.md
```diff
@@ -35,3 +35,9 @@ The following environment variables can be passed in to customize run-time funct
 - HTTP_PORT - port used by http server (default 80)
 - DHCP_RANGE - dhcp range to use for provisioning (default 172.22.0.10-172.22.0.100)
 - MARIADB_PASSWORD - The database password
+
+The following environment variables can be used to override the settings in the ironic configuration file:
```
maelk, line 39: I would make it less specific I guess, saying that configuration can be overriden by environment variables. Then say as an example the following (and show the config options so that it is understandable how to extend this to other configuration options)

furkatgofurov7: Fixed, thank you for feedback

furkatgofurov7:
> I feel a bit like we're reimplementing confd. :-)
> 
> These are oslo.config settings, right? In other places we used the automatically checked environment variables to pass values right to the daemon. Would that approach work here, too? Could we document the relevant variables names instead of modifying the config file?

@dhellmann thanks for the suggestion, I have addressed your comments and it is now only in the readme document.

derekhiggins:
> > I feel a bit like we're reimplementing confd. :-)
> > These are oslo.config settings, right? In other places we used the automatically checked environment variables to pass values right to the daemon. Would that approach work here, too? Could we document the relevant variables names instead of modifying the config file?
> 
> @dhellmann thanks for the suggestion, I have addressed your comments and it is documented now in the readme by following the format from #165

I think we should just merge #165, I don't see any reason why these Timeout values should be special and mentioned in the config.

dhellmann:
I don't have a problem with listing some variables that are commonly useful to override. Maybe we can also describe the general way to build the right name for other variables? That could be a separate PR.

This one needs a rebase.

furkatgofurov7:
> > > I feel a bit like we're reimplementing confd. :-)
> > > These are oslo.config settings, right? In other places we used the automatically checked environment variables to pass values right to the daemon. Would that approach work here, too? Could we document the relevant variables names instead of modifying the config file?
> > 
> > 
> > @dhellmann thanks for the suggestion, I have addressed your comments and it is documented now in the readme by following the format from #165
> 
> I think we should just merge #165, I don't see any reason why these Timeout values should be special and mentioned in the config.



> I don't have a problem with listing some variables that are commonly useful to override. Maybe we can also describe the general way to build the right name for other variables? That could be a separate PR.
> 
> This one needs a rebase.

These are examples of timeout values and how can they be overridden. And also we are not mentioning them in the config but README.md file.

/test-integration

furkatgofurov7:
Hi @dhellmann, is there anything that needs to be changed/improved in this PR, or we can move forward with it?

furkatgofurov7:
Hi @derekhiggins! is there anything that needs to be improved in this PR or we can proceed with it?

PR #170, opened 2020-07-03
WIP: Reload httpd on certificate change

felixest:
When the certificates used by httpd change (e.g., due to automatic
certificate renewal by cert-manager). We want to reload httpd so it uses
the new certificates. This is achieved by using inotify to get informed
about changes of the files.

As the secrets in k8s are just symlinks to the real files, the event we
see when the certificates are renewed is DELETE_SELF.

**WIP:** This only reloads the httpd for ironic-api. The changes from @maelk need to be integrated first.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,13 +30,14 @@ RUN dd bs=1024 count=2880 if=/dev/zero of=esp.img && \
 
 FROM docker.io/centos:centos8
 
-RUN dnf install -y python3 python3-requests && \
+RUN dnf install -y python3 python3-requests epel-release && \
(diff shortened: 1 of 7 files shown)


dhellmann:
This feels like the sort of thing I would expect to achieve by restarting the pod. In our case, we have the ironic database to consider, so maybe that's you're doing it within the pod instead?

Are we running a health check on the pod? Will that report the service as being down during the reload?

maelk:
/remove-lifecycle stale
I think this should still be addressed, to automate the handling of the rotation when deployed with cert-manager

PR #171, opened 2020-07-05
Convert httpd and ironic config to be rendered by jinja

steveb:
Contains the following changes:
- Make jinja render one-liner a real script
- Remove DNSMASQ_EXCEPT_INTERFACE
- Convert httpd config to static conf + jinja
- Make ironic.conf jinja template driven

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,6 +51,7 @@ COPY ./ironic.conf /tmp/ironic.conf
 RUN crudini --merge /etc/ironic/ironic.conf < /tmp/ironic.conf && \
     rm /tmp/ironic.conf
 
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -51,6 +51,7 @@ COPY ./ironic.conf /tmp/ironic.conf
 RUN crudini --merge /etc/ironic/ironic.conf < /tmp/ironic.conf && \
     rm /tmp/ironic.conf
 
+COPY ./jinjarender.py /bin/jinjarender
```
elfosardo, line 54: let's change permissions to make this executable

steveb: It is preserving the execute bit from the source file, like the other scripts in this Dockerfile

PR #173, opened 2020-07-08
Set power_off = !fast_track in [inspector]

dtantsur:
We should only set power_off to false when fast_track is true.

### DIFF
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -15,6 +15,16 @@ IRONIC_AUTOMATED_CLEAN=${IRONIC_AUTOMATED_CLEAN:-true}
 
 wait_for_interface_or_ip
 
+if [[ $IRONIC_FAST_TRACK == true ]]; then
+    INSPECTOR_POWER_OFF=false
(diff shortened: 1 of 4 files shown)


configure-ironic.sh
```diff
@@ -15,6 +15,16 @@ IRONIC_AUTOMATED_CLEAN=${IRONIC_AUTOMATED_CLEAN:-true}
 
 wait_for_interface_or_ip
 
+if [[ $IRONIC_FAST_TRACK == true ]]; then
+    INSPECTOR_POWER_OFF=false
+    # TODO(dtantsur): ipa-api-url should be populated by ironic itself, but
+    # it's not yet, so working around here.
+    INSPECTOR_EXTRA_ARGS=" ipa-api-url=http://${IRONIC_URL_HOST}:6385"
```
elfosardo, line 22: nit: extra space inside the quotes not needed

dhellmann: Let's fix that in a follow-up after the CI jobs are fixed.

PR #175, opened 2020-07-10
Framework for applying local patches

dtantsur:


Dockerfile
```diff
@@ -47,6 +47,13 @@ COPY --from=builder /tmp/ipxe/src/bin-x86_64-efi/ipxe.efi /tftpboot
 
 COPY --from=builder /tmp/esp.img /tmp/uefi_esp.img
 
+COPY patches /tmp/patches
+RUN cd /usr/lib/python3.6/site-packages && \
+    for pfile in $(ls /tmp/patches); do \
```
elfosardo, line 52: I'm not really a fan of looping on ls output, wondering if you could use a simple glob instead, like ```for pfile in /tmp/files/*``` and then just put ```filterdiff "$file"```
only problem would be with an empty dir, you should set ```shopt -s nullglob``` right before

dtantsur: Well, exactly, you either need to use `ls` or change shell options. Why do you think the latter is better?

elfosardo: in short, when using ``ls`` to get a list of filenames, or in general when parsing the output of ls, things can fail bad, depending on which system the command runs on, and which characters compose the filenames.
Maybe the filenames will only contain the right characters, but we'll never know in advance, unless we impose certain rules, that's why parsing the ``ls`` output should be discouraged.
The nullglob shell option is non-standard but ideal in this case, as you can use a glob directly, getting the correct result, and you can always change it back after the loop.
Anyway, do not want to start a big discussion around it, if this gets approved as its, that's totally fine.

PR #178, opened 2020-07-22
Update http_basic configuration

iurygregory:
Let's use a variable that will define the auth_strategy for ironic
`IRONIC_AUTH_STRATEGY` that is default to noauth. (For more details 
see discussion in https://github.com/metal3-io/ironic-image/pull/168 ) 

The variables for username and password are default to "", if you
are using `http_basic` they need to be set (will fail if one of the
variables is not set).

Update the json_rpc configuration, we will use username/password
instead of http_basic_username and http_basic_password, see [1]

[1] https://review.opendev.org/#/c/742474/

### DIFF
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -2,9 +2,10 @@
 
 . /bin/ironic-common.sh
 
-USE_HTTP_BASIC=${USE_HTTP_BASIC:-false}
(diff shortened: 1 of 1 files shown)


iurygregory:
Clossing in favor of #182

PR #180, opened 2020-07-24
Limit unauthenticated json-rpc to localhost interface

zaneb:
By default, ironic-conductor currently binds to any IP address, so it is
accessible outside of the pod. When no authentication method is used for
json-rpc, bind only to localhost so that only other containers in the
same pod can connect to the json-rpc server.

This will break any deployments that put ironic-conductor and ironic-api
in separate pods but do not specify basic_auth. This is probably a good
thing.

### DIFF
--- a/runironic.sh
+++ b/runironic.sh
@@ -2,6 +2,8 @@
 
 . /bin/configure-ironic.sh
 
+crudini --set /etc/ironic/ironic.conf DEFAULT host localhost
+
 ironic-dbsync --config-file /etc/ironic/ironic.conf upgrade
 
 # Remove log files from last deployment
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -71,7 +71,9 @@ if [ "$USE_HTTP_BASIC" = "true" ]; then
 
 	crudini --set /etc/ironic/ironic.conf DEFAULT auth_strategy http_basic
 	crudini --set /etc/ironic/ironic.conf DEFAULT http_basic_auth_user_file /shared/htpasswd-ironic
+
 	crudini --set /etc/ironic/ironic.conf json_rpc auth_strategy http_basic
+	crudini --del /etc/ironic/ironic.conf json_rpc host_ip
 	crudini --set /etc/ironic/ironic.conf json_rpc http_basic_auth_user_file /shared/htpasswd-ironic
 	crudini --set /etc/ironic/ironic.conf json_rpc http_basic_username $IRONIC_HTTP_BASIC_USERNAME
 	crudini --set /etc/ironic/ironic.conf json_rpc http_basic_password $IRONIC_HTTP_BASIC_PASSWORD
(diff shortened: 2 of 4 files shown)


dtantsur:
An alternative would be to always enable HTTP basic authentication for JSON RPC.

zaneb:
> An alternative would be to always enable HTTP basic authentication for JSON RPC.

Something external always has to provide the password, so we can't enable it within the image. What we can do is not expose things dangerously if the container is misconfigured. The options are to bind only to localhost if the password is not provided, or to crash if the password is not provided. I picked the former because it doesn't require synchronised changes to all deployment methods to avoid breakage of working configurations.

dtantsur:
> Something external always has to provide the password, so we can't enable it within the image

We can generate it here, I guess? External entities do not need it.

Of course, some care will be needed to make sure we generate only one password :)

zaneb:
ironic-api and ironic-conductor run in two separate containers (both built from this same image). So we can generate a password here, but it will have different values at each end of the connection ;)

dtantsur:
> So we can generate a password here, but it will have different values at each end of the connection ;)

In theory, we could use the `/shared` partition for that. I'm not sure if we have any atomic operations on mounted volumes, but if yes, this could work:
```bash
temp=$(mktemp /shared/password.XXX)
openssl rand -base64 32 > $temp
ln $temp /shared/json-rpc-password || echo "Already exists"
IRONIC_JSON_RPC_PASSWORD=$(cat /shared/json-rpc/password)
```

runironic.sh
```diff
@@ -2,6 +2,8 @@
 
 . /bin/configure-ironic.sh
 
+crudini --set /etc/ironic/ironic.conf DEFAULT host localhost
```
dhellmann, line 5: Why doesn't this do the same thing that the change in runironic-conductor.sh does?

zaneb: When ironic-conductor is run in a separate container we have to check to see whether it is configured to run in the same pod as ironic-api to see whether we should use localhost as the name of the conductor instance in the DB (if it's in a different pod, conductor gets the hostname from the OS).

When ironic-conductor and ironic-api are in the same container they can always communicate using the localhost interface, so there seemed no point in copying and pasting all of that logic.

An alternative would be to move the setup for this option to configure-ironic.sh. I didn't do that because ironic-api also reads the `host` option for the profiler setup. I think it probably doesn't matter, but I didn't want to get into it.

dhellmann: OK, I missed that this script was running both services in the same container.

zaneb:
> In theory, we could use the `/shared` partition for that.

True! I wasn't aware of that.
But that also relies on external configuration. Honestly, we'd probably want a separate volume so it could be backed by tmpfs, otherwise we'd end up writing the passwords to disk (I realise we're doing that already, but I plan to fix it next).
Plus, like binding to localhost, it only works for the case where the api and conductor are running in the same pod, so it's not like doing this could allow us to forget about any external configuration of the password.

What benefit is there in exposing the RPC API outside of the pod when the api and conductor are both running in the same pod?

PR #182, opened 2020-07-29
Simplify HTTP basic auth configuration

zaneb:
Make HTTP basic auth configuration simpler, more flexible, and a better fit and more secure when used with standard k8s deployment techniques.

* Add the missing basic auth configuration options for ironic-inspector
* Allow basic_auth to be configured independently on different interfaces, based on the presence of the required configuration data, rather than using a single global USE_HTTP_BASIC environment variable.
* Expect all server credentials to be passed in the form of an HTTP_BASIC_HTPASSWD environment variable containing both the username and the *hash* of the password, in the htpasswd format. This is more secure, as it allows containers not to hold a copy of the password when they don't need it purely for authenticating connections.
* Keep server auth user files locally rather than on the /shared volume, so that different servers can have different credentials.
* Expect client credentials to be passed in the form of a file named `/basic_auth/ironic-inspector/auth-config` (for ironic-inspector) or `/basic_auth/ironic-rpc/auth-config` (for the json-rpc interface to ironic-conductor), formatted as an ini config file setting the appropriate options (for basic auth, this is auth_strategy=http_basic, and the username and password options; however this mechanism should work unchanged for other auth strategies). This is more secure because in k8s the password is never passed as an environment variable nor written to disk, but remains in a tmpfs filesystem.

### DIFF
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -71,7 +71,9 @@ if [ "$USE_HTTP_BASIC" = "true" ]; then
 
 	crudini --set /etc/ironic/ironic.conf DEFAULT auth_strategy http_basic
 	crudini --set /etc/ironic/ironic.conf DEFAULT http_basic_auth_user_file /shared/htpasswd-ironic
+
 	crudini --set /etc/ironic/ironic.conf json_rpc auth_strategy http_basic
+	crudini --del /etc/ironic/ironic.conf json_rpc host_ip
 	crudini --set /etc/ironic/ironic.conf json_rpc http_basic_auth_user_file /shared/htpasswd-ironic
 	crudini --set /etc/ironic/ironic.conf json_rpc http_basic_username $IRONIC_HTTP_BASIC_USERNAME
 	crudini --set /etc/ironic/ironic.conf json_rpc http_basic_password $IRONIC_HTTP_BASIC_PASSWORD
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -2,12 +2,6 @@
 
 . /bin/ironic-common.sh
 
-USE_HTTP_BASIC=${USE_HTTP_BASIC:-false}
-IRONIC_HTTP_BASIC_USERNAME=${IRONIC_HTTP_BASIC_USERNAME:-"change_me"}
-IRONIC_HTTP_BASIC_PASSWORD=${IRONIC_HTTP_BASIC_PASSWORD:-"change_me"}
-INSPECTOR_HTTP_BASIC_USERNAME=${IRONIC_HTTP_BASIC_USERNAME:-"change_me"}
-INSPECTOR_HTTP_BASIC_PASSWORD=${IRONIC_HTTP_BASIC_PASSWORD:-"change_me"}
-
 HTTP_PORT=${HTTP_PORT:-"80"}
 MARIADB_PASSWORD=${MARIADB_PASSWORD:-"change_me"}
 NUMPROC=$(cat /proc/cpuinfo  | grep "^processor" | wc -l)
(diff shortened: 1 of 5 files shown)


zaneb:
Equivalent change in ironic-inspector: metal3-io/ironic-inspector-image#61
/cc @iurygregory

configure-ironic.sh
```diff
@@ -67,15 +63,31 @@ env | grep "^OS_" | tee -a /etc/ironic/ironic.extra
 mkdir -p /shared/html
 mkdir -p /shared/ironic_prometheus_exporter
 
-if [ "$USE_HTTP_BASIC" = "true" ]; then
+HTPASSWD_FILE=/etc/ironic/htpasswd
+if [ -n "${HTTP_BASIC_HTPASSWD}" ]; then
+    printf "%s\n" "${HTTP_BASIC_HTPASSWD}" >"${HTPASSWD_FILE}"
```
dhellmann, line 68: Do we need an environment variable at all, or could we just say that the thing running the container needs to provide that file? It's possible to mount a secret as a volume, with the contents of the secret exposed as files on the volume.

zaneb: It is possible to mount the secret as a volume, and in fact that's what I'm doing for the client auth below. I think I originally chose an environment variable because I didn't know that you could filter the keys when you mount as a volume (to avoid giving the plaintext password to a server that doesn't need it). You can, although it would be a little weird because you'd have to define two volumes on the Pod for each secret (with different lists of keys), and then mount different ones in different places in different Containers.

The HTTP_BASIC_HTPASSWD only contains a hash, so there's no security driver for mounting it as a volume - it's safe to pass as an environment variable, write to disk &c. So I think it could go either way. Is there a reason to prefer volume vs environment variable?

dhellmann: I just thought that instead of the container writing the contents of a variable to disk, the file could be provided directly. It's not a big deal, and maybe the consistency of always using environment variables is better.

configure-ironic.sh
```diff
@@ -67,15 +63,31 @@ env | grep "^OS_" | tee -a /etc/ironic/ironic.extra
...
-	htpasswd -nbB $IRONIC_HTTP_BASIC_USERNAME $IRONIC_HTTP_BASIC_PASSWORD > /shared/htpasswd-ironic
-fi
+configure_client_basic_auth() {
+    local auth_config_file="/auth/$1/auth-config"
+    if [ -f ${auth_config_file} ]; then
+        IRONIC_API_CONFIG_OPTIONS+=" --config_file ${auth_config_file}"
+    fi
+}
+
+configure_client_basic_auth ironic-inspector
```
iurygregory, line 92: Do we need to configure inronic-inspector and pass the config for the ironic-api?

zaneb: Basically this is configuring ironic-api to be a client of ironic-inspector, by adding the config file `/auth/ironic-inspector/auth-config` if it exists. This file will look something like:
```
[inspector]
auth_strategy=http_basic
username=...
password=...
```

There's a corresponding change in the ironic-inspector image to allow it to authenticate as a client to ironic-api.

runironic.sh
```diff
@@ -2,6 +2,8 @@
 
 . /bin/configure-ironic.sh
 
+crudini --set /etc/ironic/ironic.conf DEFAULT host localhost
```
juliakreger, line 5: Just to note, this will functionally restrict the services to only run as a single pair on a single host because [DEFAULT]host is what is used to register the location of where to send the RPC messages. While not an issue now, it will likely become an issue at some point.

dhellmann: I think that restriction is acceptable for the time being.

PR #183, opened 2020-07-29
Exit if ironic-api or ironic-conductor exits

zaneb:
Using the default entrypoint, which runs both the ironic-api and
ironic-conductor processes in the same container, the container should
exit immediately if either of those processes ends.


dtantsur (approved):
/lgtm

PR #184, opened 2020-07-30
Remove all unused modules

derekhiggins:
Leave httpd with the minimum needed for metal3,
in particular we wanted to disable directory listing and
http TRACE. But disabling everything we don't use seems
prudent.


derekhiggins:
Submitted downstream here https://github.com/openshift/ironic-image/pull/99

PR #186, opened 2020-08-04
Configure ironic-conductor to auth to ironic-inspector

zaneb:
The ironic-conductor process needs to authenticate to the
ironic-inspector API, so we need to pass the auth config file options to
it also.


zaneb:
Tested locally and it's working end-to-end now.
/assign @dhellmann
/cc @iurygregory

PR #187, opened 2020-08-07
feat: enable ibmc raid interface

IamFive:
enable ibmc raid interface for ironic.


IamFive:
@dtantsur  hello, it seems the test integration did not run, could u help retrigger it. thanks.

Issue #188, opened 2020-08-10
CI is failing as Ironic fails to start


kashifest:
The ironic fails to start and after BMH is applied, it doesn't even go for `registration`. 

```
kubectl get bmh -A
NAMESPACE   NAME     STATUS   PROVISIONING STATUS   CONSUMER   BMC                                                                                         HARDWARE PROFILE   ONLINE   ERROR
metal3      node-0                                             ipmi://192.168.111.1:6230                                                                                      false    
metal3      node-1                                             redfish+http://192.168.111.1:8000/redfish/v1/Systems/a5717dc1-48b0-46b0-b6b1-c90c7d4a742b                      false    
metal3      node-2                                             ipmi://192.168.111.1:6232                                                                                      false    
metal3      node-3                                             redfish+http://192.168.111.1:8000/redfish/v1/Systems/a4906598-e454-4811-8987-fa2fd48d63cc                      false    
``` 
The ironic log is complaining about the following: 

`Failed to register hardware types. For hardware type 'fake-hardware', no default value found for bios interface.: ironic.common.exception.NoValidDefaultForInterface: For hardware type 'fake-hardware', no default value found for bios interface. `

This started in the CI after the PR #162 went in. Disabling it is making the CI to pass, 

```
kubectl get bmh -A
NAMESPACE   NAME     STATUS   PROVISIONING STATUS   CONSUMER   BMC                                                                                         HARDWARE PROFILE   ONLINE   ERROR
metal3      node-0   OK       ready                            ipmi://192.168.111.1:6230                                                                   unknown            false    
metal3      node-1   OK       ready                            redfish+http://192.168.111.1:8000/redfish/v1/Systems/a5717dc1-48b0-46b0-b6b1-c90c7d4a742b   unknown            false    
metal3      node-2   OK       ready                            ipmi://192.168.111.1:6232                                                                   unknown            false    
metal3      node-3   OK       ready                            redfish+http://192.168.111.1:8000/redfish/v1/Systems/a4906598-e454-4811-8987-fa2fd48d63cc   unknown            false 
```

Enabling idrac needs further investigation.

maelk:
you also reported the following error in the chat : 
```
Failed to register hardware types. For hardware type 'fake-hardware', no default value found for bios interface.: ironic.common.exception.NoValidDefaultForInterface: For hardware type 'fake-hardware', no default value found for bios interface. 
```

kashifest:
[...]

PR #181, opened 2020-07-28
Enable 'idrac-wsman' interface for RAID Configuration

nasirkamal:
Enable 'idrac-wsman' interface in ironic.conf for inspect, management, power and raid. This is important to configure and manage Dell Power Edge servers.

This is related to PR #146, where I accidentally deleted the parent repository.

PR #162, opened 2020-06-09
enable idrac-wsman BIOS interface

demonCoder95:
I'm working on BIOS configuration with the `idrac-wsman` driver. Since this requires enabling the interface, I've edited the `ironic.conf` to do so.

PR #189, opened 2020-08-10
Disable bios interfaces idrac-wsman

kashifest:
A temporary fix for the issue #188 

We can enable idrac once we have done proper investigation.

### DIFF
--- a/ironic.conf
+++ b/ironic.conf
@@ -5,7 +5,6 @@ default_boot_interface = ipxe
 default_deploy_interface = direct
 default_inspect_interface = inspector
 default_network_interface = noop
-enabled_bios_interfaces = idrac-wsman
 enabled_boot_interfaces = pxe,ipxe,fake,redfish-virtual-media,idrac-redfish-virtual-media
 enabled_deploy_interfaces = direct,fake
 # NOTE(dtantsur): when changing this, make sure to update the driver


fmuyassarov:
I think it failed because CI didn't build ironic-image but instead pulled it from the quay.

kashifest:
> I think it failed because CI didn't build ironic-image but instead pulled it from the quay.

Yes, there is another bug in metal3-dev-env. Thats why the previous PR also went it without giving the error.

fmuyassarov:
I wondering why if we need the CI here since the images are pulled from the public registry instead of building and testing them locally.

furkatgofurov7:
As #162  causing the problem in CI currently, there is another PR #181 that might cause the CI fail again maybe once it is in.

kashifest:
@furkatgofurov7 those are not enabling bios interfaces, so lets see , we need to fix  metal3-dev-env to make sure that the local ironic image is tested properly in CI, I beleive @maelk is looking into it

maelk:
The metal3-dev-env fix is here : https://github.com/metal3-io/metal3-dev-env/pull/420

ironic.conf
```diff
@@ -5,7 +5,6 @@ default_boot_interface = ipxe
 default_deploy_interface = direct
 default_inspect_interface = inspector
 default_network_interface = noop
-enabled_bios_interfaces = idrac-wsman
```
dtantsur, line 8: The correct fix would be to add `no-bios` here.

furkatgofurov7: @dtantsur should we open a PR with the configuration you are suggesting?

PR #190, opened 2020-08-13
Use collector name in ironic.conf expected by IPA

bfournie:
Make the name consistent with inspector.ipxe.


dtantsur:
/test-integration
/lgtm

Good catch!

PR #191, opened 2020-08-20
enable idrac-wsman RAID and BIOS interfaces

demonCoder95:
This patch is as a follow-up to #188 , which was caused by a faulty configuration.
This configuration has been tested in my setup and causes no issues with ironic conductor setup.

@maelk This can be merged now.


ironic.conf
```diff
@@ -5,6 +5,7 @@ default_boot_interface = ipxe
 default_deploy_interface = direct
 default_inspect_interface = inspector
 default_network_interface = noop
+enabled_bios_interfaces = idrac-wsman, no-bios
```
dtantsur, line 8: Please no spaces. They should be handled in theory, but let's be consistent

ironic.conf
```diff
@@ -13,7 +14,7 @@ enabled_hardware_types = ipmi,idrac,irmc,fake-hardware,redfish,ibmc
 enabled_inspect_interfaces = inspector,idrac,irmc,fake,redfish
 enabled_management_interfaces = ipmitool,idrac,irmc,fake,redfish,idrac-redfish,ibmc
 enabled_power_interfaces = ipmitool,idrac,irmc,fake,redfish,idrac-redfish,ibmc
-enabled_raid_interfaces = no-raid,irmc,agent,fake,ibmc
+enabled_raid_interfaces = no-raid,irmc,agent,fake,ibmc, idrac-wsman
```
dtantsur, line 17: same

PR #193, opened 2020-09-02
Add support for noop management and manual-management

dtantsur:
This adds three closely related features:
* The noop management interface allows skip management actions
  (boot device and mode) for a node. The PXE netboot fallback
  is enabled for it to be able to local boot properly.
* The manual-management hardware type also skips power actions.

This change misses the agent power interface that can allow
the manual-management hardware type to use a running agent
for power actions to achieve a limited deploy capability
without BMC credentials [1]. We need to enable it conditional
on fast-track (and it's disabled in the metal3 CI).

[1] https://docs.openstack.org/ironic/latest/admin/agent-power.html

### DIFF
--- a/ironic.conf.j2
+++ b/ironic.conf.j2
@@ -15,9 +15,9 @@ enabled_boot_interfaces = pxe,ipxe,fake,redfish-virtual-media,idrac-redfish-virt
 enabled_deploy_interfaces = direct,fake
 # NOTE(dtantsur): when changing this, make sure to update the driver
 # dependencies in Dockerfile.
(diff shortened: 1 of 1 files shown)


dtantsur:
Aha, we have fast-track disabled. Not wow.

elfosardo (approved):

maelk:
Let's see if https://github.com/metal3-io/baremetal-operator/pull/430 would pass now. We coudl re-enable fasttrack

maelk (commented):
/lgtm

PR #194, opened 2020-09-06
Change ipxe git in dockerfile

shibaPuppy:
In the private environment of some users, it is forbbiden during redirectin.

the commonly used github URL looks better.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ FROM docker.io/centos:centos8 AS builder
 RUN yum install -y gcc git make genisoimage xz-devel grub2 grub2-efi-x64-modules shim dosfstools mtools
 WORKDIR /tmp
 COPY . .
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM docker.io/centos:centos8 AS builder
 RUN yum install -y gcc git make genisoimage xz-devel grub2 grub2-efi-x64-modules shim dosfstools mtools
 WORKDIR /tmp
 COPY . .
-RUN git clone http://git.ipxe.org/ipxe.git && \
+RUN git clone https://github.com/ipxe/ipxe.git && \
```
dtantsur, line 8: https as a bonus!

dtantsur (approved):
/test-integration
/lgtm

Dockerfile
shibaPuppy: @dtantsur 
why doesn't it merge,?

PR #196, opened 2020-09-09
Explicitly set json_rpc.auth_strategy to noauth

derekhiggins:
Using crudini to merge this into ironic.conf no longer
works as it gets ignored as it matches the DEFAULT and
it is redundant. But later we then change the DEFAULT in
set_http_basic_server_auth_strategy so we need to set
json_rpc.auth_strategy back to noauth.

Fixes: https://github.com/openshift-metal3/dev-scripts/issues/1099

### DIFF
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -76,6 +76,9 @@ set_http_basic_server_auth_strategy() {
 # Configure HTTP basic auth for ironic-api server
 if [ -f "${HTPASSWD_FILE}" ]; then
     set_http_basic_server_auth_strategy
+    # We've just set the DEFAULT auth_strategy to http_basic, we need to
+    # set the json_rpc auth_strategy back to noauth
+    crudini --set /etc/ironic/ironic.conf json_rpc auth_strategy noauth
 fi
 
 
--- a/ironic.conf
+++ b/ironic.conf
@@ -42,7 +42,9 @@ send_sensor_data = true
 send_sensor_data_interval = 160
 
 [json_rpc]
-auth_strategy = noauth
+# crudini doesn't merge this into ironic.conf as it sees it as redundant (its set globably in DEFAULT)
+# so we set it in configure-ironic.sh if we change the DEFAULT
+# auth_strategy = noauth
 host_ip = localhost
 
 [deploy]


ironic.conf
```diff
@@ -42,7 +42,9 @@ send_sensor_data = true
 send_sensor_data_interval = 160
 
 [json_rpc]
-auth_strategy = noauth
+# curdini doesn't merge this into ironic.conf as it see it as redundant (its set globably in DEFAULT)
```
dtantsur, line 45: nit: crudini

configure-ironic.sh
```diff
@@ -76,6 +76,9 @@ set_http_basic_server_auth_strategy() {
 # Configure HTTP basic auth for ironic-api server
 if [ -f "${HTPASSWD_FILE}" ]; then
     set_http_basic_server_auth_strategy
+    # We've just set the DEFAULT auth_strategy to http_basic, we need to
+    # set the json_rpc auth_strategy back to noauth
+    crudini --set /etc/ironic/ironic.conf json_rpc auth_strategy noauth
```
dtantsur, line 81: Sigh. Why don't we use authentication for JSON RPC?

dtantsur (approved):
/test-integration

configure-ironic.sh
iurygregory: Why we need to set  to `noauth` if we are going to set to `http_basic`? .-. (At least this should only set things if `http_basic` is desired.

hardys: Since https://github.com/metal3-io/ironic-image/pull/180 the json_rpc only happens via localhost, so I think we don't need auth, unlike the API where we do want the option of choosing http_basic?

derekhiggins: > Why we need to set to `noauth` if we are going to set to `http_basic`? .-. (At least this should only set things if `http_basic` is desired.

When we get to this point we have
DEFAULT.auth_strategy = http_basic

If we don't set 
json_rpc.auth_strategy = noauth 
it picks up the DEFAULT value




> Since #180 the json_rpc only happens via localhost, so I think we don't need auth, unlike the API where we do want the option of choosing http_basic?

yup, we're only listening on ::1
tcp6       0      0 ::1:8089                :::*                    LISTEN      97565/python3

maelk: I think we should have authentication even if rpc is done only via localhost. Considering the pivoting use case, Ironic runs on a node where workload is also running and could also be in host networking, hence being able to access the json RPC interface. I would rather keep json_rpc authenticated.

maelk:
I would propose to properly setup json-rpc authentication instead. Here is a PR doing the change : https://github.com/metal3-io/ironic-image/pull/200

configure-ironic.sh
derekhiggins: Yup, I'm happy to turn on auth for rpc, I'm not suggesting we don't do it, all I am trying to do here is fix the config so it works when auth is turned on for the ironic API.

ironic.conf
```diff
@@ -42,7 +42,9 @@ send_sensor_data = true
 send_sensor_data_interval = 160
 
 [json_rpc]
-auth_strategy = noauth
+# crudini doesn't merge this into ironic.conf as it sees it as redundant (its set globably in DEFAULT)
```
zaneb, line 45: This seems like a clear bug in crudini. How can it possibly know that that the option in the DEFAULT group is the fallback for one in another section without an almost-human-level of understanding [the code](https://opendev.org/openstack/ironic/src/branch/master/ironic/common/json_rpc/__init__.py#L20)? This group could be contributed by a completely independent library for all it knows.

configure-ironic.sh
zaneb: Authentication for json-rpc is [fully](https://github.com/metal3-io/ironic-image/blob/master/runironic-conductor.sh#L8-L10) [supported](https://github.com/metal3-io/ironic-image/blob/master/configure-ironic.sh#L93) by this image. It is a lot easier to set up if you have a version of ironic that includes [the fix to change the client configuration to the same format as the API clients](https://review.opendev.org/742474), but that is the only speedbump.

zaneb (commented):
This looks correct to me (given the insane presumptuousness of crudini as described, anyway). It's not clear what patch in upstream Ironic caused a change though, since this image was definitely working with basic-auth turned on for the API only.

maelk:
As part of the TLS integration work, I have put together a PR that switches to use jinja2 templating insteady of crudini. I believe it would solve this issue at the same time. Please take a look : #201

derekhiggins:
Fixed by https://github.com/metal3-io/ironic-image/pull/201

PR #198, opened 2020-09-10
Add TLS support in Ironic configuration

maelk:
This commit allows the user to start Ironic using TLS on all
endpoints, by setting the following environment variables:

- CACERT_FILE: path to the CA cert
- CERT_FILE: path to the cert
- KEY_FILE: path to the key

### DIFF
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -1,5 +1,25 @@
 #!/usr/bin/bash
 
+export IRONIC_CERT_FILE=/certs/ironic/tls.crt
+export IRONIC_KEY_FILE=/certs/ironic/tls.key
+export IRONIC_CACERT_FILE=/certs/ca/ironic/tls.crt
+export IRONIC_INSPECTOR_CERT_FILE=/certs/ironic-inspector/tls.crt
+export IRONIC_INSPECTOR_CACERT_FILE=/certs/ca/ironic-inspector/tls.crt
+
+mkdir -p /certs/ironic
+mkdir -p /certs/ironic-inspector
+mkdir -p /certs/ca/ironic
+mkdir -p /certs/ca/ironic-inspector
+
+if [ -f "$IRONIC_CERT_FILE" ] && [ ! -f "$IRONIC_KEY_FILE" ] ; then
+    echo "Missing TLS Certificate key file /certs/ironic/key"
+    exit 1
+fi
+if [ ! -f "$IRONIC_CERT_FILE" ] && [ -f "$IRONIC_KEY_FILE" ] ; then
+    echo "Missing TLS Certificate file /certs/ironic/crt"
+    exit 1
+fi
+
 . /bin/ironic-common.sh
 
 export HTTP_PORT=${HTTP_PORT:-"80"}
@@ -16,6 +36,28 @@ export IRONIC_FAST_TRACK=${IRONIC_FAST_TRACK:-true}
 export IRONIC_AUTOMATED_CLEAN=${IRONIC_AUTOMATED_CLEAN:-true}
 
 wait_for_interface_or_ip
+
+if [ -f "$IRONIC_CERT_FILE" ]; then
+    export IRONIC_TLS_SETUP="true"
+    export IRONIC_BASE_URL="https://${IRONIC_URL_HOST}:6385"
+    if [ ! -f "$IRONIC_CACERT_FILE" ]; then
+        cp "$IRONIC_CERT_FILE" "$IRONIC_CACERT_FILE"
+    fi
+else
+    export IRONIC_TLS_SETUP="false"
+    export IRONIC_BASE_URL="http://${IRONIC_URL_HOST}:6385"
+fi
+
+if [ -f "$IRONIC_INSPECTOR_CERT_FILE" ] || [ -f "$IRONIC_INSPECTOR_CACERT_FILE" ]; then
+    export IRONIC_INSPECTOR_TLS_SETUP="true"
+    export IRONIC_INSPECTOR_BASE_URL="https://${IRONIC_URL_HOST}:5050"
+    if [ ! -f "$IRONIC_INSPECTOR_CACERT_FILE" ]; then
+        cp "$IRONIC_INSPECTOR_CERT_FILE" "$IRONIC_INSPECTOR_CACERT_FILE"
+    fi
+else
+    export IRONIC_INSPECTOR_TLS_SETUP="false"
+    export IRONIC_INSPECTOR_BASE_URL="http://${IRONIC_URL_HOST}:5050"
+fi
 
 cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig
 
--- a/ironic.conf.j2
+++ b/ironic.conf.j2
@@ -50,6 +50,9 @@ max_command_attempts = 30
 [api]
 host_ip = {% if env.LISTEN_ALL_INTERFACES | lower == "true" %}::{% else %}{{ env.IRONIC_IP }}{% endif %}
 api_workers = {{ env.NUMWORKERS }}
+{% if env.IRONIC_TLS_SETUP == "true" %}
+enable_ssl_api = true
+{% endif %}
 
 [conductor]
 automated_clean = {{ env.IRONIC_AUTOMATED_CLEAN }}
@@ -77,12 +80,15 @@ fast_track = {{ env.IRONIC_FAST_TRACK }}
 dhcp_provider = none
 
 [inspector]
-endpoint_override = http://{{ env.IRONIC_URL_HOST }}:5050
+endpoint_override = {{ env.IRONIC_INSPECTOR_BASE_URL }}
 power_off = {{ false if env.IRONIC_FAST_TRACK == "true" else true }}
+{% if env.IRONIC_INSPECTOR_TLS_SETUP == "true" %}
+cafile = {{ env.IRONIC_INSPECTOR_CACERT_FILE }}
+{% endif %}
 # TODO(dtantsur): ipa-api-url should be populated by ironic itself, but it's
 # not, so working around here.
 # NOTE(dtantsur): keep inspection arguments synchronized with inspector.ipxe
-extra_kernel_params = console=ttyS0 ipa-insecure=1 ipa-inspector-collectors=default,extra-hardware,logs ipa-inspection-dhcp-all-interfaces=1 ipa-collect-lldp=1 {% if env.IRONIC_FAST_TRACK == "true" %} ipa-api-url=http://{{ env.IRONIC_URL_HOST }}:6385 {% endif %}
+extra_kernel_params = console=ttyS0 ipa-insecure=1 ipa-inspector-collectors=default,extra-hardware,logs ipa-inspection-dhcp-all-interfaces=1 ipa-collect-lldp=1 {% if env.IRONIC_FAST_TRACK == "true" %} ipa-api-url={{ env.IRONIC_BASE_URL }} {% endif %}
 
 [ipmi]
 # use_ipmitool_retries transfers the responsibility of retrying to ipmitool
@@ -101,7 +107,6 @@ use_ipmitool_retries = true
 min_command_interval = 5
 command_retry_timeout = 60
 
-
 [json_rpc]
 # We assume that when we run API and conductor in the same container, they use
 # authentication over localhost, using the same credentials as API, to prevent
@@ -116,6 +121,10 @@ http_basic_auth_user_file = /etc/ironic/htpasswd-rpc
 host_ip = localhost
 {% else %}
 host_ip = {% if env.LISTEN_ALL_INTERFACES | lower == "true" %}::{% else %}{{ env.IRONIC_IP }}{% endif %}
+{% endif %}
+{% if env.IRONIC_TLS_SETUP == "true" %}
+use_ssl = true
+cafile = {{ env.IRONIC_CACERT_FILE }}
 {% endif %}
 
 [oslo_messaging_notifications]
@@ -142,4 +151,10 @@ use_swift = false
 kernel_append_params = console=ttyS0 nofb nomodeset vga=normal ipa-insecure=1
 
 [service_catalog]
-endpoint_override = http://{{ env.IRONIC_URL_HOST }}:6385
+endpoint_override = {{ env.IRONIC_BASE_URL }}
+
+{% if env.IRONIC_TLS_SETUP == "true" %}
+[ssl]
+cert_file = {{ env.IRONIC_CERT_FILE }}
+key_file = {{ env.IRONIC_KEY_FILE }}
+{% endif %}
--- a/runhttpd.sh
+++ b/runhttpd.sh
@@ -2,6 +2,7 @@
 
 . /bin/ironic-common.sh
 
+IRONIC_CERT_FILE=/certs/ironic/tls.crt
 HTTP_PORT=${HTTP_PORT:-"80"}
 
 # Whether to enable fast_track provisioning or not
@@ -12,10 +13,16 @@ wait_for_interface_or_ip
 mkdir -p /shared/html
 chmod 0777 /shared/html
 
-if [[ $IRONIC_FAST_TRACK == true ]]; then
-    INSPECTOR_EXTRA_ARGS="ipa-api-url=http://${IRONIC_URL_HOST}:6385"
+if [ -f "$IRONIC_CERT_FILE" ]; then
+    IRONIC_BASE_URL="https://${IRONIC_URL_HOST}"
 else
-    INSPECTOR_EXTRA_ARGS=""
+    IRONIC_BASE_URL="http://${IRONIC_URL_HOST}"
+fi
+
+if [[ $IRONIC_FAST_TRACK == true ]]; then
+    INSPECTOR_EXTRA_ARGS=" ipa-api-url=${IRONIC_BASE_URL}:6385 ipa-inspection-callback-url=${IRONIC_BASE_URL}:5050/v1/continue"
+else
+    INSPECTOR_EXTRA_ARGS=" ipa-inspection-callback-url=${IRONIC_BASE_URL}:5050/v1/continue"
 fi
 
 # Copy files to shared mount
--- a/inspector.ipxe
+++ b/inspector.ipxe
@@ -5,6 +5,6 @@ echo In inspector.ipxe
 imgfree
 # NOTE(dtantsur): keep inspection kernel params in [mdns]params in
 # ironic-inspector-image and configuration in configure-ironic.sh
-kernel --timeout 60000 http://IRONIC_IP:HTTP_PORT/images/ironic-python-agent.kernel console=ttyS0 ipa-inspection-callback-url=http://IRONIC_IP:5050/v1/continue ipa-inspection-collectors=default,extra-hardware,logs systemd.journald.forward_to_console=yes BOOTIF=${mac} ipa-debug=1 ipa-inspection-dhcp-all-interfaces=1 ipa-collect-lldp=1 EXTRA_ARGS initrd=ironic-python-agent.initramfs || goto retry_boot
+kernel --timeout 60000 http://IRONIC_IP:HTTP_PORT/images/ironic-python-agent.kernel console=ttyS0 ipa-insecure=1 ipa-inspection-collectors=default,extra-hardware,logs systemd.journald.forward_to_console=yes BOOTIF=${mac} ipa-debug=1 ipa-inspection-dhcp-all-interfaces=1 ipa-collect-lldp=1 EXTRA_ARGS initrd=ironic-python-agent.initramfs || goto retry_boot
 initrd --timeout 60000 http://IRONIC_IP:HTTP_PORT/images/ironic-python-agent.initramfs || goto retry_boot
 boot


configure-ironic.sh
```diff
@@ -46,6 +51,36 @@ connection = mysql+pymysql://ironic:${MARIADB_PASSWORD}@localhost/ironic?charset
...
+$([ ! -z "$INSECURE" ] && echo "insecure = $INSECURE" || echo "insecure = false")
+# TODO(dtantsur): ipa-api-url should be populated by ironic itself, but it's
+# not, so working around here.
+# NOTE(dtantsur): keep inspection arguments synchronized with inspector.ipxe
+extra_kernel_params = console=ttyS0 ipa-insecure=1 ipa-inspector-collectors=default,extra-hardware,logs ipa-inspection-dhcp-all-interfaces=1 ipa-collect-lldp=1 ${INSPECTOR_EXTRA_ARGS}
+[pxe]
+pxe_append_params = nofb nomodeset vga=normal ipa-insecure=1
+[service_catalog]
+endpoint_override = https://${IRONIC_URL_HOST}:6385
+$([ ! -z "$CACERT_FILE" ] && echo "cafile = $CACERT_FILE")
```
dtantsur, line 71: You don't need anything here, we never really access ironic this way.

configure-ironic.sh
```diff
@@ -46,6 +51,36 @@ connection = mysql+pymysql://ironic:${MARIADB_PASSWORD}@localhost/ironic?charset
...
+$([ ! -z "$INSECURE" ] && echo "insecure = $INSECURE" || echo "insecure = false")
+[ssl]
+$([ ! -z "$CACERT_FILE" ] && echo "cafile = $CACERT_FILE")
```
dtantsur, line 74: Are you planning on client certificate validation? Won't work with the ramdisk, let's skip for now.

configure-ironic.sh
```diff
@@ -46,6 +51,36 @@ connection = mysql+pymysql://ironic:${MARIADB_PASSWORD}@localhost/ironic?charset
...
+$([ ! -z "$CERT_FILE" ] && echo "cert_file = $CERT_FILE")
+$([ ! -z "$KEY_FILE" ] && echo "key_file = $KEY_FILE")
+[json_rpc]
+use_ssl = true
+$([ ! -z "$CACERT_FILE" ] && echo "cafile = $CACERT_FILE")
+$([ ! -z "$INSECURE" ] && echo "insecure = $INSECURE" || echo "insecure = false")
```
dtantsur, line 80: We definitely should not use insecure with JSON RPC, it's completely internal to ironic.

configure-ironic.sh
```diff
@@ -46,6 +51,36 @@ connection = mysql+pymysql://ironic:${MARIADB_PASSWORD}@localhost/ironic?charset
...
 
+EOF
+
+if [ ! -z "$CACERT_FILE" ] || [ ! -z "$INSECURE" ]; then
+    crudini --merge /etc/ironic/ironic.conf <<EOF
+[inspector]
+endpoint_override = https://${IRONIC_URL_HOST}:5050
+power_off = ${INSPECTOR_POWER_OFF}
+$([ ! -z "$CACERT_FILE" ] && echo "cafile = $CACERT_FILE")
+$([ ! -z "$INSECURE" ] && echo "insecure = $INSECURE" || echo "insecure = false")
```
dtantsur, line 62: I'm worried about supporting insecure. What's the use case for that?

configure-ironic.sh
```diff
@@ -46,6 +51,36 @@ connection = mysql+pymysql://ironic:${MARIADB_PASSWORD}@localhost/ironic?charset
...
 fast_track = ${IRONIC_FAST_TRACK}
 
+EOF
+
+if [ ! -z "$CACERT_FILE" ] || [ ! -z "$INSECURE" ]; then
+    crudini --merge /etc/ironic/ironic.conf <<EOF
+[inspector]
+endpoint_override = https://${IRONIC_URL_HOST}:5050
+power_off = ${INSPECTOR_POWER_OFF}
+$([ ! -z "$CACERT_FILE" ] && echo "cafile = $CACERT_FILE")
```
dtantsur, line 61: I think it's the right time to migrate to jinja for ironic.conf

runhttpd.sh
```diff
@@ -13,11 +13,25 @@ mkdir -p /shared/html
...
     INSPECTOR_EXTRA_ARGS=""
 fi
 
+if [ ! -z "$CACERT_FILE" ] || [ ! -z "$INSECURE" ]; then
+    INSPECTOR_EXTRA_ARGS="${INSPECTOR_EXTRA_ARGS} ipa-inspection-callback-url=https://${IRONIC_URL_HOST}:5050/v1/continue"
+else
+    INSPECTOR_EXTRA_ARGS="${INSPECTOR_EXTRA_ARGS} ipa-inspection-callback-url=http://${IRONIC_URL_HOST}:5050/v1/continue"
+fi
+
+if [ ! -z "$IPA_INSECURE" ]; then
```
dtantsur, line 31: This should be unconditional until we have at least one way to pass certificates to the ramdisk (tracked in https://storyboard.openstack.org/#!/story/2007214)

dtantsur (changes requested):
Heh, I was about to start a similar work next week :) Some comments inline.

configure-ironic.sh
maelk: ok, I'll remove the CA

configure-ironic.sh
maelk: so no configuration is needed in the service_catalog section ?

configure-ironic.sh
maelk: ok, I'll remove

configure-ironic.sh
maelk: I just added support for it, as it is a configuration option, but I can remove it from everywhere

configure-ironic.sh
maelk: agreed, let me rework this PR

runhttpd.sh
maelk: ok

configure-ironic.sh
```diff
@@ -19,7 +19,11 @@ if [[ $IRONIC_FAST_TRACK == true ]]; then
     INSPECTOR_POWER_OFF=false
     # TODO(dtantsur): ipa-api-url should be populated by ironic itself, but
     # it's not yet, so working around here.
-    INSPECTOR_EXTRA_ARGS=" ipa-api-url=http://${IRONIC_URL_HOST}:6385"
+    if [ ! -z "$CACERT_FILE" ] || [ ! -z "$INSECURE" ]; then
```
kashifest, line 22: Is the 'OR' condition because we foresee a configuration where we would have the ca-cert file and yet we would not enable TLS? or should we use 'AND', or rather just check the existense of ca-cert to enable TLS?

kashifest: even when INSECURE is false the value is set right? is this condition correct?

maelk: I was thinking we could have a configuration without CA cert, but the user wants to use insecure to not verify the cert with the CA, hence the "or". However, as pointed out by @dtantsur we should probably not set the CA cert at all, and not allow for insecure.

dtantsur (commented):
Seems mostly okay, but I'm still not sure why you want CAFILE and CERT_FILE to be different. Is it intended? In the simplest configuration it's the same file.

ironic.conf.j2
```diff
@@ -71,23 +117,19 @@ pxe_config_template = $pybasedir/drivers/modules/ipxe_config.template
...
-min_command_interval = 5
-command_retry_timeout = 60
+[service_catalog]
+endpoint_override = {{ env["IRONIC_BASE_URL"] }}:6385
+{% if "CACERT_FILE" in env %}
+cafile = {{ env["CACERT_FILE"] }}
+{% endif %}
+
+{% if "CERT_FILE" in env %}
+[ssl]
```
dtantsur, line 132: Just a note. The built-in TLS support works, but may not be the most robust in the world. Ideally, we should use something like nginx for that. Likely not in the first iteration though.

.gitignore
```diff
@@ -0,0 +1,2 @@
+*.crt
```
dtantsur, line 1: This seems based on your development environment, not sure if we should commit it..

configure-ironic.sh
```diff
@@ -2,59 +2,40 @@
...
-    INSPECTOR_EXTRA_ARGS=" ipa-api-url=http://${IRONIC_URL_HOST}:6385"
+if [ -n "$CERT_FILE" ] && [ -z "$KEY_FILE" ] ; then
+    echo "Missing TLS Certificate key file"
+    exit 1
+fi
+if [ -z "$CERT_FILE" ] && [ -n "$KEY_FILE" ] ; then
+    echo "Missing TLS Certificate file"
+    exit 1
+fi
+
```
dtantsur, line 26: If `CERT_FILE` is set, but `CACERT_FILE` is not, lets default `CACERT_FILE=$CERT_FILE`. Otherwise the case of not providing a CA won't work.

ironic.conf.j2
```diff
@@ -21,6 +21,8 @@ use_stderr = true
 require_agent_token = true
 # NOTE(dtantsur): the default md5 is not compatible with FIPS mode
 hash_ring_algorithm = sha256
+my_ip = {{ env["IRONIC_IP"] }}
```
dtantsur, line 24: very nit: you can use dot access in jinja (`env.IRONIC_IP`).

ironic.conf.j2
```diff
@@ -71,23 +117,19 @@ pxe_config_template = $pybasedir/drivers/modules/ipxe_config.template
...
-# where :
-#    X = command_retry_timeout / min_command_interval
-#    Y = min_command_interval
-# If use_ipmitool_retries is false, then ironic retries X times, with an
-# interval of Y in between each tries.
-min_command_interval = 5
-command_retry_timeout = 60
+[service_catalog]
+endpoint_override = {{ env["IRONIC_BASE_URL"] }}:6385
+{% if "CACERT_FILE" in env %}
```
dtantsur, line 127: nit: not needed. no HTTP calls are made with this configuration in reality, it's only used for `endpoint_override`.

dtantsur (changes requested):
I've read an explanation on the BMO patch, and it makes sense to me. The patch looks good, there is only one comment on the configuration part - see inline.

.gitignore
maelk: I put it there to make sure no certificate or keys get commited. it's not for my dev env specifically, just a precaution. I can remove it if you think it is unneeded

configure-ironic.sh
maelk: fixed, thank you. I added an export to set that variable in configure-ironic.sh when it is unset

ironic.conf.j2
maelk: I changed it, thanks

ironic.conf.j2
maelk: ok, removed

maelk:
/test-integration

@dtantsur I'll make a follow-up PR to run ironic with WSGI behind Apache.

.gitignore
dtantsur: I'd rather not have it here.

ironic.conf.j2
```diff
@@ -71,23 +117,16 @@ pxe_config_template = $pybasedir/drivers/modules/ipxe_config.template
 tftp_master_path = /shared/tftpboot
 tftp_root = /shared/tftpboot
 uefi_pxe_config_template = $pybasedir/drivers/modules/ipxe_config.template
+pxe_append_params = nofb nomodeset vga=normal ipa-insecure=1
```
dtantsur, line 120: Oh, nearly forgot! Redfish virtual media does not use (because it does not use PXE), you need to update this option: https://opendev.org/openstack/ironic/src/branch/master/ironic/conf/redfish.py#L66 (I intend to make host validation work with virtual media, but it's not quite there).

dtantsur (changes requested):

ironic.conf.j2
maelk: do we need to set ipa-inspection-callback-url and possibly ipa-api-url here ?

dtantsur: AFAIR no

ironic.conf.j2
```diff
@@ -31,29 +33,73 @@ deploy_logs_local_path = /shared/log/ironic/deploy
...
 [dhcp]
 dhcp_provider = none
 
+[inspector]
+endpoint_override = {{ env.IRONIC_BASE_URL }}:5050
+power_off = {{ false if env.IRONIC_FAST_TRACK == "true" else true }}
+{% if "CACERT_FILE" in env %}
+cafile = {{ env.CACERT_FILE }}
+{% endif %}
+# TODO(dtantsur): ipa-api-url should be populated by ironic itself, but it's
```
dtantsur, line 71: Btw, this issue has been fixed upstream

dtantsur (approved):
The approach looks good to me

dtantsur:
Ouch, needs resolving a merge conflict now.

kashifest:
lgtm

dtantsur (approved):

maelk:
The CI had a hole, the ironic and inspector images were not properly tested

maelk:
/cc @zaneb 
/cc @dhellmann 
/cc @derekhiggins 
How does this look for you ?

zaneb (commented):
There are definitely at least some bugs here. For instance, the RPC socket binds to `${IRONIC_IP}` instead of `::` now, meaning that you can't connect to it using the loopback interface. But `runironic-conductor.sh` sets the `[DEFAULT]/host` config option to localhost when auth is not configured, and that is how the conductor is identified in the database so ironic-api will try to connect to it using the loopback interface. Also the RPC socket is now publicly accessible outside the pod even if you haven't configured authentication.

However, I'm finding this very difficult to review because of the number of different changes all combined into one patch:
* Switching to jinja2
* Enabling TLS configuration
* Cleaning up obsolete workarounds

There are a lot of subtleties in the ironic configuration (e.g. the same option being used by both api and conductor, but with different meanings), and 2-3 ways of running the services (both in the same container or in separate containers), only one of which is tested in CI. I think separating the changes is necessary in order to be able to review thoroughly enough to have confidence in the result.

maelk:
@zaneb Thank you for the explanation. I did not realize there was additional logic in runironic-conductor.sh. I have now split the jinja2 migration in its own PR : #201 . That PR should address your comments. This PR is then focusing on TLS setup only. This PR is based on #201, so only the second commit brings the TLS support.

configure-ironic.sh
```diff
@@ -1,5 +1,14 @@
 #!/usr/bin/bash
 
+if [ -n "$IRONIC_CERT_FILE" ] && [ -z "$IRONIC_KEY_FILE" ] ; then
+    echo "Missing TLS Certificate key file"
+    exit 1
+fi
+if [ -z "$IRONIC_CERT_FILE" ] && [ -n "$IRONIC_KEY_FILE" ] ; then
+    echo "Missing TLS Certificate file"
+    exit 1
+fi
```
zaneb, line 10: Maybe this is just me, but I don't see the need to do two lots of configuration - mounting the files somewhere and then telling the container where you've mounted them. You could just have a well-defined place where this script looks for the files, and you either mount them there or you don't. That eliminates the possibility of a mismatch, and the need to handle that case (which currently we just leave up to Ironic, by the looks of it).

ironic.conf.j2
```diff
@@ -77,12 +80,15 @@ fast_track = {{ env.IRONIC_FAST_TRACK }}
 dhcp_provider = none
 
 [inspector]
-endpoint_override = http://{{ env.IRONIC_URL_HOST }}:5050
+endpoint_override = {{ env.IRONIC_INSPECTOR_BASE_URL }}:5050
```
zaneb, line 83: If we're going to call this a URL, I kind of feel like the port number ought to be moved inside it.

zaneb (commented):
Thanks! This was *much* easier to review :)

configure-ironic.sh
maelk: ok, thanks for the idea, I will do that.

ironic.conf.j2
maelk: ok, I will update, thanks

maelk:
I have now defined specific locations for the certificates :

- /certs/ironic/crt
- /certs/ironic/key
- /certs/ironic-inspector/crt (if ca not provided)
- /certs/ca/ironic/ca
- /certs/ca/ironic-inspector/ca

The inspector CA defaults to inspector certificate if provided, otherwise to ironic CA (that itself defaults to ironic cert if not provided). Please take a look.

configure-ironic.sh
```diff
@@ -64,20 +69,68 @@ mkdir -p /shared/html
...
+#   i.e. API is authenticated. We want to authenticate RPC by default, but the user might override.
+#   Then try to infere the authentication strategy and credentials from /auth/ironic-rpc/auth-config.
+#   If not present, then generate a username and password, create the config file and set
+#   HTTP_BASIC_HTPASSWD_RPC
 if [ -n "${HTTP_BASIC_HTPASSWD}" ]; then
     printf "%s\n" "${HTTP_BASIC_HTPASSWD}" >"${HTPASSWD_FILE}"
+
+    if [ -z "${HTTP_BASIC_HTPASSWD_RPC:-}" ]; then
+
+        # When running both API and Conductor in the same container, we'll try to get the credentials
```
dtantsur, line 86: We should just stop doing it, really. It's an artefact of the early days of metal3 that we even allow that.

maelk: I agree, however, many things in our repo still rely on that script. I have some PR to change it, however, I would still maintain it for a while for compatibility. Some downstream product are relying on it, and might need time to adopt the new changes.

ironic-common.sh
```diff
@@ -11,6 +11,12 @@ function wait_for_interface_or_ip() {
       export IRONIC_IP=$(ip -br addr show | grep "${PROVISIONING_IP}" | grep -Po "[^\s]+/[0-9]+" | sed -e 's%/.*%%' | head -n 1)
       sleep 1
     done
+    # When an interface has multiple IP addresses, having IRONIC_IP set at this point means that the desired provisioning ip is set on the
+    # interface. However, the address returned might not be the desired one (no control over the order), so setting it back to the
+    # desired IP
+    if [ ! -z "${IRONIC_IP}" ]; then
+      export IRONIC_IP=${PROVISIONING_IP}
```
derekhiggins, line 18: I hit this error when trying out the patch with metal3
```
[root@master-1 core]# crictl logs 9e349ef4d8fdb |& grep 172.22.0.3                                                                                                                                                
Waiting for 172.22.0.3/24 to be configured on an interface
2020-09-23 14:41:03.656 1 DEBUG oslo_service.service [-] my_ip                          = 172.22.0.3/24 log_opt_values /usr/lib/python3.6/site-packages/oslo_config/cfg.py:2585                                   
2020-09-23 14:41:03.677 1 DEBUG oslo_service.service [-] conductor.bootloader           = http://172.22.0.3/24:6180/uefi_esp.img log_opt_values /usr/lib/python3.6/site-packages/oslo_config/cfg.py:2593          
2020-09-23 14:41:03.681 1 CRITICAL ironic [-] Unhandled error: ValueError: 172.22.0.3/24 is not IPv4 or IPv6 address                                                                                              
2020-09-23 14:41:03.681 1 ERROR ironic ValueError: 172.22.0.3/24 is not IPv4 or IPv6 address
```
I guess its down to PROVISIONING_IP being a CIDR
    - name: PROVISIONING_IP
      value: 172.22.0.3/24

maelk: That's weird, considering https://github.com/metal3-io/baremetal-operator/blob/master/tools/run_local_ironic.sh#L26 where that variable comes from https://github.com/metal3-io/metal3-dev-env/blob/master/lib/network.sh#L56 . PROVISIONING_IP is set to 172.22.0.2 . It does not have a CIDR while being passed down to the container. Are you using metal3-dev-env or something specific for Openshift ?

maelk: My guess is that the PROVISIONING_IP environment parameter is used differently in the Openshift version of Metal3-dev-env ? it sounds weird that something called PROVISIONING_IP actually contains a CIDR

derekhiggins: Yup, looks like in openshift its coming from the machine-api-operator, here
https://github.com/openshift/machine-api-operator/blob/1a2c56d/pkg/operator/baremetal_pod.go#L535
which gets a CIDR
https://github.com/openshift/machine-api-operator/blob/1a2c56d/pkg/operator/baremetal_config.go#L178
the same thing is passed into multiple different containers, I'll see what happens if I change it

maelk: I have added a sed command to remove the /24 in case it is present to maintain the compatibility for Openshift use case. please take a look

maelk:
/test-integration
@dtantsur @dhellmann @derekhiggins @hardys How does this look to you ?

derekhiggins:
I  tried this (hash 5d41bd365905aaaa56694d7ae8354b73bbc7a499) with openshift and am getting "No valid authentication is available" with json-rpc, the API ironic.conf has

```
[json_rpc]
auth_strategy = http_basic
http_basic_auth_user_file = /etc/ironic/htpasswd-rpc

host_ip = ::
```

I seem to be missing username/password for the API, I'll go back and try the patch that switches to jinja by itself to see what missing

maelk:
did you check the /auth/ironic-rpc/auth-config file ? that's where the configuration is for API to talk to RPC (it needs to be mounted to the container).
I did not touch that part of the code... https://github.com/metal3-io/ironic-image/pull/201/files#diff-80ccf82644ea197268ef4e8274fc59d5R92

ironic-common.sh
derekhiggins: thanks, I'll fix this in the other components but maintaining the sed for the moment helps

derekhiggins:
> did you check the /auth/ironic-rpc/auth-config file ? that's where the configuration is for API to talk to RPC (it needs to be mounted to the container).
> I did not touch that part of the code... https://github.com/metal3-io/ironic-image/pull/201/files#diff-80ccf82644ea197268ef4e8274fc59d5R92

Ya your probably correct, I think the mount might be wrong, looking into it and will report back

maelk:
rebased on top of the new commit in the Jinja PR.
/test-integration

maelk:
@zaneb @derekhiggins @dtantsur Now that the Jinja2 PR is merged, can we focus on this one and on the Inspector one ? Thank you!

zaneb:
> I tried this (hash [5d41bd3](https://github.com/metal3-io/ironic-image/commit/5d41bd365905aaaa56694d7ae8354b73bbc7a499)) with openshift and am getting "No valid authentication is available" with json-rpc, the API ironic.conf has

OpenShift doesn't configure basic-auth for json-rpc, so I'm not sure what could be happening here.
 
> ```
> [json_rpc]
> auth_strategy = http_basic
> http_basic_auth_user_file = /etc/ironic/htpasswd-rpc
> 
> host_ip = ::
> ```

This section is irrelevant to ironic-api (it's the server-side config for json-rpc; ironic-conductor is the server).

derekhiggins:
> > I tried this (hash [5d41bd3](https://github.com/metal3-io/ironic-image/commit/5d41bd365905aaaa56694d7ae8354b73bbc7a499)) with openshift and am getting "No valid authentication is available" with json-rpc, the API ironic.conf has

The problem was probably some version of what was being fixed by #196 , it appears to be irrelevant now as the jinja patch fixed it. So will close #196

derekhiggins:
> @zaneb @derekhiggins @dtantsur Now that the Jinja2 PR is merged, can we focus on this one and on the Inspector one ? Thank you!

Sorry I've been meaning to try this out over the last few days but got pulled into something else, will do it this afternoon.

maelk:
we'd like this PR and the inspector one to be merged soon, considering that the reviews were all addressed and it would unblock the whole bunch of PRs that build on it (12 PRs accross BMO, Metal3-dev-env and CAPM3) :D 

Could you please also take a look at the Inspector PR ? https://github.com/metal3-io/ironic-inspector-image/pull/62

derekhiggins (commented):
I deployed openshift with this and the TLS inspector-image. 

The callback urls in inspector.ipxe were http so inspection didn't work,
see comment on runhttpd.sh inline

Cleaning worked fine but when I tried to deploy IPA got killed by the OOM killer, have you seen this happen in your tests(although its probably not related to your patch)?

runhttpd.sh
```diff
@@ -12,10 +12,16 @@ wait_for_interface_or_ip
 mkdir -p /shared/html
 chmod 0777 /shared/html
 
+if [ -n "$CERT_FILE" ]; then
```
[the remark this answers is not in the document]
maelk: Kashif fixed it, please take a look

ironic.conf.j2
```diff
@@ -142,4 +151,10 @@ use_swift = false
 kernel_append_params = console=ttyS0 nofb nomodeset vga=normal ipa-insecure=1
 
 [service_catalog]
-endpoint_override = http://{{ env.IRONIC_URL_HOST }}:6385
+endpoint_override = {{ env.IRONIC_BASE_URL }}
+
+{% if env.IRONIC_TLS_SETUP == "true" %}
+[ssl]
+cert_file = /certs/ironic/tls.crt
+key_file = /certs/ironic/tls.key
```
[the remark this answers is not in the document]
maelk: same here

maelk:
Hi @derekhiggins ! Kashif addressed your comments. Regarding the OOM, no I have not seen it anywhere in my tests unfortunately, so I can't say anything there, sorry. Does this look good to you now ?

derekhiggins:
/lgtm

> Hi @derekhiggins ! Kashif addressed your comments. Regarding the OOM, no I have not seen it anywhere in my tests unfortunately, so I can't say anything there, sorry. Does this look good to you now ?

thanks for the updates, lgtm, I'm guessing the OOM error I saw was something to do with the image being deployed

PR #200, opened 2020-09-17
Add basic authentication setup for json-rpc

maelk:
The default htpasswd file is incorrect when using basic auth
for json rpc. This sets the configuration properly.

This approach goes against https://github.com/metal3-io/ironic-image/pull/196 . We have a requirement to authenticate all endpoints of Ironic, even if listening only on the loopback interface. So we would prefer this fix over removing the authentication on json_rpc.

/cc @dtantsur 
/cc @zaneb 
/cc @dhellmann

### DIFF
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -76,6 +76,7 @@ set_http_basic_server_auth_strategy() {
 # Configure HTTP basic auth for ironic-api server
 if [ -f "${HTPASSWD_FILE}" ]; then
     set_http_basic_server_auth_strategy
+    set_http_basic_server_auth_strategy json-rpc
 fi
 
 


configure-ironic.sh
```diff
@@ -76,6 +76,7 @@ set_http_basic_server_auth_strategy() {
 # Configure HTTP basic auth for ironic-api server
 if [ -f "${HTPASSWD_FILE}" ]; then
     set_http_basic_server_auth_strategy
+    set_http_basic_server_auth_strategy json-rpc
```
zaneb, line 79: This is currently done [in `runironic-conductor.sh`](https://github.com/metal3-io/ironic-image/blob/master/runironic-conductor.sh#L10), and afaik conductor is the only component doing the server side of json-rpc so I don't think it's necessary here.
Also, [it's `json_rpc` not `json-rpc`](https://opendev.org/openstack/ironic/src/branch/master/ironic/conf/json_rpc.py#L56).

maelk: the issue seems to be that we do not use runironic-conductor.sh in metal3-dev-env, we use runironic.sh. And in that case it is not set. Should I then do it in runironic.sh ? or here and remove it from runironic-conductor.sh ?

derekhiggins: Even in metal3 where we are using runironic-conductor.sh something appears to be wrong
see https://github.com/openshift-metal3/dev-scripts/issues/1099

auth isn't being turned on (that I can see) for rpc and the global DEFAULT is being used but not working as its not configured.

```
[derekh@r640-u07 dev-scripts]$ oc exec -it $POD -c metal3-ironic-api -- bash -c 'env | grep HTP ; ls /auth'
HTTP_BASIC_HTPASSWD=ironic-user:$2y$05$hdYBXIW3pBoKsTzcQTlAOedHKekwaSouvcjMPHsESnWQJLYd6pDry
ls: cannot access '/auth': No such file or directory
command terminated with exit code 2

[derekh@r640-u07 dev-scripts]$ oc exec -it $POD -c metal3-ironic-conductor -- bash -c 'env | grep HTP ; ls /auth'
ironic-inspector

[derekh@r640-u07 dev-scripts]$ oc exec -it $POD -c metal3-ironic-inspector -- bash -c 'env | grep HTP ; ls /auth'
HTTP_BASIC_HTPASSWD=inspector-user:$2y$05$e76JI1MKWw7NOLzSVBtXC.APUtNQlRtDcy2VBmbms0Di9zsVxEwvS
ironic

```

zaneb: > the issue seems to be that we do not use runironic-conductor.sh in metal3-dev-env, we use runironic.sh. And in that case it is not set.

Ah, yep, forgot about that. At the time I probably assumed that since the client and server were running in the same *container* in that case (so you can't even put them in different pods if you wanted), that you would never need to not bind to localhost (true) and also never need auth (not necessarily true - you might want to protect the API from other containers in the same pod).

> Should I then do it in runironic.sh ? or here and remove it from runironic-conductor.sh ?

There was a reason I had to do it in runironic-conductor.sh, but I don't remember what it was :/

Honestly we should get rid of runironic.sh if it's humanly possible. Failing that, we should probably configure auth all the time from within runironic.sh - since we have both ends of the connection right there, there's no need to wait for someone to configure it externally.

maelk:
ok, I'll then go with changing the deployment of Ironic to use two containers, one for api and one for conductor

PR #201, opened 2020-09-20
Switch ironic configuration rendering to Jinja2

maelk:
This centralizes the cryptic configuration done in multiple files. It also removes all calls to crudini, that was remove options set in a section if it was set to the same value in [DEFAULT].

The type of deployment of Ironic is set at the beginning of each script in IRONIC_DEPLOYMENT : 

- runironic.sh -> Combined
- runironic-api.sh -> API
- runironic-conductor.sh -> Conductor

The configuration file is templated based on this variable.

In addition, it configures json_rpc to be authenticated when API and conductor are running in the same container and
HTTP_BASIC_HTPASSWD is set. API and json_rpc would accept the same credentials. This is to prevent other processes in the same host to access the RPC interface over localhost. The LISTEN_ALL_INTERFACES environment variable is added to allow restricting on which IP address ironic listens to for API and RPC.


maelk:
This should make #200 unnecessary.

ironic.conf.j2
```diff
@@ -1,5 +1,10 @@
 [DEFAULT]
+{% if "HTTP_BASIC_HTPASSWD" in env %}
```
dhellmann, line 2: Should we check that the value isn't empty?

dhellmann (commented):
This looks OK, but I'm starting to wonder if we would be better off building an operator to manage Ironic instead of putting all of this logic in the container entry point.

One question inline about how the auth strategy is detected.

ironic.conf.j2
maelk: good point. I'll update.

ironic.conf.j2
```diff
@@ -33,29 +47,81 @@ deploy_logs_local_path = /shared/log/ironic/deploy
...
+
+
+[json_rpc]
+# We assume that when we run API and conductor in the same container, they use
+# authentication over localhost, using the same credentials as API, to prevent
+# unauthenticated connections from other processes in the same host since the
+# containers are in host networking.
+{% if ( env.IRONIC_DEPLOYMENT == "Combined" or env.IRONIC_DEPLOYMENT == "Conductor" ) and ( "HTTP_BASIC_HTPASSWD" in env and env.HTTP_BASIC_HTPASSWD | length ) %}
+auth_strategy = http_basic
+http_basic_auth_user_file = /etc/ironic/htpasswd
```
zaneb, line 112: This is better than nothing for now, but if the combined container is going to continue to be a thing then I think we want to use separate credentials for json-rpc (could easily be generated inside the container, to avoid complicating the interface).

zaneb (commented):
This looks good to me.

ironic.conf.j2
maelk: I have some on-going work in BMO to stop using the combined version. I'd like to keep support for it in this image for now to make sure that it is backwards compatible, but when the BMO change will have been there for a while, we can move to deprecate it.

maelk: I will add json-rpc credentials generation

maelk:
I have now added the credentials generation for the RPC. I have considered the following use cases. I have introduced a new variable HTTP_BASIC_HTPASSWD_RPC. The user can provide HTTP_BASIC_HTPASSWD and HTTP_BASIC_HTPASSWD_RPC. If

- we are running conductor and HTTP_BASIC_HTPASSWD_RPC is not set but HTTP_BASIC_HTPASSWD is,
  fallback to use HTTP_BASIC_HTPASSWD for RPC, to preserve compatibility.
- we are running combined API and conductor, and HTTP_BASIC_HTPASSWD_RPC is set, use it
- we are running combined API and conductor, and HTTP_BASIC_HTPASSWD_RPC is unset, but HTTP_BASIC_HTPASSWD 
  is, i.e. API is authenticated. We want to authenticate RPC by default, but the user might override with auth-config files.
  Then try to infere the authentication strategy and credentials from /auth/ironic-rpc/auth-config.
  If not present, then generate a username and password, create the config file and set
  HTTP_BASIC_HTPASSWD_RPC

It is implemented in the third commit of this PR.
@zaneb How does this look to you ?

maelk:
/test-integration
@dtantsur @dhellmann @derekhiggins @hardys How does this look to you ?

dhellmann:
This looks clean to me. I think the other reviewers are more familiar with the way the actual settings are used, so I'll leave it open for one of them to look at.

/approve

configure-ironic.sh
```diff
@@ -34,10 +27,69 @@ mkdir -p /shared/html
...
+                    IRONIC_RPC_TMP_USERNAME="$(crudini --get /auth/ironic-rpc/auth-config json-rpc username)" || exit 1
+                    IRONIC_RPC_TMP_PASSWORD="$(crudini --get /auth/ironic-rpc/auth-config json-rpc password)" || exit 1
+                fi
+            # We do not have an auth config file, so we generate one
+            else
+                IRONIC_RPC_TMP_USERNAME="$(tr -dc 'a-zA-Z0-9' < /dev/urandom | fold -w 12 | head -n 1)"
+                IRONIC_RPC_TMP_PASSWORD="$(tr -dc 'a-zA-Z0-9' < /dev/urandom | fold -w 12 | head -n 1)"
+                mkdir -p "/auth/ironic-rpc"
+                cat << EOF > "/auth/ironic-rpc/auth-config"
+[json-rpc]
```
zaneb, line 62: Maybe both work(?), but it's `json_rpc` [in the code](https://opendev.org/openstack/ironic/src/branch/master/ironic/conf/json_rpc.py#L55). I think we should standardise on that, because in other places (like the crudini commands on line 53-54) it probably does matter whether it's `-` or `_`.

configure-ironic.sh
```diff
@@ -64,20 +27,68 @@ mkdir -p /shared/html
...
+                if [ "${IRONIC_RPC_TMP_STRATEGY}" == "basic" ]; then
+                    IRONIC_RPC_TMP_USERNAME="$(crudini --get /auth/ironic-rpc/auth-config json-rpc username)" || exit 1
+                    IRONIC_RPC_TMP_PASSWORD="$(crudini --get /auth/ironic-rpc/auth-config json-rpc password)" || exit 1
+                fi
+            # We do not have an auth config file, so we generate one
+            else
+                IRONIC_RPC_TMP_USERNAME="$(tr -dc 'a-zA-Z0-9' < /dev/urandom | fold -w 12 | head -n 1)"
+                IRONIC_RPC_TMP_PASSWORD="$(tr -dc 'a-zA-Z0-9' < /dev/urandom | fold -w 12 | head -n 1)"
+                mkdir -p "/auth/ironic-rpc"
+                cat << EOF > "/auth/ironic-rpc/auth-config"
```
zaneb, line 61: One thing I didn't think about before is that we're writing this to a file that is not on a tmpfs mount (like secrets are), so it could get written to disk at some point. Probably not a reason not to do this, but bindmounting it in from a secret is definitely more secure.

configure-ironic.sh
```diff
@@ -34,10 +27,69 @@ mkdir -p /shared/html
...
+            # to ensure that the setup will work if the user gives a specific config for rpc set to no_auth
+            if [ -f "/auth/ironic-rpc/auth-config" ]; then
+                IRONIC_RPC_TMP_STRATEGY="$(crudini --get /auth/ironic-rpc/auth-config json-rpc auth_strategy)" || exit 1
+                if [ "${IRONIC_RPC_TMP_STRATEGY}" == "basic" ]; then
+                    IRONIC_RPC_TMP_USERNAME="$(crudini --get /auth/ironic-rpc/auth-config json-rpc username)" || exit 1
+                    IRONIC_RPC_TMP_PASSWORD="$(crudini --get /auth/ironic-rpc/auth-config json-rpc password)" || exit 1
+                fi
+            # We do not have an auth config file, so we generate one
+            else
+                IRONIC_RPC_TMP_USERNAME="$(tr -dc 'a-zA-Z0-9' < /dev/urandom | fold -w 12 | head -n 1)"
```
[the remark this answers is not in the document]
maelk: ok, I'll change it!

configure-ironic.sh
```diff
@@ -34,10 +27,69 @@ mkdir -p /shared/html
...
+# - we are running combined API and conductor, and HTTP_BASIC_HTPASSWD_RPC is set, use it
+# - we are running combined and HTTP_BASIC_HTPASSWD_RPC is unset, but HTTP_BASIC_HTPASSWD is,
+#   i.e. API is authenticated. We want to authenticate RPC by default, but the user might override.
+#   Then try to infere the authentication strategy and credentials from /auth/ironic-rpc/auth-config.
+#   If not present, then generate a username and password, create the config file and set
+#   HTTP_BASIC_HTPASSWD_RPC
 if [ -n "${HTTP_BASIC_HTPASSWD}" ]; then
     printf "%s\n" "${HTTP_BASIC_HTPASSWD}" >"${HTPASSWD_FILE}"
+
+    if [ -z "${HTTP_BASIC_HTPASSWD_RPC:-}" ]; then
```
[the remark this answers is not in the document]
maelk: ok, let me give it try.

configure-ironic.sh
maelk: that's definitely a typo! thanks for the catch.

configure-ironic.sh
maelk: I agree that it is not great, but if we want to differentiate the RPC username/password from the API ones, then in the combined case, there is no other option that writing to disk if the user does not provide the credentials. If the user provides the auth-config, then that's more secure.

configure-ironic.sh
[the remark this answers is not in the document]
maelk: I have now removed that variable. I kept the behaviour that in case of "Combined" we try to get the authentication info from the auth-config and if it is not provided, we generate some

maelk:
@zaneb Thank you for the comments. I updated the PR.
/test-integration

derekhiggins:
/approve

works with installer, tested without and also with a rpc credentials passed in

maelk:
Great! Is this ready to go in then ? If so please lgtm.

zaneb:
LGTM but I am not a reviewer for this repo.

maelk:
@dhellmann or @dtantsur Would you mind lgtm if lgty ? thank you!

PR #202, opened 2020-10-06
Make searching for Provisioning IP robust

zaneb:
Using grep with a user-supplied query that is likely to contain `.`
characters is not the most robust method of searching for a string in
some output. For example, passing `PROVISIONING_IP=192.168.1.2` would also
match e.g. 192.168.122.1. Avoid this by searching for a fixed string,
including the characters before and after the IP address.

The best way to do this would be using the json output from the ip
command:
```
    [ -n $(ip -j addr show | jq -r ".[] | .addr_info[] | .local | select(. == ${IRONIC_IP})") ]
```
and something similar to get the IP address when only the provisioning
interface is specfied:
```
    ip -j add show scope global up dev ${PROVISIONING_INTERFACE} | jq -r .[0].addr_info[0].local
```
but continuing to do it with grep avoids having to install the `jq` tool.

This also avoids the weird dance where we set `IRONIC_IP` only to
overwrite it with the value we had at the beginning, and also avoids
adding a minimum 1s delay.

### DIFF
--- a/ironic-common.sh
+++ b/ironic-common.sh
@@ -5,18 +5,11 @@ function wait_for_interface_or_ip() {
   # If $PROVISIONING_IP is specified, then we wait for that to become available on an interface, otherwise we look at $PROVISIONING_INTERFACE for an IP
   if [ ! -z "${PROVISIONING_IP}" ];
   then
-    export IRONIC_IP=""
-    until [ ! -z "${IRONIC_IP}" ]; do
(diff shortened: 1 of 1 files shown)


maelk (commented):
looks good. it should probably be applied to inspector image too.

zaneb:
metal3-io/ironic-inspector-image#63 is the inspector-image PR

dtantsur (approved):

PR #204, opened 2020-10-13
Sync up

derekhiggins:
Syncing up patches from the downstream (openshift) image to here.


elfosardo (approved):

ironic.conf.j2
```diff
@@ -58,7 +58,7 @@ enable_ssl_api = true
 automated_clean = {{ env.IRONIC_AUTOMATED_CLEAN }}
 # NOTE(dtantsur): keep aligned with [pxe]boot_retry_timeout below.
 deploy_callback_timeout = 4800
-send_sensor_data = true
+send_sensor_data = false
```
dtantsur, line 61: I don't think we should change this.

main-packages-list.txt
```diff
@@ -0,0 +1,24 @@
+crudini
+dnsmasq
+gdisk
+genisoimage
+httpd
+httpd-tools
+ipmitool
+iproute
+iscsi-initiator-utils
```
dtantsur, line 9: I wonder why we have this, we don't use the iscsi deploy..

dtantsur (approved):

ironic.conf.j2
derekhiggins: should change it at all, or upstream?

Issue #203, opened 2020-10-09
dhcp-sequential-ip issues


maelk:
In order to prevent collisions in case of burst, we enabled dhcp-sequential-ip for dnsmasq. However, this turned out to be even more problematic in bigger scale deployments. 

In a case of 30 nodes deployment, when the bmhs are created all at the same time, or the deployment is triggered (in CAPI all workers start to provision at the same time), then we see many issues, where dnsmasq does not answer to a DISCOVER from a node and prints "no address available" in the logs. After several retries the node usually gets an ip, but some of the PXE firmware do not retry enough to get the ip and instead fall back to boot from HDD, failing the introspection.

In addition, when booting from HDD, if the client already had an ip (for example .250), then it will request it directly upon reboot, dnsmasq will ack it, but then keep allocating addresses from there (.251, .252, .253 etc.) . 

Overall I think we should remove this option. It seemed to fix a collision issue, but actually caused more problems in "big" scale deployments.

dhellmann:
https://github.com/metal3-io/metal3-docs/pull/139 may also help with this problem.

maelk:
agreed, but meanwhile I am afraid we might have introduced something that could cause quite some issues to users, compared to the previous situation

dhellmann:
> agreed, but meanwhile I am afraid we might have introduced something that could cause quite some issues to users, compared to the previous situation

Sure, I agree that rolling back the sequential IP setting also makes sense.

PR #205, opened 2020-10-14
Revert dhcp-sequential-ip

fmuyassarov:
Fixes https://github.com/metal3-io/ironic-image/issues/203.


derekhiggins:
/approve
this also seems to have caused problems when using PXE over IPv6.

PR #206, opened 2020-10-20
Make Ironic connect to Mariadb using TLS

namnx228:
This PR is related to:
https://github.com/metal3-io/metal3-dev-env/pull/525
https://github.com/metal3-io/baremetal-operator/pull/689

This PR needs to be merged before the two PRs above can pass CI tests

### DIFF
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -5,6 +5,7 @@ export IRONIC_KEY_FILE=/certs/ironic/tls.key
 export IRONIC_CACERT_FILE=/certs/ca/ironic/tls.crt
 export IRONIC_INSPECTOR_CERT_FILE=/certs/ironic-inspector/tls.crt
 export IRONIC_INSPECTOR_CACERT_FILE=/certs/ca/ironic-inspector/tls.crt
+export MARIADB_CACERT_FILE=/certs/ca/mariadb/tls.crt
 
 mkdir -p /certs/ironic
 mkdir -p /certs/ironic-inspector
@@ -63,6 +64,12 @@ if [ -f "$IRONIC_INSPECTOR_CERT_FILE" ] || [ -f "$IRONIC_INSPECTOR_CACERT_FILE" 
 else
     export IRONIC_INSPECTOR_TLS_SETUP="false"
     export IRONIC_INSPECTOR_BASE_URL="http://${IRONIC_URL_HOST}:5050"
+fi
+
+if  [ -f "$MARIADB_CACERT_FILE" ]; then
+    export MARIADB_TLS_ENABLED="true"
+else
+    export MARIADB_TLS_ENABLED="false"
 fi
 
 cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig
--- a/ironic.conf.j2
+++ b/ironic.conf.j2
@@ -70,7 +70,13 @@ send_sensor_data_interval = 160
 bootloader = http://{{ env.IRONIC_URL_HOST }}:{{ env.HTTP_PORT }}/uefi_esp.img
 
 [database]
-connection = mysql+pymysql://ironic:{{ env.MARIADB_PASSWORD }}@localhost/ironic?charset=utf8
+{% if env.MARIADB_TLS_ENABLED == "true" %}
+connection = mysql+pymysql://ironic:{{ env.MARIADB_PASSWORD }}@127.0.0.1/ironic?charset=utf8&ssl=on&ssl_ca={{ env.MARIADB_CACERT_FILE }}
+{% else %}
+connection = mysql+pymysql://ironic:{{ env.MARIADB_PASSWORD }}@127.0.0.1/ironic?charset=utf8
+{% endif %}
+
+
 
 [deploy]
 default_boot_option = local
--- a/runmariadb.sh
+++ b/runmariadb.sh
@@ -3,6 +3,18 @@ PATH=$PATH:/usr/sbin/
 DATADIR="/var/lib/mysql"
 MARIADB_PASSWORD=${MARIADB_PASSWORD:-"change_me"}
 MARIADB_CONF_FILE="/etc/my.cnf.d/mariadb-server.cnf"
+MARIADB_CERT_FILE=/certs/mariadb/tls.crt
+MARIADB_KEY_FILE=/certs/mariadb/tls.key
+
+mkdir -p $(dirname ${MARIADB_CERT_FILE})
+if [ -f "$MARIADB_CERT_FILE" ] && [ ! -f "$MARIADB_KEY_FILE" ] ; then
+    echo "Missing TLS private key file ${MARIADB_KEY_FILE}"
+    exit 1
+fi
+if [ ! -f "$MARIADB_CERT_FILE" ] && [ -f "$MARIADB_KEY_FILE" ] ; then
+    echo "Missing TLS Certificate file ${MARIADB_CERT_FILE}"
+    exit 1
+fi
 
 ln -sf /proc/self/fd/1 /var/log/mariadb/mariadb.log
 
@@ -11,6 +23,14 @@ if [ ! -d "${DATADIR}/mysql" ]; then
     crudini --set "$MARIADB_CONF_FILE" mysqld max_heap_table_size 1M
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_buffer_pool_size 5M
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_log_buffer_size 512K
+
+    # Config MariaDB to enable TLS
+    if [ -f "$MARIADB_CERT_FILE" ]
+    then
+      crudini --set "$MARIADB_CONF_FILE" mariadb-10.3 ssl on
+      crudini --set "$MARIADB_CONF_FILE" mariadb-10.3 ssl_cert "${MARIADB_CERT_FILE}"
+      crudini --set "$MARIADB_CONF_FILE" mariadb-10.3 ssl_key "${MARIADB_KEY_FILE}"
+    fi
 
     mysql_install_db --datadir="$DATADIR"
 


ironic.conf.j2
```diff
@@ -66,7 +66,8 @@ send_sensor_data_interval = 160
 bootloader = http://{{ env.IRONIC_URL_HOST }}:{{ env.HTTP_PORT }}/uefi_esp.img
 
 [database]
-connection = mysql+pymysql://ironic:{{ env.MARIADB_PASSWORD }}@localhost/ironic?charset=utf8
+connection = mysql+pymysql://ironic:{{ env.MARIADB_PASSWORD }}@127.0.0.1/ironic?charset=utf8&ssl=on&ssl_ca=/certs/ca/ironic/tls.crt
```
maelk, line 69: you should parameterize this, we do not always deploy with ssl. You need to find out if mariadb uses ssl, and then add those parameters only in that case

runmariadb.sh
```diff
@@ -12,6 +12,18 @@ if [ ! -d "${DATADIR}/mysql" ]; then
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_buffer_pool_size 5M
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_log_buffer_size 512K
 
+    # Config MariaDB to enable TLS
+    if [ -d "/certs" ]
```
maelk, line 16: you should maybe verify that they are all there individually, and error in case one would be missing. The current way might make it more difficult to debug if mariadb refuses to start due to a missing file.

runmariadb.sh
```diff
@@ -12,6 +12,18 @@ if [ ! -d "${DATADIR}/mysql" ]; then
...
+    # Config MariaDB to enable TLS
+    if [ -d "/certs" ] 
+    then
+      mkdir -p "${DATADIR}/certs/ca/"
+      cp "/certs/mariadb" "${DATADIR}/certs/" -r
+      cp "certs/ca/mariadb" "${DATADIR}/certs/ca/" -r
+      crudini --set "$MARIADB_CONF_FILE" mariadb-10.3 ssl on
+      crudini --set "$MARIADB_CONF_FILE" mariadb-10.3 ssl_cert "${DATADIR}/certs/mariadb/tls.crt"
+      crudini --set "$MARIADB_CONF_FILE" mariadb-10.3 ssl_key "${DATADIR}/certs/mariadb/tls.key"
+      crudini --set "$MARIADB_CONF_FILE" mariadb-10.3 ssl_ca "${DATADIR}/certs/ca/mariadb/tls.crt"
```
maelk, line 24: I don't think we need the CA here, we do not use mTLS.

maelk (changes requested):

runmariadb.sh
namnx228: True, we don't need it now. But in the future we may need mTLS, so should we keep the CA here since it doesn't hurt anyone?

namnx228:
@maelk all the required changes have been made. Please check when you have time. Thanks

configure-ironic.sh
```diff
@@ -123,11 +123,19 @@ EOF
     fi
 fi
 
+add_options_to_j2_config () {
+  if [[ "${IRONIC_TLS_SETUP}" == "true" ]]
+  then
+  crudini --set "/etc/ironic/ironic.conf.j2" Database connection "mysql+pymysql://ironic:{{ env.MARIADB_PASSWORD }}@127.0.0.1/ironic?charset=utf8&ssl=on&ssl_ca=/certs/ca/ironic/tls.crt"
```
maelk, line 129: I don't think using crudini here is the correct way to do it. Jinaj2 is a templating language, so you can just use an if/else structure within the ironic.conf.j2, rather than this

configure-ironic.sh
```diff
@@ -123,11 +123,19 @@ EOF
     fi
 fi
 
+add_options_to_j2_config () {
+  if [[ "${IRONIC_TLS_SETUP}" == "true" ]]
```
maelk, line 127: you should probably not use IRONIC_TLS_SETUP here. you're binding the database TLS setup to the existence of the API or RPC certificates. you should instead make a separate variable for Mariadb, based on the mariadb certificates

ironic.conf.j2
maelk: use an if else structure here for example

runmariadb.sh
```diff
@@ -12,6 +12,38 @@ if [ ! -d "${DATADIR}/mysql" ]; then
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_buffer_pool_size 5M
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_log_buffer_size 512K
 
+    # Config MariaDB to enable TLS
+    if [[  ${IRONIC_TLS_SETUP} == "true" ]]
```
maelk, line 16: same thing here. and the check on the files could be done when setting the mariadb specific variable.

namnx228:
Hi @maelk,
All the changes have been made. Please take a look.

namnx228:
Hi @maelk,
All issues solved!
Please take a look. Thanks

ironic.conf.j2
maelk: The IRONIC_TLS_SETUP variable is tied to a file that is unrelated to mariadb, so you should tie this to the presence or not of the mariadb CA cert. You can create a new variable for that

namnx228: Done it! Thanks

configure-ironic.sh
```diff
@@ -5,11 +5,16 @@ export IRONIC_KEY_FILE=/certs/ironic/tls.key
 export IRONIC_CACERT_FILE=/certs/ca/ironic/tls.crt
 export IRONIC_INSPECTOR_CERT_FILE=/certs/ironic-inspector/tls.crt
 export IRONIC_INSPECTOR_CACERT_FILE=/certs/ca/ironic-inspector/tls.crt
+export MARIADB_CERT_FILE=/certs/mariadb/tls.crt
+export MARIADB_KEY_FILE=/certs/mariadb/tls.key
```
maelk, line 9: from ironic point of view (this file is only for ironic, not mariadb) the only thing we need is either the CA or the actual certificate itself. in order to support seamlessly rotation of certificates from cert-manager, we should rather use the CA. so you can replace the use of both file by the single use of the CA certificate only. And you should refer to that CA in ironic.conf.j2, and that would be a separate CA from Ironic's, like we have a different CA for inspector. If the user wants to use a single CA for all components, they can just mount the same secret multiple times, but we should probably support different CAs for each component.

namnx228: As mTLS is not supported now, even we have a CA for MariaDB, but currently MariaDB only needs to know its key and certificate to enable TLS. So it should be better if Ironic refers to these files to determine whether should it enable TLS with MariaDB. This reference is only to the file location, to check if files exist, but it doesn't care about the content, so I bet it should be fine when the certificate is replaced.

runmariadb.sh
```diff
@@ -12,6 +24,26 @@ if [ ! -d "${DATADIR}/mysql" ]; then
...
+
+      # If mTLS is enabled -> Uncomment the below section
+        # if [[ ! -a "/certs/ca/mariadb/tls.crt" ]]
+        # then
+        #   echo "CA certificate not found"
+        #   exit 0
+        # fi
+        # mkdir -p "${DATADIR}/certs/ca/"
+        # cp "certs/ca/mariadb" "${DATADIR}/certs/ca/" -r
+        # crudini --set "$MARIADB_CONF_FILE" mariadb-10.3 ssl_ca "${DATADIR}/certs/ca/mariadb/tls.crt"
```
maelk, line 44: nit : maybe remove this ?

ironic.conf.j2
```diff
@@ -66,7 +66,13 @@ send_sensor_data_interval = 160
 bootloader = http://{{ env.IRONIC_URL_HOST }}:{{ env.HTTP_PORT }}/uefi_esp.img
 
 [database]
-connection = mysql+pymysql://ironic:{{ env.MARIADB_PASSWORD }}@localhost/ironic?charset=utf8
+{% if env.MARIADB_TLS_ENABLED == "true" %}
+connection = mysql+pymysql://ironic:{{ env.MARIADB_PASSWORD }}@127.0.0.1/ironic?charset=utf8&ssl=on&ssl_ca=/certs/ca/mariadb/tls.crt
```
maelk, line 70: nit: we could use the env variable to get the certificate path, to have it configured in one place only

maelk (commented):
some small nits, otherwise looks good to me

runmariadb.sh
namnx228: Done!

ironic.conf.j2
namnx228: Fixed! Please take a look again. Thanks

configure-ironic.sh
```diff
@@ -62,6 +66,15 @@ else
     export IRONIC_INSPECTOR_BASE_URL="http://${IRONIC_URL_HOST}:5050"
 fi
 
+if [ -f "$MARIADB_CERT_FILE" ] || [ -f "$MARIADB_CACERT_FILE" ]; then
+    export MARIADB_TLS_ENABLED="true"
+    if [ ! -f "$MARIADB_CACERT_FILE" ]; then
```
zaneb, line 71: This is a bit weird that we have two files you can add, but we only use one of them, but I guess it makes sense in that it gives you the option to mount the same volume as is attached to the mariadb container if you are feeling lazy. Does that actually work though? Can you use the server certificate file as the CA cert? Or is it only working because we didn't actually enable CA verification?

runmariadb.sh
```diff
@@ -3,6 +3,18 @@ PATH=$PATH:/usr/sbin/
 DATADIR="/var/lib/mysql"
 MARIADB_PASSWORD=${MARIADB_PASSWORD:-"change_me"}
 MARIADB_CONF_FILE="/etc/my.cnf.d/mariadb-server.cnf"
+MARIADB_CERT_FILE=/certs/mariadb/tls.crt
+MARIADB_KEY_FILE=/certs/mariadb/tls.key
+
+mkdir -p /certs/mariadb
+if [ -f "$MARIADB_CERT_FILE" ] && [ ! -f "$MARIADB_KEY_FILE" ] ; then
+    echo "Missing TLS private key file /certs/mariadb/tls.key"
+    exit 0
```
zaneb, line 12: If there's ever a time for `exit 1` surely this is it.

ironic.conf.j2
```diff
@@ -66,7 +66,13 @@ send_sensor_data_interval = 160
 bootloader = http://{{ env.IRONIC_URL_HOST }}:{{ env.HTTP_PORT }}/uefi_esp.img
 
 [database]
-connection = mysql+pymysql://ironic:{{ env.MARIADB_PASSWORD }}@localhost/ironic?charset=utf8
+{% if env.MARIADB_TLS_ENABLED == "true" %}
+connection = mysql+pymysql://ironic:{{ env.MARIADB_PASSWORD }}@127.0.0.1/ironic?charset=utf8&ssl=on&ssl_ca={{ env.MARIADB_CACERT_FILE }}
```
zaneb, line 70: Do we need to explicitly set `ssl_verify_cert`?

runmariadb.sh
```diff
@@ -12,6 +24,16 @@ if [ ! -d "${DATADIR}/mysql" ]; then
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_buffer_pool_size 5M
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_log_buffer_size 512K
 
+    # Config MariaDB to enable TLS
+    if [ -f "$MARIADB_CERT_FILE" ]
+    then
+      mkdir -p "${DATADIR}/certs/"
+      cp "/certs/mariadb" "${DATADIR}/certs/" -r
```
zaneb, line 31: What is the point of this? It has the effect of copying the private key to a volume where it may get persisted to disk.

namnx228: Thanks for your review. The reason for this copying to the database directory is that the certificate and key needs to be owned by user `mysql`. Otherwise, the database cannot read these files. I am not sure if  it is a good idea if we change the ownership of files in a mounted volume. Besides, TLS can only be enabled when the volume is mounted, so I am not sure if we gain any security benefits when not storing the key in the container. What is your idea about this, @zaneb ?

configure-ironic.sh
```diff
@@ -62,6 +66,12 @@ else
     export IRONIC_INSPECTOR_BASE_URL="http://${IRONIC_URL_HOST}:5050"
 fi
 
+if [ -f "$MARIADB_CERT_FILE" ] && [ -f "$MARIADB_CACERT_FILE" ]; then
```
maelk, line 69: If you are not default the CA to the actual cert, then you should remove the cert file from this condition, as the ca file would be empty

namnx228: We don't need to do that. If Ironic cannot find the MariaDB CA certificate using the provided link, it will automatically connect without TLS. This behaviour has been tested

maelk: I was not clear with my comment, sorry. You are not using the cert file in ironic, because you use the CA. Why would you require to have both the cert itself and the CA when you use only the CA. (I meant cert file, not CA file in the previous comment). You are putting a dependency on a file that you do not use. If you remove this then you don't need to mount the cert secret at all, just the CA

configure-ironic.sh
namnx228: You are right. The normal certificate cannot be used as a CA certificate. SQLAlchemy will complain if it cannot verify the CA certificate. I have fixed it. Good catch, @zaneb

runmariadb.sh
namnx228: Fixed! Thanks

ironic.conf.j2
namnx228: I think the option `ssl_verify_server_cert` is required if we use mTLS, which is not in our case.

configure-ironic.sh
namnx228: Oh I get your idea. I thought you were saying that we should delete the certificate file. Sorry for that. Updated.

runmariadb.sh
namnx228: Your clarification makes sense that we can gain secuirty benefit when not storing the key file on disk. 
The reason why we needed the key file being owned by `mysql` is that its default permission is 600 (only owner can read and write).
```sh 
stat -c '%a' /certs/mariadb/tls.key
600
```
What if we change the permission of the key file to 604 before mount it to the container? WDYT @zaneb @maelk ?

ironic.conf.j2
namnx228: I see. That is fixed! Thanks

runmariadb.sh
zaneb: Thanks for the clarification. I agree the answer is to mount it with the right permissions, rather than try to change it in the container.
How are you mounting the secret now? This should just work by default when mounted from a Secret. e.g. I checked with the existing mounts used for basic auth and the default mode is 644:

```
sh-4.4# ls -Hl /auth/ironic/password
-rw-r--r--. 1 root root 16 Oct 22 14:29 /auth/ironic/password
```

namnx228: From metal3-dev-env, we have two options to mount the key to the container:
- Using secret: I didn't check if the key has the permission 644 by default. I agree with you that by default the file mounted by a secret has permission 644, but the private key file (generated by Openssl, for example) is explicitly set to 600 when it is created. I am not sure by being mounted to a secret, this permission can be changed to 644.

zaneb: For a Secret the permissions on the file generated by openssl are irrelevant, since you're going to copy the contents of the file into the Secret. What matters is [the permissions for the mounted Secret](https://kubernetes.io/docs/concepts/configuration/secret/#secret-files-permissions) which can be modified but default to 644 which is what we want anyway.

If you use a bind mount then I agree that you will need to manually set the appropriate mode on the file. However, nobody would ever use this method in a real deployment, and for dev-env purposes we can just put it in the script.

namnx228: The bind mount is used when Ironic run standalone outside of the K8s cluster. 
Anyway, I have updated the PR. Please take a look. Thanks

ironic.conf.j2
namnx228: I have tested the use of `ssl_verify_server_cert`. Interestingly, we cannot add this option because Ironic denies this parameter. It seems like Ironic doesn't directly pass the parameter to the mariaDB client configuration. However, just `ssl=on` is enough since SQLAlchemy by default verifies the certificate. This behaviour has been tested. I changed the target IP address to `127.0.1.1` and this is the error:
```
ssl.CertificateError: hostname '127.0.1.1' doesn't match '127.0.0.1'
: ssl.CertificateError: hostname '127.0.1.1' doesn't match '127.0.0.1'
```

zaneb: Ah, I had been looking at the docs for the [python connector for mysql](https://dev.mysql.com/doc/connector-python/en/connector-python-connectargs.html), where the option is called `ssl_verify_cert` and defaults to False.

However, what I failed to realise is that we are actually using pymysql, and [in pymysql](https://github.com/PyMySQL/PyMySQL/blob/v0.10.1/pymysql/connections.py#L336) the option is called `check_hostname` and it defaults to True if a CA is specified (which is the only sane way of implementing SSL). Good to see that testing confirms this works.

FWIW the `ssl=on` part [doesn't appear to do anything](https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_3_20/lib/sqlalchemy/dialects/mysql/mysqldb.py#L203-L211).

configure-ironic.sh
```diff
@@ -5,11 +5,13 @@ export IRONIC_KEY_FILE=/certs/ironic/tls.key
 export IRONIC_CACERT_FILE=/certs/ca/ironic/tls.crt
 export IRONIC_INSPECTOR_CERT_FILE=/certs/ironic-inspector/tls.crt
 export IRONIC_INSPECTOR_CACERT_FILE=/certs/ca/ironic-inspector/tls.crt
+export MARIADB_CACERT_FILE=/certs/ca/mariadb/tls.crt
 
 mkdir -p /certs/ironic
 mkdir -p /certs/ironic-inspector
 mkdir -p /certs/ca/ironic
 mkdir -p /certs/ca/ironic-inspector
+mkdir -p /certs/ca/mariadb
```
derekhiggins, line 14: I don't think this is needed, if the CA cert file doesn't exist we aren't using this directory

runmariadb.sh
```diff
@@ -3,6 +3,18 @@ PATH=$PATH:/usr/sbin/
...
+MARIADB_CERT_FILE=/certs/mariadb/tls.crt
+MARIADB_KEY_FILE=/certs/mariadb/tls.key
+
+mkdir -p $(dirname ${MARIADB_CERT_FILE})
+if [ -f "$MARIADB_CERT_FILE" ] && [ ! -f "$MARIADB_KEY_FILE" ] ; then
+    echo "Missing TLS private key file ${MARIADB_KEY_FILE}"
+    exit 1
+fi
+if [ ! -f "$MARIADB_CERT_FILE" ] && [ -f "$MARIADB_KEY_FILE" ] ; then
+  echo "Missing TLS Certificate file ${MARIADB_CERT_FILE}"
```
derekhiggins, line 15: nit: inconsistent indent here

derekhiggins (commented):
can we squash these into one commit? 2 of them are improvements on the original commit.

configure-ironic.sh
namnx228: Fixed! Thanks

runmariadb.sh
namnx228: Fixed! Thanks

ironic.conf.j2
maelk: Even if ssl=on does not do anything, I think it can be valuable to have it here since it makes it clear that we are configuring the connection with ssl for whoever reads this configuration file

namnx228:
Thank you @derekhiggins, @zaneb , @maelk for your review.
@dhellmann @dtantsur could you please take a look and consider if it is ready to be merged or not? Thanks

dtantsur (approved):

PR #207, opened 2020-10-23
Make sensor_data collection configurable

iurygregory:
Previously we disabled for some reason, but we had people using
the ironic prometheus exporter in Metal3.

Add a new env variable `SEND_SENSOR_DATA` that will be used to
define if it will be enabled or not.


maelk:
Can we make this configurable through the environment variables ?

iurygregory:
@maelk done

dtantsur (approved):

PR #208, opened 2020-10-26
Merge RUN commands to reduce layers number

elfosardo:
When possible, we should merge RUN commands to reduce the number
of layers in the image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,10 +46,7 @@ RUN dnf install -y python3 python3-requests && \
     dnf clean all && \
     rm -rf /var/cache/{yum,dnf}/*
 
-RUN mkdir -p /tftpboot
-COPY --from=builder /tmp/ipxe/src/bin/undionly.kpxe /tftpboot
-COPY --from=builder /tmp/ipxe/src/bin-x86_64-efi/snponly.efi /tftpboot
-COPY --from=builder /tmp/ipxe/src/bin-x86_64-efi/ipxe.efi /tftpboot
+COPY --from=builder /tmp/ipxe/src/bin/undionly.kpxe /tmp/ipxe/src/bin-x86_64-efi/snponly.efi /tmp/ipxe/src/bin-x86_64-efi/ipxe.efi /tftpboot/
 
 COPY --from=builder /tmp/esp.img /tmp/uefi_esp.img
 


Dockerfile
```diff
@@ -44,9 +44,9 @@ RUN dnf install -y python3 python3-requests && \
     dnf upgrade -y && \
     dnf --setopt=install_weak_deps=False install -y $(cat /tmp/main-packages-list.txt) && \
     dnf clean all && \
-    rm -rf /var/cache/{yum,dnf}/*
+    rm -rf /var/cache/{yum,dnf}/* && \
+    mkdir -p /tftpboot
```
dhellmann, line 48: Maybe I'm missing something, but it doesn't seem like merging these steps into 1 command is really the best approach. Some of the layers aren't going to be different from build to build, so could be cached, right? Like if this mkdir command came first in its own RUN command, that layer of the image would almost never change.

Having all of the dnf commands together in one step does make sense. Is the curl command setting something up for dnf to be able to install the tripleo packages, or is that independent?

elfosardo: Please consider that when processing a RUN command the files updated in the container are not examined to determine if a cache hit exists. In that case just the command string itself is used to find a match.
In this specific case, having one RUN command is better than the two RUN commands we had before, I'm not sure having the mkdir first, in a different RUN command, would change much, but we can try.
The layer created using RUN with all the dnf commands won't be cached, so it doesn't make sense to decouple them.
Also consider that having dnf alone in a RUN line could cause caching issues.
The curl command enables the tripleo repos and it's necessary to install all the openstack packages, I would leave it there.

dhellmann: The unchanging layers wouldn't be cached during the build, but would be cached when pushing or pulling them, right? In this case the layer with the directory in it isn't very big, so it would hardly matter. But as a general approach it seems like lumping all changes into a single layer makes it impossible to take advantage of any caching at all.

elfosardo: The layer cache is just for the building process.
In general best practice says to merge all the RUN commands to avoid creating multiple layers, because these rarely invalidate the cache.
The RUN command will invalidate the cache only if the command string change, so this change is useful to reduce the number of layers, both during build as during push/pull.
For example, using pull, by default three layers of an image at a time are pulled by the docker daemon, so reducing the number of layers will reduce the pull time.

Dockerfile
```diff
@@ -44,9 +44,9 @@ RUN dnf install -y python3 python3-requests && \
...
 
-RUN mkdir -p /tftpboot
```
zaneb, line 49: IIUC this command isn't even needed, because COPY will create the directory for you.
A lot more layers could be saved by combining this and the next few lines:

```
COPY --from=builder /tmp/ipxe/src/bin/undionly.kpxe /tmp/ipxe/src/bin-x86_64-efi/snponly.efi /tmp/ipxe/src/bin-x86_64-efi/ipxe.efi /tftpboot/
```

elfosardo: correct, although it would be more prone to invalidating cache, so it's a deal beetween reducing build time or pull/push time.
I guess we probably prefer the latter?

zaneb: I don't understand how that would make it more prone to invalidating the cache tbh.

elfosardo: because you have more files in one COPY command, if just one of them changes, the entire layer is invalidated in the cache

zaneb: Right, but even with multiple COPY commands if the first one changes then the cache of the subsequent layers is discarded anyway. I don't see any a priori reason to think that the files in the later commands are more likely to change than those in the earlier commands, and even if there were then I'd expect it to have a negligible impact on build time.

zaneb (commented):
/approve

zaneb:
I don't have reviewer rights on this repo (*cough* #213), so all I can do is approve.

PR #209, opened 2020-10-26
Provide a way to apply upstream patches to the image

elfosardo:
This change allows users to apply patches to the images during build
time directly from gerrit using the new arg PATCH.
At the moment, it's limited to a single project per image; future
patches will allow applying patches from multiple projects, and
also from local directories.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,9 +33,10 @@ FROM docker.io/centos:centos8
 
 ENV PKGS_LIST=main-packages-list.txt
 ARG EXTRA_PKGS_LIST
+ARG PATCH_LIST
 
-COPY ${PKGS_LIST} ${EXTRA_PKGS_LIST} /tmp/
-COPY prepare-image.sh /bin/
+COPY ${PKGS_LIST} ${EXTRA_PKGS_LIST} ${PATCH_LIST} /tmp/
+COPY prepare-image.sh patch-image.sh /bin/
 
 RUN prepare-image.sh && \
   rm -f /bin/prepare-image.sh
--- a/README.md
+++ b/README.md
@@ -43,3 +43,16 @@ The ironic configuration can be overridden by various environment variables. The
 - OS_CONDUCTOR__INSPECT_TIMEOUT=1800 - timeout (seconds) for waiting for node inspection
 - OS_CONDUCTOR__CLEAN_CALLBACK_TIMEOUT=1800 - timeout (seconds) to wait for a callback from the ramdisk doing the cleaning
 - OS_PXE__BOOT_RETRY_TIMEOUT=1200 - timeout (seconds) to enable boot retries.
+
+Applying Patches to the image
+-----------------------------
+
+When building the image, it is possible to specify a patch of one or more
+upstream projects to apply to the image, passing a file with the patch list
+using the PATCH_LIST build argument.
+At the moment, only projects hosted in opendev.org are supported.
+
+Each line of the file is in the form "project_dir refspec" where:
+- project is the last part of the project url including the org, for example openstack/ironic
+- refspec is the gerrit refspec of the patch we want to test, for example refs/changes/67/759567/1
+
--- a/patch-image.sh
+++ b/patch-image.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/bash
+set -ex
+
+patch_file="/tmp/${PATCH_LIST}"
+
+while IFS= read -r line
+do
+    # each line is in the form "project_dir refsspec" where:
+    # - project is the last part of the project url including the org,
+    # for example openstack/ironic
+    # - refspec is the gerrit refspec of the patch we want to test,
+    # for example refs/changes/67/759567/1
+    PROJECT=$(echo $line | cut -d " " -f1)
+    PROJ_NAME=$(echo $PROJECT | cut -d "/" -f2)
+    PROJ_URL="https://opendev.org/$PROJECT"
+    REFSPEC=$(echo $line | cut -d " " -f2)
+
+    cd /tmp
+    git clone "$PROJ_URL"
+    cd "$PROJ_NAME"
+    git fetch "$PROJ_URL" "$REFSPEC"
+    git checkout FETCH_HEAD
+
+    SKIP_GENERATE_AUTHORS=1 SKIP_WRITE_GIT_CHANGELOG=1 python3 setup.py sdist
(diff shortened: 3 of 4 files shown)


patch-image.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/bash
+set -ex
+
+# PATCH argument is in the form "project refs" where refs is a list of
+# comma-separated gerrit refs, fo example refs/changes/67/759567/1
```
dtantsur, line 5: This is a good start, but ideally we need to support more than one project (for depends-on handling).

dtantsur (approved):

patch-image.sh
elfosardo: yep, I mentioned that in the commit message

patch-image.sh
```diff
@@ -0,0 +1,21 @@
...
+PROJ_NAME=$(echo $1 | cut -d " " -f1)
+REFSPECS=$(echo $1 | cut -d " " -f2)
+
+cd /tmp
+git clone "https://opendev.org/openstack/$PROJ_NAME"
+cd "$PROJ_NAME"
+for REFSPEC in $REFSPECS; do
```
zaneb, line 12: The comment says it's a comma-separated list, but I don't see us splitting on commas anywhere?

patch-image.sh
```diff
@@ -0,0 +1,21 @@
...
+    git fetch "https://review.opendev.org/openstack/$PROJ_NAME" "$REFSPEC"
+    git -c user.name=Nobody -c user.email=nobody@example.com cherry-pick FETCH_HEAD
```
zaneb, line 14: This is cherry-picking patches on top of master, but it's not obvious that master is always what you want to be based on. (e.g. Zuul does optimistic merging, so if this were used in a Zuul gate test it wouldn't necessarily test the right thing.) It would seem simpler to just pass a single refspec and check that out. If you have multiple changes in a series, you can just point to the last one.

patch-image.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/bash
+set -ex
+
+# PATCH argument is in the form "project refs" where refs is a list of
+# comma-separated gerrit refs, fo example refs/changes/67/759567/1
+PROJ_NAME=$(echo $1 | cut -d " " -f1)
+REFSPECS=$(echo $1 | cut -d " " -f2)
+
+cd /tmp
+git clone "https://opendev.org/openstack/$PROJ_NAME"
```
zaneb, line 10: I wonder if we should allow passing the full clone URL, so that not only upstream OpenStack patches can be tested.

patch-image.sh
dtantsur: For a Zuul CI job we'll need to pass a link to a local checkout, I wonder if we could make it possible.

elfosardo: @zaneb so just using checkout instead of cherry-pick? that could work better for series, I agree. I tested this only on single patches and honestly didn't think about that.
@dtantsur we just need to pass the last refspec in the series, why a link ?

patch-image.sh
```diff
@@ -0,0 +1,30 @@
...
+    # refs/changes/67/759567/1,refs/changes/67/787632/11
+    PROJECT=$(echo $line | cut -d " " -f1)
+    PROJ_NAME=$(echo $PROJECT | cut -d "/" -f2)
+    PROJ_URL="https://opendev.org/$PROJECT"
+    REFSPECS=$(echo $line | cut -d " " -f2)
+
+    cd /tmp
+    git clone "$PROJ_URL"
+    cd "$PROJ_NAME"
+    for REFSPEC in "${REFSPECS//,/ }"; do
```
zaneb, line 21: It doesn't make sense to checkout in a loop. We should just provide the refspec of the last change instead of a list.

prepare-image.sh
```diff
@@ -8,3 +8,7 @@ dnf upgrade -y
 dnf --setopt=install_weak_deps=False install -y $(cat /tmp/main-packages-list.txt)
 dnf clean all
 rm -rf /var/cache/{yum,dnf}/*
+if [ -s "/tmp/${PATCH_LIST}" ]; then
```
derekhiggins, line 11: This test returns true for directories
```
+ '[' -s /tmp/ ']'                                              
+ /bin/patch-image.sh     
+ patch_file=/tmp/
+ IFS=
+ read -r line
/bin/patch-image.sh: line 6: read: read error: 0: Is a directory

```

README.md
```diff
@@ -43,3 +43,15 @@ The ironic configuration can be overridden by various environment variables. The
...
 - OS_CONDUCTOR__CLEAN_CALLBACK_TIMEOUT=1800 - timeout (seconds) to wait for a callback from the ramdisk doing the cleaning
 - OS_PXE__BOOT_RETRY_TIMEOUT=1200 - timeout (seconds) to enable boot retries.
+
+Applying Patches to the image
+-----------------------------
+
+When building the image, it is possible to specify a patch of one or more
+upstream projects to apply to the image, passing a file with the patch list
+using the PATCH_LIST build argument.
+
```
derekhiggins, line 53: This might be obvious but should we say that the project needs to be hosted on opendev.org's gerrit?

prepare-image.sh
elfosardo: yes, I forgot to correct this one, thanks!

README.md
elfosardo: It's true at the moment, I will add a note on that

Dockerfile
```diff
@@ -35,7 +35,7 @@ ENV PKGS_LIST=main-packages-list.txt
 ARG EXTRA_PKGS_LIST
 
 COPY ${PKGS_LIST} ${EXTRA_PKGS_LIST} /tmp/
```
derekhiggins, line 37: A previous version of this PR included "${PATCH_LIST}" in this line, but its now gone, the weird thing is that it still works, this brought me down a bit of a rabbit hole. As far as I can see leaving "EXTRA_PKGS_LIST" blank results in the copy command treating it as a "*", so the entire contents of the local directory get copied into /tmp (including my PATCH_LIST file). So it works but isn't what we would want as the /tmp directory is getting cluttered in the image.

elfosardo: @derekhiggins we saw this behavior in other situations and this is actually a bug in podman, although I can't find the exact reference now
the ${PATCH_LIST} is actually needed here, thanks for catching this

PR #211, opened 2020-10-28
Use [] around IPv6 address for conductor host config

### LINKED ISSUE #210: Basic auth on json-rpc broken with IPv6
There is a [bug in Ironic](https://storyboard.openstack.org/#!/story/2008288) with json-rpc when specifying an IPv6 address for the `[DEFAULT].host` config option for ironic-conductor.

Since #201 we explicitly set this option to `$IRONIC_IP` except when the json-rpc auth type is `noauth` (in which case it's forced to be `localhost`).

Until the bug is fixed, basic auth on json-rpc is unusable with IPv6.

zaneb:
Work around https://storyboard.openstack.org/#!/story/2008288 by setting
the DEFAULT.host config option to the host portion of a URL, meaning
that IPv6 addresses are surrounded by [], rather than just the raw IP
address. This is how it is ultimately used inside Ironic.

This solution works both with and without the fix for the issue in
Ironic (https://review.opendev.org/759877).

Fixes #210

### DIFF
--- a/ironic.conf.j2
+++ b/ironic.conf.j2
@@ -33,7 +33,7 @@ my_ip = {{ env.IRONIC_IP }}
 # If we run both API and conductor in the same pod, use localhost
 host = localhost
 {% else %}
-host = {{ env.IRONIC_IP }}
+host = {{ env.IRONIC_URL_HOST }}
 {% endif %}
 
 isolinux_bin = /usr/share/syslinux/isolinux.bin


dtantsur (approved):
WFM, although I'd prefer it eventually reverted.

derekhiggins:
/lgtm

> WFM, although I'd prefer it eventually reverted.
same

PR #212, opened 2020-10-30
Secure boot compatible ESP image

dtantsur:
Avoid rebuilding GRUB2 since it invalidates its signature. Ironic
is able to build correct ISO images if you provide it with
a distribution specific EFI config path.

See https://review.opendev.org/760434


maelk:
Is this change compatible with any deployed image ? If our target image would be Ubuntu, Suse or anything else, would it keep working properly ?

dtantsur:
> Is this change compatible with any deployed image ?

This change is unrelated to the deployed image, the grub and its configuration are only used during provisioning.

elfosardo (approved):

PR #214, opened 2020-11-03
Limit the default number of API workers to 4

dtantsur:
12 is a lot of workers. With up to 100MiB of RAM per each, ironic-api
processes may occupy more than 1GiB of RAM in such configuration.

Since even CERN uses 2 workers in their configuration, we should
be fine with no more than 4. A similar change is made upstream:
https://review.opendev.org/761185

### DIFF
--- a/configure-ironic.sh
+++ b/configure-ironic.sh
@@ -24,8 +24,11 @@ fi
 
 export HTTP_PORT=${HTTP_PORT:-"80"}
 export MARIADB_PASSWORD=${MARIADB_PASSWORD:-"change_me"}
-export NUMPROC=$(cat /proc/cpuinfo  | grep "^processor" | wc -l)
-export NUMWORKERS=$(( NUMPROC < 12 ? NUMPROC : 12 ))
+# TODO(dtantsur): remove the explicit default once we get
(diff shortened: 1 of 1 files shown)


configure-ironic.sh
```diff
@@ -24,8 +24,11 @@ fi
 
 export HTTP_PORT=${HTTP_PORT:-"80"}
 export MARIADB_PASSWORD=${MARIADB_PASSWORD:-"change_me"}
-export NUMPROC=$(cat /proc/cpuinfo  | grep "^processor" | wc -l)
-export NUMWORKERS=$(( NUMPROC < 12 ? NUMPROC : 12 ))
+# TODO(dtantsur): remove the explicit default once we get
```
dhellmann, line 27: Is there any reason for us not to just pick a number here instead of making it dynamic at all?

dtantsur: How beefy are the CI nodes? I guess if they have 1-2 cores, we don't want to launch 4 API workers.

dhellmann: Ah, that's fair I guess.

maelk:
lgtm

PR #216, opened 2020-11-05
Use collector config param in ironic.conf expected by IPA

derekhiggins:
This had been fixed already but the wrong name came back,
presumably missed on a rebase.


derekhiggins:
Now fixed here https://github.com/metal3-io/ironic-image/pull/232

PR #219, opened 2020-11-10
Take advantage of multi files COPY to simplify Dockerfile

elfosardo:
Move scripts under scripts dir and config files under config dir,
then copy everything at the same time when possible.
Also no need to specify ./ with COPY, the default path is always
inside the context of the build.


elfosardo:
@hardys @maelk any chance I can get  review on this ?

Dockerfile
```diff
@@ -33,7 +33,7 @@ FROM docker.io/centos:centos8
 
 ARG PKGS_LIST=main-packages-list.txt
 
-COPY ${PKGS_LIST} /tmp/main-packages-list.txt
+COPY ${PKGS_LIST} /tmp/
```
zaneb, line 36: If the user supplies a different arg for PKGS_LIST we still want it to end up at /tmp/main-packages-list.txt, so I don't think this change is correct.

zaneb (commented):
/lgtm cancel

Dockerfile
elfosardo: whoops this was not supposed to be here!

elfosardo:
@zaneb hi! Can you please have another look? Thanks!

elfosardo:
@zaneb @hardys @maelk can you please give this another look? Thanks!

elfosardo:
@zaneb @hardys @maelk can you please give this another look? Thanks!

Dockerfile
```diff
@@ -45,29 +45,17 @@ COPY --from=builder /tmp/ipxe/src/bin/undionly.kpxe /tmp/ipxe/src/bin-x86_64-efi
...
-# TODO(dtantsur): remove this script when we stop supporting running both
-# API and conductor processes via one entry point.
-COPY ./runironic.sh /bin/runironic
-
-COPY ./dnsmasq.conf.j2 /etc/dnsmasq.conf.j2
-COPY ./inspector.ipxe.j2 /tmp/inspector.ipxe.j2
-COPY ./dualboot.ipxe /tmp/dualboot.ipxe
+# TODO(dtantsur): remove scripts/runironic script when we stop supporting
+# running both API and conductor processes via one entry point.
+COPY scripts/ /bin/
```
zaneb, line 52: Should this be `scripts/*`? This is correct syntax for rsync, but I can't find any indication in the [Dockerfile docs](https://docs.docker.com/engine/reference/builder/#copy) that the COPY directives treats a source with a trailing slash differently, rsync-style.

elfosardo: from [COPY Docker docs](https://docs.docker.com/engine/reference/builder/#copy):
"If <src> is a directory, the entire contents of the directory are copied, including filesystem metadata."

zaneb: And right after that:

>     Note
> 
>     The directory itself is not copied, just its contents.

zaneb: You know I read that like 5 times, and each time my brain transposed the position of the `not` :facepalm:

dtantsur (approved):

PR #217, opened 2020-11-07
Add new entry point script to dump ramdisk host provisioning logs

akiselev1:
This is the first step in series of PRs implementing this design proposal: https://github.com/metal3-io/metal3-docs/blob/master/design/ironic-debuggability-improvement.md

PR #220, opened 2020-11-10
Add inotify-tools to the image

akiselev1:
We need inotifywait as the most efficient way to watch for new log files in https://github.com/metal3-io/metal3-docs/blob/master/design/ironic-debuggability-improvement.md


dhellmann:
Could this be part of #217?

PR #223, opened 2020-11-17
Don't use base image from docker.io

zaneb:
Docker are throttling requests, so get the CentOS base image from the
CentOS registry.


zaneb:
It appears this is not maintained so probably not something we should rely on being up to date.
There is a [request](https://pagure.io/centos-infra/issue/77) open for CentOS to publish their images to quay.io.
/close

PR #225, opened 2020-11-18
Move upgrade and install logic to a script

elfosardo:
In the effort of cleaning the Dockerfile, we move the upgrade and
install logic to a script.


elfosardo:
@bfournie @dtantsur @derekhiggins please review this when you have the chance

PR #237, opened 2021-01-04
Allow specify package versions

elfosardo:
With this change we can specify package versions to be installed
directly in the packages list (main or extra).

### DIFF
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -5,10 +5,10 @@ set -euxo pipefail
 dnf install -y python3 python3-requests
 curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current-tripleo
 dnf upgrade -y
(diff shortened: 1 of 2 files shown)


iurygregory (approved):

prepare-image.sh
```diff
@@ -5,10 +5,10 @@ set -euxo pipefail
 dnf install -y python3 python3-requests
 curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current-tripleo
 dnf upgrade -y
-dnf --setopt=install_weak_deps=False install -y $(cat /tmp/${PKGS_LIST})
+xargs -rd'\n' dnf --setopt=install_weak_deps=False install -y < /tmp/${PKGS_LIST}
```
derekhiggins, line 8: Can we add a "-t" to the xargs command so that the dnf command that ran is logged (and below)

elfosardo: yes, we can! good idea :)

dtantsur (approved):

PR #239, opened 2021-01-28
Supply a default for COPY args

derekhiggins:
Supplying a blank arg to COPY results in all of
the files in the source directory being copied into
the image into /tmp for docker/podman build (or "/"
if using imagebuilder). Give it a default of a file
that exists to avoid this happening.


iurygregory (approved):

elfosardo (approved):
thanks Derek, tested and it works well

PR #240, opened 2021-02-01
Make registry configurable for image build

Xenwar:
This PR parameterizes the docker registry for building images.


Xenwar:
replaced by an alternative solution.

PR #242, opened 2021-03-03
Configure conductor heartbeat timeout

kashifest:
This PR enables ironic conductor's hearbeat timeout to be configurable through env/configmap. The goal is to make the IPA hearbeat checkup interval configurable as described here https://docs.openstack.org/ironic-python-agent/mitaka/#heartbeat


kashifest:
Apparently it comes out that it is already doable by exporting the correct variables. So this PR is not needed. Closing it.
/close

PR #246, opened 2021-03-19
Include vlan interfaces in introspection report

bfournie:
Add the inspector kernel param to include vlan interfaces and
associated IP addresses in the introspection report. This feature
was added to IPA in
https://review.opendev.org/c/openstack/ironic-python-agent/+/760570.

By default all VLAN interfaces are included in the report using LLDP
info from the switch to determine the VLANs. An override is provided
to define a particular VLAN interface if LLDP is not enabled.


dtantsur:
/test-integration
/lgtm

Travis failed because of pull limits, not sure it's worth rechecking.

maelk:
I have re-triggered travis

PR #247, opened 2021-03-25
Restart Ironic when certificate is updated

namnx228:
When the TLS certificate is updated, Ironic will reload (or restart) to get the new certificate.

### DIFF
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -2,7 +2,8 @@
 
 set -euxo pipefail
 
-dnf install -y python3 python3-requests
+dnf install -y python3 python3-requests epel-release 'dnf-command(config-manager)'
+dnf config-manager --set-disabled epel
 curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current-tripleo --no-stream
 dnf upgrade -y
 xargs -rtd'\n' dnf --setopt=install_weak_deps=False install -y < /tmp/${PKGS_LIST}
@@ -11,7 +12,7 @@ if [[ ! -z ${EXTRA_PKGS_LIST:-} ]]; then
         xargs -rtd'\n' dnf --setopt=install_weak_deps=False install -y < /tmp/${EXTRA_PKGS_LIST}
     fi
 fi
-
+dnf install -y --enablerepo=epel inotify-tools
 dnf clean all
 rm -rf /var/cache/{yum,dnf}/*
 if [[ ! -z ${PATCH_LIST:-} ]]; then
--- a/scripts/configure-ironic.sh
+++ b/scripts/configure-ironic.sh
@@ -8,6 +8,7 @@ export IRONIC_INSECURE=${IRONIC_INSECURE:-false}
 export IRONIC_INSPECTOR_CERT_FILE=/certs/ironic-inspector/tls.crt
 export IRONIC_INSPECTOR_CACERT_FILE=/certs/ca/ironic-inspector/tls.crt
 export IRONIC_INSPECTOR_INSECURE=${IRONIC_INSPECTOR_INSECURE:-$IRONIC_INSECURE}
+export RESTART_CONTAINER_CERTIFICATE_UPDATED=${RESTART_CONTAINER_CERTIFICATE_UPDATED:-"false"}
 
 # Define the VLAN interfaces to be included in introspection report, e.g.
 #   all - all VLANs on all interfaces using LLDP information
--- a/scripts/runhttpd
+++ b/scripts/runhttpd
@@ -10,6 +10,7 @@ INSPECTOR_RESULT_HTTPD_CONFIG=/etc/httpd/conf.d/ironic-inspector.conf
 export IRONIC_INSPECTOR_CERT_FILE=/certs/ironic-inspector/tls.crt
 export IRONIC_INSPECTOR_KEY_FILE=/certs/ironic-inspector/tls.key
 export INSPECTOR_REVERSE_PROXY_SETUP=${INSPECTOR_REVERSE_PROXY_SETUP:-"false"}
+export RESTART_CONTAINER_CERTIFICATE_UPDATED=${RESTART_CONTAINER_CERTIFICATE_UPDATED:-"false"}
 
 # Whether to enable fast_track provisioning or not
 IRONIC_FAST_TRACK=${IRONIC_FAST_TRACK:-true}
@@ -73,5 +74,11 @@ sed -i -e 's|\(^[[:space:]]*\)\(DocumentRoot\)\(.*\)|\1\2 "/shared/html"|' \
 # Log to std out/err
 sed -i -e 's%^ \+CustomLog.*%    CustomLog /dev/stderr combined%g' /etc/httpd/conf/httpd.conf
 sed -i -e 's%^ErrorLog.*%ErrorLog /dev/stderr%g' /etc/httpd/conf/httpd.conf
+
+if [[ "$IRONIC_INSPECTOR_TLS_SETUP" == "true"  && "${RESTART_CONTAINER_CERTIFICATE_UPDATED}" == "true" ]]; then
+    inotifywait -m -e delete_self "${IRONIC_INSPECTOR_CERT_FILE}" | while read file event; do
+      kill -WINCH $(pgrep httpd)
+    done &
+fi
 
 exec /usr/sbin/httpd -DFOREGROUND
--- a/scripts/runironic-api
+++ b/scripts/runironic-api
@@ -19,5 +19,12 @@ done
 
 python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ))' < /etc/httpd-ironic-api.conf.j2 > /etc/httpd/conf.d/ironic.conf
 sed -i "/Listen 80/c\#Listen 80" /etc/httpd/conf/httpd.conf
+
+if [[ "$IRONIC_TLS_SETUP" == "true"  && "${RESTART_CONTAINER_CERTIFICATE_UPDATED}" == "true" ]]; then
+    inotifywait -m -e delete_self "${IRONIC_CERT_FILE}" | while read file event; do
+      kill -WINCH $(pgrep httpd)
+    done &
+fi
+
 exec /usr/sbin/httpd -DFOREGROUND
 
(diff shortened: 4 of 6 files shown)


ironic-scripts/runironic-conductor
```diff
@@ -13,4 +13,12 @@ until ironic-dbsync --config-file /etc/ironic/ironic.conf upgrade; do
   echo "WARNING: ironic-dbsync failed, retrying"
   sleep 1
 done
+
+if [ "$IRONIC_TLS_SETUP" == "true" ]; then
+    inotifywait -m -e delete_self "${IRONIC_CERT_FILE}" | while read file event; do
+     kill $(pgrep ironic)
+     /usr/bin/ironic-conductor &
```
dtantsur, line 20: No starting anything in the background please. The previous line will likely cause the container to stop anyway.

dtantsur (changes requested):
I don't think this approach will fly, you probably need whatever party orchestrates the containers to restart them.

namnx228:
> I don't think this approach will fly, you probably need whatever party orchestrates the containers to restart them.

Yes, I agree that this approach is not an optimal way. However, at the moment, we are needing this feature soon, so a straightforward solution is good to go, IMO. But I think we will need to design a more K8s native solution in the future.

maelk:
Considering that cert-manager can rotate the certificates at any time, without notifications (it just updates the secret), I don't see a better solution than Ironic being able to detect that the certificate changed and reload it, or since it does not support that for now, having a helper script that will cause a restart of Ironic.

s3rj1k:
inotify would not work on NFS for example

prepare-image.sh
```diff
@@ -2,7 +2,7 @@
 
 set -euxo pipefail
 
-dnf install -y python3 python3-requests
+dnf install -y python3 python3-requests epel-release
```
dtantsur, line 5: Please make enabling epel conditional on RESTART_CONTAINER_CERTIFICATE_UPDATED, not everyone wants to have it (and it may even conflict with tripleo repos).

ironic-scripts/configure-ironic.sh
```diff
@@ -8,6 +8,7 @@ export IRONIC_INSECURE=${IRONIC_INSECURE:-false}
 export IRONIC_INSPECTOR_CERT_FILE=/certs/ironic-inspector/tls.crt
 export IRONIC_INSPECTOR_CACERT_FILE=/certs/ca/ironic-inspector/tls.crt
 export IRONIC_INSPECTOR_INSECURE=${IRONIC_INSPECTOR_INSECURE:-$IRONIC_INSECURE}
+export RESTART_CONTAINER_CERTIFICATE_UPDATED=${RESTART_CONTAINER_CERTIFICATE_UPDATED:-"true"}
```
dtantsur, line 11: Should be false, since this is not the traditional way of handling configuration.

dtantsur (changes requested):

ironic-scripts/runironic-api
```diff
@@ -19,5 +19,12 @@ done
 
 python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ))' < /etc/httpd-ironic-api.conf.j2 > /etc/httpd/conf.d/ironic.conf
 sed -i "/Listen 80/c\#Listen 80" /etc/httpd/conf/httpd.conf
+
+if [ "$IRONIC_TLS_SETUP" == "true"  && "${RESTART_CONTAINER_CERTIFICATE_UPDATED}" == "true" ]; then
+    inotifywait -m -e delete_self "${IRONIC_CERT_FILE}" | while read file event; do
+      kill $(pgrep httpd)
+    done &
```
zaneb, line 26: How does this interact with the fact that we are calling `exec` below, so the subshell spawned here becomes a child of the httpd process?

ironic-scripts/runironic-api
```diff
@@ -19,5 +19,12 @@ done
 
 python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ))' < /etc/httpd-ironic-api.conf.j2 > /etc/httpd/conf.d/ironic.conf
 sed -i "/Listen 80/c\#Listen 80" /etc/httpd/conf/httpd.conf
+
+if [ "$IRONIC_TLS_SETUP" == "true"  && "${RESTART_CONTAINER_CERTIFICATE_UPDATED}" == "true" ]; then
+    inotifywait -m -e delete_self "${IRONIC_CERT_FILE}" | while read file event; do
+      kill $(pgrep httpd)
```
zaneb, line 25: This should use `kill -HUP` (for a graceful restart), and I suspect the same is true of the others.

ironic-scripts/configure-ironic.sh
```diff
@@ -8,6 +8,7 @@ export IRONIC_INSECURE=${IRONIC_INSECURE:-false}
 export IRONIC_INSPECTOR_CERT_FILE=/certs/ironic-inspector/tls.crt
 export IRONIC_INSPECTOR_CACERT_FILE=/certs/ca/ironic-inspector/tls.crt
 export IRONIC_INSPECTOR_INSECURE=${IRONIC_INSPECTOR_INSECURE:-$IRONIC_INSECURE}
+export RESTART_CONTAINER_CERTIFICATE_UPDATED=${RESTART_CONTAINER_CERTIFICATE_UPDATED:-"false"}
```
zaneb, line 11: It seems to me that rather than have an environment variable we could just check whether inotifywait exists, and assume that if the functionality was built into the container that we want to use it.

ironic-scripts/runmariadb
```diff
@@ -18,6 +19,13 @@ fi
 
 ln -sf /proc/self/fd/1 /var/log/mariadb/mariadb.log
 
+# Restart mysqld when the certificate is updated
+if [ -f "$MARIADB_CERT_FILE" && "${RESTART_CONTAINER_CERTIFICATE_UPDATED}" == "true" ]; then
+    inotifywait -m -e delete_self "${MARIADB_CERT_FILE}" | while read file event; do
+       kill $(pgrep mysqld)
```
zaneb, line 25: `mysqld_safe` automatically restarts if mysqld exits, so this is consistent with the commit message but inconsistent with the implementation of the other services (where we exit the container).

prepare-image.sh
namnx228: @dtantsur The environment variable RESTART_CONTAINER_CERTIFICATE_UPDATED can only be set when we run the container, but the package needs to be installed when the image is built. Do you have any ideas how should we handle it?

dtantsur: You can provide options in Dockerfiles as well

namnx228: When we use the image, for example in BMO, we only pull the images from quay.io. We don't rebuild the image using the Dockerfile. In this case, we cannot pick the option that is provided in Dockerfiles.

ironic-scripts/runironic-api
namnx228: Yes, I have checked. The subshell becomes a child of the httpd process

ironic-scripts/runironic-api
namnx228: I tried to send SIGHUP signal to the ironic-conductor process and I can see that it reloads the config, but it does not re-read the TLS certificate file. So the only way is to restart the container.

ironic-scripts/configure-ironic.sh
namnx228: I think the inotify-tools will always be baked into the image, but it is the user's choice to use it or not. That is why we need this environment variable.

ironic-scripts/runironic-api
s3rj1k: just a thought, how about ditching apache in favor of nginx? 
(nginx can dynamically load SSL certs)

namnx228: Btw, thanks for the suggestion to use `kill -WINCH`. It is helpful.

namnx228: @s3rj1k Apache server can also do the same, actually. The reason why I restart the httpd is to make it consistent with the other containers

ironic-scripts/runmariadb
namnx228: Thanks! Fixed!

prepare-image.sh
dtantsur: Well, you have to rebuild it or publish an alternative image. Or you can only install inotify from EPEL without pulling in anything else.
```
09:45 <dtantsur> hi folks! I remember long ago it was not recommended to use EPEL with OpenStack repositories. What's the situation now?
10:26 <apevec> dtantsur, the same
```

prepare-image.sh
```diff
@@ -2,9 +2,11 @@
 
 set -euxo pipefail
 
-dnf install -y python3 python3-requests
-curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current-tripleo
+dnf install -y python3 python3-requests epel-release 
+
+curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current-tripleo --no-stream
```
dtantsur, line 7: The `--no-stream` change is unrelated and should be posted separately.

prepare-image.sh
```diff
@@ -2,9 +2,11 @@
 
 set -euxo pipefail
 
-dnf install -y python3 python3-requests
-curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current-tripleo
+dnf install -y python3 python3-requests epel-release
```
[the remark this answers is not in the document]
namnx228: Ok, I have updated that only the inotify-tools package is installed from EPEL

namnx228:
This PR needs to wait for https://github.com/metal3-io/ironic-image/pull/259 to go in before it can pass the Travis CI test

prepare-image.sh
```diff
@@ -11,6 +11,7 @@ if [[ ! -z ${EXTRA_PKGS_LIST:-} ]]; then
         xargs -rtd'\n' dnf --setopt=install_weak_deps=False install -y < /tmp/${EXTRA_PKGS_LIST}
     fi
 fi
+dnf install -y https://download-ib01.fedoraproject.org/pub/epel/8/Everything/x86_64/Packages/i/inotify-tools-3.14-19.el8.x86_64.rpm
```
dtantsur, line 14: This will break once this version is deleted. What you need is to install epel-release but disable the added repo via dnf (`dnf config-manager --set-disabled epel` or something like that). Then you can use `--enablerepo=epel` when installing inotify-tools specifically.

dtantsur (approved):

maelk (commented):
/lgtm

PR #248, opened 2021-03-26
Remove net-tools installation

elfosardo:
Now that https://github.com/metal3-io/baremetal-operator/pull/728 has
merged, we can remove net-tools installation command.

### DIFF
--- a/ironic-scripts/runironic-api
+++ b/ironic-scripts/runironic-api
@@ -14,10 +14,7 @@ while true ; do
   sleep 5
 done
 
-
-# TODO: Delete the line of code below and uncomment the next line after the PR https://github.com/metal3-io/baremetal-operator/pull/728 goes in
-[ ! "$(netstat -l | grep ${HTTP_PORT})" ] && . /bin/configure-httpd-ipa.sh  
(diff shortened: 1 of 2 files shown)


dtantsur (changes requested):
net-tools were installed for using netstat in scripts/runironic-api, that logic should also be removed

ironic-scripts/runironic-api
```diff
@@ -14,9 +14,6 @@ while true ; do
   sleep 5
 done
 
-
-# TODO: Delete the line of code below and uncomment the next line after the PR https://github.com/metal3-io/baremetal-operator/pull/728 goes in
-[ ! "$(netstat -l | grep ${HTTP_PORT})" ] && . /bin/configure-httpd-ipa.sh  
 # . /bin/configure-httpd-ipa.sh
```
dtantsur, line 17: I think the comment suggested uncommenting this line.

elfosardo: yes, I forgot to remove the comment......

dtantsur (approved):
/test-integration

PR #251, opened 2021-04-09
fix: ensure dualboot.ipxe read permission for others

kunyi731:
In order for ipxe to read this dualboot.ipxe, the file has to be readable.
This can be a problem if the image was built on systems with a strict umask
such as 0027. Ensure the file is readable using chmod in Dockerfile.

Otherwise one might see errors like the following:

```
Filename: http://21.0.12.28:6180/dualboot.ipxe
http://21.0.12.28:6180/dualboot.ipxe... Operation not permitted (http://ipxe.org/410c618f)
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,6 +54,7 @@ COPY ironic-config/ironic.conf.j2 /etc/ironic/
 COPY ironic-scripts/ /bin/
 COPY ironic-config/dnsmasq.conf.j2 /etc/
 COPY ironic-config/inspector.ipxe.j2 ironic-config/dualboot.ipxe /tmp/
+RUN chmod 644 /tmp/dualboot.ipxe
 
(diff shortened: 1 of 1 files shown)


dtantsur (changes requested):
I think it should rather be done in https://github.com/metal3-io/ironic-image/blob/master/ironic-scripts/runhttpd#L30 (same for inspector.ipxe).

kunyi731:
> I think it should rather be done in https://github.com/metal3-io/ironic-image/blob/master/ironic-scripts/runhttpd#L30 (same for inspector.ipxe).

Thanks, I can move the change there. BTW, would it be better to use [install](https://man7.org/linux/man-pages/man1/install.1.html) rather than 'cp'/'chmod'?

kunyi731:
Created new PR: https://github.com/metal3-io/ironic-image/pull/256
Thanks!

PR #252, opened 2021-04-12
Merge ironic and ironic-inspector repositories, 2 Dockerfiles

elfosardo:
This change moves Dockerfile, configuration and scripts from the
ironic-inspector container image repository to the ironic one.

This approach provides two separate Dockerfiles in the same repository,
one for ironic and one for ironic-inspector, that share part of the
logic needed to build the images.
The scripts and configurations are kept separated for the time being as
some of them conflict with each other (e.g. runhttpd).

The original Dockerfile is kept until we make the necessary changes to
the CI.

This change is mutually exclusive with #253

### DIFF
--- a/Dockerfile.ironic
+++ b/Dockerfile.ironic
@@ -1,0 +1,62 @@
+## Build iPXE w/ IPv6 Support
+## Note: we are pinning to a specific commit for reproducible builds.
+## Updated as needed.
+FROM docker.io/centos:centos8 AS ironic-builder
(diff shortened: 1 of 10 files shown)


dtantsur (changes requested):
I'd prefer we don't go down this patch. This defeats one of the upsides of merging images: sharing most of the Python base and the httpd service.

elfosardo:
consensus is on single dockerfile, so closing this

PR #258, opened 2021-04-28
Set webserver_verify_ca to bool or certificate path

bfournie:
Set the ironic conf setting `webserver_verify_ca` to False
if IRONIC_INSECURE is True (to disable TLS validation) or set
it to path to CA_BUNDLE file if new IRONIC_CA_PATH env
variable is set.

By default it will be set to True.

### DIFF
--- a/ironic-config/ironic.conf.j2
+++ b/ironic-config/ironic.conf.j2
@@ -30,6 +30,13 @@ host = localhost
 host = {{ env.IRONIC_URL_HOST }}
 {% endif %}
 
+# If a path to a certificate is defined, use that first for webserver
+{% if env.WEBSERVER_CACERT_FILE %}
+webserver_verify_ca =  {{ env.WEBSERVER_CACERT_FILE }}
+{% elif env.IRONIC_INSECURE == "true" %}
+webserver_verify_ca = false
+{% endif %}
+
 isolinux_bin = /usr/share/syslinux/isolinux.bin
 
 # NOTE(dtantsur): this path is specific to the GRUB image that is built into


dtantsur (changes requested):
`IRONIC_CA_PATH` is a CA that ironic services use, it does not necessarily match the CA that the images use (and it's not the case in OpenShift). We need a new option, and we need it to have a higher priority than IRONIC_INSECURE (which is always True for OpenShift).

bfournie:
> `IRONIC_CA_PATH` is a CA that ironic services use, it does not necessarily match the CA that the images use (and it's not the case in OpenShift). We need a new option, and we need it to have a higher priority than IRONIC_INSECURE (which is always True for OpenShift).

IRONIC_CA_PATH is a new option, unfortunately the name is bit too similar to the existing one - IRONIC_CACERT_FILE. I'll change the name to make it clearer and move the check before INSECURE

ironic-config/ironic.conf.j2
```diff
@@ -30,6 +30,15 @@ host = localhost
 host = {{ env.IRONIC_URL_HOST }}
 {% endif %}
 
+# If a path to a certificate is defined, use that first for webserver
+{% if env.WEBSERVER_CACERT_FILE %}
+webserver_verify_ca =  {{ env.WEBSERVER_CACERT_FILE }}
+{% elif env.IRONIC_INSECURE == "true" %}
+webserver_verify_ca = false
+{% else %}
+webserver_verify_ca = true
```
dtantsur, line 39: nit: no need for this branch, true is the default

dtantsur (approved):
/approve

ironic-config/ironic.conf.j2
bfournie: Done

dtantsur:
/test-integration

I'm not sure how to re-run travis, maybe you can rebase?

dtantsur (approved):

elfosardo:
@bfournie as Dmitry said you'll have to rebase this :/

PR #259, opened 2021-04-28
Fix the issue that the OS is not a Stream version

namnx228:
This PR aims to solve this error: `__main__.InvalidArguments: --stream provided, but OS is not the Stream version. Please ensure the host is Stream.`


dtantsur (approved):

dtantsur:
/test-integration
/approve
/lgtm

Maybe we should consider switching to Stream, but right now it's blocking the image builds.

elfosardo (approved):

scripts/runmariadb
```diff
@@ -17,6 +17,7 @@ if [ ! -f "$MARIADB_CERT_FILE" ] && [ -f "$MARIADB_KEY_FILE" ] ; then
 fi
 
 ln -sf /proc/self/fd/1 /var/log/mariadb/mariadb.log
+rm -f /usr/bin/mysqld_safe_helper # Avoid this script being used which results in a permission denied error.
```
dtantsur, line 20: This doesn't look like an appropriate fix. In any case, we need a much more detailed explanation here than "permission denied error".

dtantsur: I think what we need is correct permissions ` /var/log/mariadb/`. You're simply removing the mysql executable.

scripts/runmariadb
```diff
@@ -17,6 +17,7 @@ if [ ! -f "$MARIADB_CERT_FILE" ] && [ -f "$MARIADB_KEY_FILE" ] ; then
 fi
 
 ln -sf /proc/self/fd/1 /var/log/mariadb/mariadb.log
```
dtantsur, line 19: Can we maybe use some normal approach to logging rather than this hack? https://mariadb.com/kb/en/error-log/#writing-the-error-log-to-stderr-on-unix suggests it should be possible to just log to stderr.

dtantsur:
/lgtm

I'm approving this to unblock the CI, but could we please take an action items to fix the mysql issue properly as soon as possible?

namnx228:
> /lgtm
> 
> I'm approving this to unblock the CI, but could we please take an action items to fix the mysql issue properly as soon as possible?

Sure, I am working on it.

PR #260, opened 2021-05-04
Fix the "permisison denied" bug properly

namnx228:
Issue: For some reason the script mysqld_safe_helper is used to run mysqld. This script runs with `--user=mysql` and `--log-error=/var/log/mariadb/mariadb.log`. This causes the `permisison denied` error because user mysql tries to  write to `/var/log/mariadb/mariadb.log` which, in this repo, is a symlink of root's stdout. 
Error message:
`/usr/bin/mysqld_safe_helper: Can't create/write to file '/var/log/mariadb/mariadb.log' (Errcode: 13 "Permission denied")`
This PR solves the problem by removing the symlink hack and make the error log go to stdout in a proper way.

### DIFF
--- a/scripts/runmariadb
+++ b/scripts/runmariadb
@@ -18,7 +18,6 @@ if [ ! -f "$MARIADB_CERT_FILE" ] && [ -f "$MARIADB_KEY_FILE" ] ; then
 fi
 
 ln -sf /proc/self/fd/1 /var/log/mariadb/mariadb.log
-rm -f /usr/bin/mysqld_safe_helper # Avoid this script being used which results in a permission denied error.
 
 # Restart mysqld when the certificate is updated
 if [[ -f "$MARIADB_CERT_FILE" && "${RESTART_CONTAINER_CERTIFICATE_UPDATED}" == "true" ]]; then
@@ -32,6 +31,7 @@ if [ ! -d "${DATADIR}/mysql" ]; then
     crudini --set "$MARIADB_CONF_FILE" mysqld max_heap_table_size 1M
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_buffer_pool_size 5M
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_log_buffer_size 512K
+    crudini --set "$MARIADB_CONF_FILE" mysqld user root
 
     # Config MariaDB to enable TLS
     if [ -f "$MARIADB_CERT_FILE" ]


scripts/runmariadb
```diff
@@ -16,14 +16,12 @@ if [ ! -f "$MARIADB_CERT_FILE" ] && [ -f "$MARIADB_KEY_FILE" ] ; then
...
 
-ln -sf /proc/self/fd/1 /var/log/mariadb/mariadb.log
-rm -f /usr/bin/mysqld_safe_helper # Avoid this script being used which results in a permission denied error.
-
 if [ ! -d "${DATADIR}/mysql" ]; then
     crudini --set "$MARIADB_CONF_FILE" mysqld max_connections 64
     crudini --set "$MARIADB_CONF_FILE" mysqld max_heap_table_size 1M
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_buffer_pool_size 5M
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_log_buffer_size 512K
+    crudini --set "$MARIADB_CONF_FILE" mariadb-10.3 skip_log_error # Error log will be redirected to stderr
```
dtantsur, line 24: Are you sure the group name is actually "mariadb-10.3", not mysqld as above?

namnx228: It should work in both group, I guess. I choose "mariadb-10.3" since in the example they use a similar group. https://mariadb.com/kb/en/error-log/#writing-the-error-log-to-stderr-on-unix.

scripts/runmariadb
```diff
@@ -16,15 +16,16 @@ if [ ! -f "$MARIADB_CERT_FILE" ] && [ -f "$MARIADB_KEY_FILE" ] ; then
...
-rm -f /usr/bin/mysqld_safe_helper # Avoid this script being used which results in a permission denied error.
-
 if [ ! -d "${DATADIR}/mysql" ]; then
     crudini --set "$MARIADB_CONF_FILE" mysqld max_connections 64
     crudini --set "$MARIADB_CONF_FILE" mysqld max_heap_table_size 1M
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_buffer_pool_size 5M
     crudini --set "$MARIADB_CONF_FILE" mysqld innodb_log_buffer_size 512K
 
+    # Error log will be redirected to stderr
+    crudini --set "$MARIADB_CONF_FILE" mariadb-10.3 skip_log_error
```
dtantsur, line 26: Let's maybe use mysqld so that we're not broken if/when centos updates mariadb?

namnx228:
@dtantsur I tried to use `skip-log-error`, but then mysqld_safe wrote the error log to `/var/log/mysql/<hostname>.err`. I go through the mysqld_safe script and see that it will write the log to a file rather than to stderr. 
Look like we have to use the symlink hack. So the possible solution now is to run mysqld by root rather than user mysql

elfosardo:
> @dtantsur I tried to use `skip-log-error`, but then mysqld_safe wrote the error log to `/var/log/mysql/<hostname>.err`. I go through the mysqld_safe script and see that it will write the log to a file rather than to stderr.
> Look like we have to use the symlink hack. So the possible solution now is to run mysqld by root rather than user mysql

@namnx228 have you tried adding the tty group to the mysqld user ?

namnx228:
> > @dtantsur I tried to use `skip-log-error`, but then mysqld_safe wrote the error log to `/var/log/mysql/<hostname>.err`. I go through the mysqld_safe script and see that it will write the log to a file rather than to stderr.
> > Look like we have to use the symlink hack. So the possible solution now is to run mysqld by root rather than user mysql
> 
> @namnx228 have you tried adding the tty group to the mysqld user ?

I haven't tried it. However, do we have a specific reason to avoid using root user here? I know that in general, root should not be used, but in this case, I have a feeling that using root doesn't cause any troubles.

elfosardo:
> > > @dtantsur I tried to use `skip-log-error`, but then mysqld_safe wrote the error log to `/var/log/mysql/<hostname>.err`. I go through the mysqld_safe script and see that it will write the log to a file rather than to stderr.
> > > Look like we have to use the symlink hack. So the possible solution now is to run mysqld by root rather than user mysql
> > 
> > 
> > @namnx228 have you tried adding the tty group to the mysqld user ?
> 
> I haven't tried it. However, do we have a specific reason to avoid using root user here? I know that in general, root should not be used, but in this case, I have a feeling that using root doesn't cause any troubles.

in general we should always adhere to the principle of least privilege, or do our best to
this applies on services running on servers as on containers
considering that there are workarounds to avoid running a service as root also in this case, IMHO we should give them a try before resorting to the run-as-root solution

namnx228:
@elfosardo I have tried to add `mysql` user to the `tty` group, but the permission denied still occurred, unfortunately.

elfosardo:
@namnx228 could you please provide the steps you used to do the test?
I did test that myself and it worked for me

namnx228:
> @namnx228 could you please provide the steps you used to do the test?
> I did test that myself and it worked for me
This is the commit that I push to test the tty group: https://github.com/metal3-io/ironic-image/commit/3f3d76bae5c9aec908ccec36e288a34f5bffb95a
I did the test locally in the same way as we do in the CI integration test. Maybe I could do something wrong.
I can cherry pick that commit onto the branch of this PR, then run the CI test. Let's see how it goes.

dtantsur:
/approve

Root inside a container doesn't give a potential attacker much, so I think we're fine with that.

PR #263, opened 2021-05-11
Allow Node network_data in any format

zaneb:
Currently, metal³ does not set the network_data field in the Ironic
Node. However it does pass the network_data upon provisioning, and
currently places no restrictions on the format of the network_data.

In order to pass the network_data to the deploy ramdisk, we need to also
set the network_data field in the Node. However, if the same
network_data is used then this would cause the data to start getting
validated against the Nova network_data schema, which may break existing
Hosts that had invalid or non-standard data.

To allow us to start passing the network_data without placing new
restrictions on the format, configure the schema to allow any JSON
object.


dtantsur:
> In order to pass the network_data to the deploy ramdisk, we need to also set the network_data field in the Node.

But why? I thought the plan was to make BMO request image building from a new controller via a new CR. Why do we need to make Ironic aware of that?

PR #264, opened 2021-05-19
Keep only one httpd instance for the Ironic deployment

namnx228:
Currently, when deploying ironic and ironic-inspector, there are two httpd instances. One is deployed in the script `runironic-api`, and this instance is used to serve file for IPA and to run as a WSGI web server for ironic-api. The other one is deployed in the script `runhttpd`, and it is used to also serve file for IPA and to run as a reverse proxy for ironic-inspector. This causes the problem when two httpd instances open the same port to serve file for IPA. This PR removes one instance to fix the issue. Also, this removal will simplify the configuration and boost the overall performance.

### DIFF
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -13,6 +13,10 @@ if [[ ! -z ${EXTRA_PKGS_LIST:-} ]]; then
     fi
 fi
 dnf install -y --enablerepo=epel inotify-tools
+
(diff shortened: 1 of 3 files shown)


elfosardo (approved):

dtantsur (approved):

namnx228:
The PR has passed the CI tests. Please take a look again. @elfosardo @dtantsur

PR #272, opened 2021-06-02
Add support for deriving the PROVISIONING_INTERFACE from a mac address

asalkeld:
The point of this PR is to allow users to:

1. not have to provide the PROVISIONING_INTERFACE (as it has to be the same for all hosts)
2. instead make use of the know mac addresses that we have in the bmh.Spec.BootMACAddress to calculate the interface

An example of how to do this is here (openshift specific): https://github.com/openshift/cluster-baremetal-operator/pull/149

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@ All of the containers must share a common mount point or data store.  Ironic req
 - final image to be deployed onto node in qcow2 format
 
 The following environment variables can be passed in to customize run-time functionality:
-- PROVISIONING_INTERFACE - interface to use for ironic, dnsmasq(dhcpd) and httpd (default provisioning)
+- PROVISIONING_MACS - a comma seperated list of mac address of the master nodes (used to determine the PROVISIONING_INTERFACE)
+- PROVISIONING_INTERFACE - interface to use for ironic, dnsmasq(dhcpd) and httpd (default provisioning, this is calculated if the above PROVISIONING_MACS is provided)
 - DNSMASQ_EXCEPT_INTERFACE - interfaces to exclude when providing DHCP address (default "lo")
 - HTTP_PORT - port used by http server (default 80)
 - DHCP_RANGE - dhcp range to use for provisioning (default 172.22.0.10-172.22.0.100)
--- a/scripts/ironic-common.sh
+++ b/scripts/ironic-common.sh
@@ -1,4 +1,21 @@
-export PROVISIONING_INTERFACE=${PROVISIONING_INTERFACE:-"provisioning"}
+function get_provisioning_interface() {
+  if [ -n "${PROVISIONING_INTERFACE}" ]; then
+    # don't override the PROVISIONING_INTERFACE if one is provided
+    echo ${PROVISIONING_INTERFACE}
+    return
+  fi
+
+  local interface="provisioning"
+  for mac in ${PROVISIONING_MACS//,/ } ; do
+    if ip -br link show up | grep -q "$mac"; then
+      interface=$(ip -br link show up | grep "$mac" | cut -f 1 -d ' ')
+      break
+    fi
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -29,7 +29,9 @@ All of the containers must share a common mount point or data store.  Ironic req
 - final image to be deployed onto node in qcow2 format
 
 The following environment variables can be passed in to customize run-time functionality:
-- PROVISIONING_INTERFACE - interface to use for ironic, dnsmasq(dhcpd) and httpd (default provisioning)
+- MAC_<nodename> - the mac address of the master nodes (used to determine the PROVISIONING_INTERFACE)
+- NODE_NAME - the name of the node that the container is running on
```
dtantsur, line 33: Why not just provide PROVISIONING_MAC? NODE_NAME does not seem to be required.

asalkeld: at the time that we deploy, we don't know what machine the deployment will run on, therefore we don't know what bmh mac to use. So what i am doing is setting the mac of each master and using the node name env, so the container can know which mac to use

imain: I do think we could just grep for the mac in all interfaces instead of having the node name.  However, this also works fine.

asalkeld:
I think our integration tests are broken

```
++ export CAPM3_VERSION=v1alpha5
++ CAPM3_VERSION=v1alpha5
++ CAPM3_VERSION_LIST=v1alpha4
++ echo v1alpha4
++ grep -wq v1alpha5
++ echo 'Invalid CAPM3 version : v1alpha5. Not in : v1alpha4'
Invalid CAPM3 version : v1alpha5. Not in : v1alpha4
++ exit 1
```

scripts/ironic-common.sh
```diff
@@ -1,4 +1,21 @@
-export PROVISIONING_INTERFACE=${PROVISIONING_INTERFACE:-"provisioning"}
+function get_provisioning_interface() {
+  if [ -n "${PROVISIONING_INTERFACE}" ]; then
+    # don't override the PROVISIONING_INTERFACE if one is provided
+    echo ${PROVISIONING_INTERFACE}
+    return
+  fi
+
+  mac_env_name=MAC_${NODE_NAME}
```
elfosardo, line 8: this should probably be a local variable

scripts/ironic-common.sh
```diff
@@ -1,4 +1,21 @@
...
+  mac=${!mac_env_name}
```
elfosardo, line 9: ditto

elfosardo (changes requested):

scripts/ironic-common.sh
asalkeld: will do (both)

scripts/ironic-common.sh
```diff
@@ -1,4 +1,22 @@
...
+function get_provisioning_interface() {
+  if [ -n "${PROVISIONING_INTERFACE}" ]; then
+    # don't override the PROVISIONING_INTERFACE if one is provided
+    echo ${PROVISIONING_INTERFACE}
+    return
+  fi
+
+  for mac in ${PROVISIONING_MACS//,/ } ; do
+    if ip -br link show up | grep -q "$mac"; then
+      export PROVISIONING_INTERFACE=$(ip -br link show up | grep "$mac" | cut -f 1 -d ' ')
```
elfosardo, line 10: shouldn't this return after it finds the PROVISIONING_INTERFACE value ?

elfosardo (changes requested):

scripts/ironic-common.sh
asalkeld: it echo's the value at the end of the function

dtantsur (commented):
/approve

PR #275, opened 2021-06-09
Add cipher_suite_versions config

iurygregory:
This commit adds the new ironic configuration option
``[ipmi]/cipher_suite_versions``.
Allowing ipmi to try different values in case of the
default cipher suite is not supported.


dtantsur:
Similarly to the openshift patch, please fix the syntax. Python lists are not accepted in conf files, you need `3,17`.

PR #277, opened 2021-06-15
Update the inspector endpoint to port 5049 when using the reverse proxy

namnx228:
When using httpd as the reverse proxy for ironic inspector, the httpd server will listen at 5050, and forward the traffic to port 5049. For that reason, we need to set the ironic inspector endpoint to this port. Otherwise, the traffic between ironic inspector and other components are not encrypted.
Also, this PR solves a problem with non-proxy case. Currently, if TLS is enabled, but we don't use the reverse proxy, it can cause problem. This PR solves that.

### DIFF
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -14,9 +14,6 @@ if [[ ! -z ${EXTRA_PKGS_LIST:-} ]]; then
 fi
 dnf install -y --enablerepo=epel inotify-tools
 
-# Delete the line below after the PR: https://github.com/metal3-io/baremetal-operator/pull/886 goes in.
-dnf install -y net-tools
-
 dnf clean all
 rm -rf /var/cache/{yum,dnf}/*
 if [[ ! -z ${PATCH_LIST:-} ]]; then
--- a/scripts/runhttpd
+++ b/scripts/runhttpd
@@ -50,8 +50,10 @@ if [ -f "$IRONIC_INSPECTOR_CERT_FILE" ]; then
     export IRONIC_INSPECTOR_TLS_SETUP="true"
     # Delete the line below after the PR: https://github.com/metal3-io/baremetal-operator/pull/886 goes in. 
     INSPECTOR_PORT=5050
-    # Delete the netstate and grep statement after the PR above goes in. 
-    netstat -ln | grep ${INSPECTOR_PORT} -q || render_j2_config $INSPECTOR_ORIG_HTTPD_CONFIG $INSPECTOR_RESULT_HTTPD_CONFIG
+
+    if [[ "${INSPECTOR_REVERSE_PROXY_SETUP}" == "true" ]]; then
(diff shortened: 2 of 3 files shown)


kashifest:
Would this require change in [dev-env](https://github.com/metal3-io/metal3-dev-env/blob/master/lib/network.sh#L163)  and [bmo](https://github.com/metal3-io/baremetal-operator/blob/master/ironic-deployment/default/ironic_bmo_configmap.env#L7)  as well??

Note that, in each of the repo there are multiple places where you need to change it if you need to

smoshiur1237:
> Would this require change in [dev-env](https://github.com/metal3-io/metal3-dev-env/blob/master/lib/network.sh#L163) and [bmo](https://github.com/metal3-io/baremetal-operator/blob/master/ironic-deployment/default/ironic_bmo_configmap.env#L7) as well??
> 
> Note that, in each of the repo there are multiple places where you need to change it if you need to

I think we don't need, as reverseproxy(5050) will communicate with ironic-inspector, in that case when we call ironic inspector the communication goes via reverse proxy(5050).

namnx228:
> > Would this require change in [dev-env](https://github.com/metal3-io/metal3-dev-env/blob/master/lib/network.sh#L163) and [bmo](https://github.com/metal3-io/baremetal-operator/blob/master/ironic-deployment/default/ironic_bmo_configmap.env#L7) as well??
> > Note that, in each of the repo there are multiple places where you need to change it if you need to
> 
> I think we don't need, as reverse proxy(5050) will communicate with ironic-inspector(5049), in that case when metal3 or bmo call ironic-inspector the communication goes via reverse proxy(5050).

@kashifest Moshiur is right, we don't need to change anywhere else.

dtantsur:
Will this work if reverse proxy is not used?

namnx228:
> Will this work if reverse proxy is not used?

Good question. Now it doesn't work if the reverse proxy is not used. I will update the PR to make it work.

namnx228:
> > Will this work if reverse proxy is not used?
> 
> Good question. Now it doesn't work if the reverse proxy is not used. I will update the PR to make it work.

@dtantsur I plan to open another PR to make it works with the reverse proxy soon since there is a couple of things needed to be done there. This PR is more like a bug fix because now the ironic-inspector and the reverse proxy are trying to open the same port 5050, so occasionally ironic-inspector crashes and the CI test fails. How do you think?

dtantsur:
I mean, you're pretty much breaking the non-proxied case, no? We cannot fix bugs by introducing more bugs.

Maybe we're ready to drop the non-proxied case - then we should do it explicitly.

namnx228:
> I mean, you're pretty much breaking the non-proxied case, no? We cannot fix bugs by introducing more bugs.
> 
> Maybe we're ready to drop the non-proxied case - then we should do it explicitly.

I think we should support the non-proxy case since someone may not want to use it. I have pushed another commit to this PR to solve the problem with non-proxy case. Since there are two different problems solved by this PR, I guess we should have two commits. I will update the description of the Pr. Please take a look again. Thanks.

PR #279, opened 2021-07-05
Revert "Wait for expected number of drivers starting API"

dtantsur:
The root cause has been fixed in Ironic.

This reverts commit 3c8acf4f3b26a5055362b091aec2dfeb8b06d89d.

### DIFF
--- a/scripts/runironic-api
+++ b/scripts/runironic-api
@@ -4,16 +4,6 @@ export IRONIC_DEPLOYMENT="API"
 
 . /bin/configure-ironic.sh
 
-# Wait for ironic to load all expected drivers
-# the DB query returns the number of unique hardware_type in the conductor_hardware_interfaces table
-CONF_DRIVERS=$(crudini --get /etc/ironic/ironic.conf DEFAULT enabled_hardware_types | tr ',' '\n' | wc -l)
-while true ; do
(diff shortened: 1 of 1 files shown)


andfasano:
> The root cause has been fixed in Ironic.
> 
> This reverts commit [3c8acf4](https://github.com/metal3-io/ironic-image/commit/3c8acf4f3b26a5055362b091aec2dfeb8b06d89d).

@dtantsur in which version this problem has been fixed? Does it mean that also [this readiness BMO check](https://github.com/metal3-io/baremetal-operator/blob/4a7f34ff2f0af32333ec483a809ac92c98f7f937/pkg/provisioner/ironic/dependencies.go#L64-L88) could be removed?

dtantsur:
> Does it mean that also this readiness BMO check could be removed?

No, the BMO check is actually correct. This check here was introduced to work around the bug fixed by https://review.opendev.org/c/openstack/ironic/+/764911.

andfasano:
> > Does it mean that also this readiness BMO check could be removed?
> 
> No, the BMO check is actually correct. This check here was introduced to work around the bug fixed by https://review.opendev.org/c/openstack/ironic/+/764911.

Ack, thanks

PR #288, opened 2021-07-26
Update tripleo_repos location

adetalhouet:
The upstream openstack tripleo team decided to move the `tripleo_repos` folder under `plugins/module_utils/` https://github.com/openstack/tripleo-repos/commit/77dc5794ac880abc57927af08e70e99d2a972f84
Due to this, when building the ironic container, we end us with a `404 Not Found` error.
Updating the script as per as the new upstream repo structure.

### DIFF
--- a/resources/vbmc/Dockerfile
+++ b/resources/vbmc/Dockerfile
@@ -1,7 +1,7 @@
 FROM docker.io/centos:centos8
 
 RUN dnf install -y python3 python3-requests python3-pip && \
-     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/plugins/module_utils/tripleo_repos/main.py | python3 | python3 - -b master current-tripleo --no-stream && \
+     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/plugins/module_utils/tripleo_repos/main.py | python3 - -b master current-tripleo --no-stream && \
      dnf upgrade -y && \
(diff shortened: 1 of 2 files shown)


elfosardo (changes requested):
we also need to change the Dockerfile of the vbmc image resources/vbmc/Dockerfile

adetalhouet:
> we also need to change the Dockerfile of the vbmc image resources/vbmc/Dockerfile
Ok, done. Thanks for reviewing.

resources/vbmc/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM docker.io/centos:centos8
 
 RUN dnf install -y python3 python3-requests python3-pip && \
-     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current-tripleo --no-stream && \
+     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/plugins/module_utils/tripleo_repos/main.py | python3 | python3 - -b master current-tripleo --no-stream && \
```
elfosardo, line 4: remove the new python 3 command

elfosardo (changes requested):

resources/vbmc/Dockerfile
adetalhouet: my bad. fixed

PR #291, opened 2021-08-05
Fix missing http port in iso url

elfosardo:
During virtualmedia deploy the port in the iso url is missing causing
the deploy to fail.

For more info please see https://bugzilla.redhat.com/show_bug.cgi?id=1990316

### DIFF
--- a/scripts/configure-ironic.sh
+++ b/scripts/configure-ironic.sh
@@ -34,6 +34,7 @@ fi
 
 . /bin/ironic-common.sh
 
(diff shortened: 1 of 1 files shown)


scripts/configure-ironic.sh
```diff
@@ -34,6 +34,7 @@ fi
 
 . /bin/ironic-common.sh
 
+export HTTP_PORT=${HTTP_PORT:-"80"}
```
namnx228, line 37: Should it be 6180?

PR #293, opened 2021-08-19
Add procps package to make `ps` available

namnx228:
`procps` is missing in centos 8 stream. However, we need it to use `ps` at some places, like this: https://github.com/metal3-io/ironic-image/blob/abc12be6df2a01e489e70b54aa7e291710fe1ef5/scripts/runhttpd#L91
This PS adds `procps` to the image, so `ps` is available.


dtantsur:
/lgtm

In the future could you keep the list sorted for sanity reasons?

PR #298, opened 2021-09-06
Remove the trailing slash from the heartbeat rule

dtantsur:
Just `/v1/heartbeat` is also valid.

### DIFF
--- a/ironic-config/apache2-ironic-api.conf.j2
+++ b/ironic-config/apache2-ironic-api.conf.j2
@@ -39,7 +39,7 @@ Listen 6385
     <Location /v1/lookup >
         Require all granted
     </Location>
-    <Location /v1/heartbeat/ >
+    <Location /v1/heartbeat >
         Require all granted
     </Location>
     <Location ~ "^/(v1/?)?$">


fmuyassarov (approved):

fmuyassarov (commented):
Would you mind to fix indentation issue as well on this PR @dtantsur or can I do it in a separate PR too
```

        {% if "HTTP_BASIC_HTPASSWD" in env and env.HTTP_BASIC_HTPASSWD | length %}
           AuthType Basic
           AuthName "Restricted WSGI area"
           AuthUserFile "/etc/ironic/htpasswd"
           Require valid-user
        {% else %}
           Require all granted
        {% endif %}
```
should be 
```
        {% if "HTTP_BASIC_HTPASSWD" in env and env.HTTP_BASIC_HTPASSWD | length %}
            AuthType Basic
            AuthName "Restricted WSGI area"
            AuthUserFile "/etc/ironic/htpasswd"
            Require valid-user
        {% else %}
            Require all granted
        {% endif %}
```

dtantsur:
I think it's fine to fix separately, it's not even the same spot.

fmuyassarov:
> I think it's fine to fix separately, it's not even the same spot.

I've opened https://github.com/metal3-io/ironic-image/pull/299 for that.

fmuyassarov:
@derekhiggins would you mind to approve once more? Prow should be up again now. Thanks.

PR #300, opened 2021-09-14
Bump sushy-tools to 0.18.0

iurygregory:
This commit bumps the version of sushy-tools to 0.18.0, this is the
latest release that includes new features.


bfournie (approved):

iurygregory:
Since prow is down I don't think the lgtm will work

PR #304, opened 2021-09-29
Add option to build image using source code directly

elfosardo:
This patch adds a build-time argument that allows to build the ironic container image using source code instead of RPMs
At the moment this option should only be used for testing purpose, to simplify building images using the latest code or local patches

It is possible to test it using metal3-dev-env with the patch https://github.com/metal3-io/metal3-dev-env/pull/839

### DIFF
--- a/ironic-source-list.txt
+++ b/ironic-source-list.txt
@@ -1,0 +1,11 @@
+crudini
+ironic @ git+https://opendev.org/openstack/ironic
+ironic-inspector @ git+https://opendev.org/openstack/ironic-inspector
+ironic-prometheus-exporter
+proliantutils
+PyMySQL>=0.8.0
+python-dracclient
+python-ibmcclient
+python-scciclient
+sushy
+sushy-oem-idrac
--- a/main-packages-list.txt
+++ b/main-packages-list.txt
@@ -1,4 +1,3 @@
-crudini
 dnsmasq
 genisoimage
 httpd
@@ -8,20 +7,11 @@ ipmitool
 iproute
 mariadb-server
 mod_ssl
-openstack-ironic-api
-openstack-ironic-conductor
-openstack-ironic-inspector
+parted
 procps
-python3-dracclient
-python3-gunicorn
-python3-ibmcclient
-python3-ironic-prometheus-exporter
+psmisc
 python3-jinja2
 python3-mod_wsgi
-python3-proliantutils
-python3-scciclient
-python3-sushy
-python3-sushy-oem-idrac
 qemu-img
 sqlite
 syslinux-nonlinux
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -2,28 +2,54 @@
 
 set -euxo pipefail
 
+IRONIC_PKG_LIST=/tmp/ironic-${INSTALL_TYPE}-list.txt
+
 echo "install_weak_deps=False" >> /etc/dnf/dnf.conf
 # Tell RPM to skip installing documentation
 echo "tsflags=nodocs" >> /etc/dnf/dnf.conf
 
 dnf install -y python3 python3-requests epel-release 'dnf-command(config-manager)'
 dnf config-manager --set-disabled epel
-curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/plugins/module_utils/tripleo_repos/main.py | python3 - -b master current-tripleo
-dnf upgrade -y
+
+# RPM install #
+if [[ $INSTALL_TYPE == "rpm" ]]; then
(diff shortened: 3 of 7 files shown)


prepare-image.sh
```diff
@@ -2,28 +2,52 @@
...
+    dnf upgrade -y
+    xargs -rtd'\n' dnf install -y < $IRONIC_PKG_LIST
+fi
+
+# SOURCE install #
+if [[ $INSTALL_TYPE == "source" ]]; then
+    dnf upgrade -y
+    dnf install -y python3-pip python3-devel gcc
+    pip3 install --upgrade pip
+    pip3 install -r $IRONIC_PKG_LIST
```
dtantsur, line 26: Please use upper-constraints.

prepare-image.sh
```diff
@@ -2,28 +2,52 @@
...
+if [[ $INSTALL_TYPE == "rpm" ]]; then
+    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/plugins/module_utils/tripleo_repos/main.py | python3 - -b master current-tripleo
+    dnf upgrade -y
+    xargs -rtd'\n' dnf install -y < $IRONIC_PKG_LIST
+fi
+
+# SOURCE install #
+if [[ $INSTALL_TYPE == "source" ]]; then
+    dnf upgrade -y
+    dnf install -y python3-pip python3-devel gcc
```
dtantsur, line 24: You'll probably need a few other development libraries: ffi, ssl, etc.

ironic-source-list.txt
```diff
@@ -0,0 +1,11 @@
+crudini
```
namnx228, line 1: This one is on both lists. Should we keep it in main-package list?

elfosardo:
it works! \o/

main-packages-list.txt
```diff
@@ -10,22 +9,11 @@ iproute
 iscsi-initiator-utils
```
dtantsur, line 9: Side note: I wonder why we need this...

prepare-image.sh
```diff
@@ -2,28 +2,55 @@
...
+    pip3 install --upgrade pip
+    pip3 install --prefix /usr -r $IRONIC_PKG_LIST -c https://raw.githubusercontent.com/openstack/requirements/master/upper-constraints.txt
```
dtantsur, line 26: Ah, so you use the upstream constraints. This is fine, but do you realize that the services and vendor dependencies are not there?

namnx228 (commented):
/lgtm

prepare-image.sh
```diff
@@ -2,28 +2,54 @@
...
 xargs -rtd'\n' dnf install -y < /tmp/${PKGS_LIST}
+
 if [[ ! -z ${EXTRA_PKGS_LIST:-} ]]; then
     if [[ -s /tmp/${EXTRA_PKGS_LIST} ]]; then
         xargs -rtd'\n' dnf install -y < /tmp/${EXTRA_PKGS_LIST}
     fi
 fi
-dnf install -y --enablerepo=epel inotify-tools
+
+dnf install -y --enablerepo=epel inotify-tools python3-gunicorn
```
dtantsur, line 44: gunicorn can be installed from pypi just as well

prepare-image.sh
```diff
@@ -2,28 +2,54 @@
...
+    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/plugins/module_utils/tripleo_repos/main.py | python3 - -b master current-tripleo
+    dnf upgrade -y
+    xargs -rtd'\n' dnf install -y < $IRONIC_PKG_LIST
+fi
+
+# SOURCE install #
+if [[ $INSTALL_TYPE == "source" ]]; then
+    dnf upgrade -y
+    dnf install -y python3-pip python3-devel gcc git-core
+    pip3 install --upgrade pip
```
dtantsur, line 25: I'd rather avoid an unconditional upgrade of pip, we've had enough issues with it upstream

prepare-image.sh
```diff
@@ -2,28 +2,54 @@
...
+if [[ $INSTALL_TYPE == "rpm" ]]; then
+    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/plugins/module_utils/tripleo_repos/main.py | python3 - -b master current-tripleo
+    dnf upgrade -y
+    xargs -rtd'\n' dnf install -y < $IRONIC_PKG_LIST
+fi
+
+# SOURCE install #
+if [[ $INSTALL_TYPE == "source" ]]; then
+    dnf upgrade -y
+    dnf install -y python3-pip python3-devel gcc git-core
```
dtantsur, line 24: Should be removed from the final image

dtantsur:
/approve

Works as a proof of concept, although I'd see a few things cleaned up next.

PR #305, opened 2021-10-01
Install nc for checking health of services

Xenwar:
Install `nc` In order to test readiness and liveness of services. Some of the services.

### DIFF
--- a/main-packages-list.txt
+++ b/main-packages-list.txt
@@ -9,6 +9,7 @@ iproute
 iscsi-initiator-utils
 mariadb-server
 mod_ssl
+nc
 openstack-ironic-api
 openstack-ironic-conductor
 openstack-ironic-inspector


kashifest:
lgtm

Rozzii (approved):

elfosardo:
@Xenwar thank you for the patch, can you please provide more information on how the nc package will be used? I see only the installation here but no command added

Dockerfile
```diff
@@ -35,6 +35,7 @@ RUN if [ $(uname -m) = "x86_64" ]; then \
 
 FROM $BASE_IMAGE
 
+RUN dnf install -y nc
```
elfosardo, line 38: the nc package should be added to the list of packages in main-packages-list.txt

elfosardo (changes requested):

Xenwar:
> @Xenwar thank you for the patch, can you please provide more information on how the nc package will be used? I see only the installation here but no command added

The usage can be seen on this PR. https://github.com/metal3-io/baremetal-operator/pull/975/files
I could not add it as it is done at a Pod level.

Dockerfile
Xenwar: Thanks, will move it there

Xenwar: done.

Xenwar:
To give some context, I am trying to add readiness/liveness probes at a Pod level. To that end, one can use service specific utilities such as dhcpclient, tftp client and dns client binaries. However, this would make the image unnecessarily large. 
I chose `nc` as it is both generic and small.

Xenwar:
@elfosardo I have made the changes. Do you have more feedback on this PR ? 
Thanks.

elfosardo:
@Xenwar are there any alternatives to installing nc?
I'm not particularly inclined to install more packages if they're not strictly related to ironic or its components

Xenwar:
> @Xenwar are there any alternatives to installing nc? I'm not particularly inclined to install more packages if they're not strictly related to ironic or its components

There are other binaries as well such as dhcp-client, but if we use nc, it covers them.
If there is another way of verifying the health of ironic, then I am also ok with not installing nc.

PR #320, opened 2021-11-08
Create EFI esp.img on aarch64

yselkowitz:


dtantsur:
/approve
/test-v1b1-integration 

I guess we could optimize this bit eventually (maybe even add a separate script). But let's start here.

dtantsur:
Refactoring: https://github.com/metal3-io/ironic-image/pull/322

PR #322, opened 2021-11-09
Move EFI partition preparation into a script

dtantsur:


prepare-efi.sh
```diff
@@ -0,0 +1,34 @@
...
+## be copied in, however that requires more advanced scripting beyond
+## an MVP.
+dd bs=1024 count=6400 if=/dev/zero of=$DEST
+mkfs.msdos -F 12 -n 'ESP_IMAGE' $DEST
+
+mmd -i $DEST EFI
+mmd -i $DEST EFI/BOOT
+mcopy -i $DEST -v /boot/efi/EFI/BOOT/$BOOTEFI ::EFI/BOOT
+mcopy -i $DEST -v /boot/efi/EFI/$OS/$GRUBEFI ::EFI/BOOT
+mdir -i $DEST ::EFI/BOOT;
```
elfosardo, line 34: you can remove the semicolon at the end

Dockerfile
```diff
@@ -13,31 +13,8 @@ RUN git clone --depth 1 --branch v1.21.1 https://github.com/ipxe/ipxe.git && \
...
-      mmd -i esp.img EFI && \
-      mmd -i esp.img EFI/BOOT && \
-      mcopy -i esp.img -v /boot/efi/EFI/BOOT/BOOTAA64.EFI ::EFI/BOOT && \
-      mcopy -i esp.img -v /boot/efi/EFI/centos/grubaa64.efi ::EFI/BOOT && \
-      mdir -i esp.img ::EFI/BOOT; \
-    else \
-      touch /tmp/esp.img; \
-    fi
+COPY prepare-efi.sh /bin/
+RUN prepare-efi.sh centos
```
elfosardo, line 17: nit: the default is centos so that can be removed

dtantsur: I want to be explicit because it's going to be different downstream

prepare-efi.sh
dtantsur: Will change if I need to update (I guess it's not a blocker?)

namnx228 (commented):
/lgtm

elfosardo:
/approve
lgtm'ed also by @namnx228 which is in the reviewers list

PR #323, opened 2021-11-09
Refactor patch-image script

elfosardo:


patch-image.sh
```diff
@@ -24,7 +18,6 @@ do
 
     SKIP_GENERATE_AUTHORS=1 SKIP_WRITE_GIT_CHANGELOG=1 python3 setup.py sdist
     pip3 install dist/*.tar.gz
```
yanndegat, line 20: hi @elfosardo 
in the other scripts (configure-ironic.sh and prepare-image.sh), the pip install command is used with the `--prefix /usr` option.

if ever we want to patch a lib, it may conficts by keeping them in /usr/local here, don't you think it may conficts with the ones potentially already installed in /usr ?

elfosardo: hi, thanks for the review
I can't see where pip is used in configure-ironic.sh
about prepare-image.sh, I haven't seen any conflict happen, but we should probably change this as well for consistency
I'll update the patch

yanndegat: ok. thanks, maybe it was a local change on my box for configure-ironic.sh

i've made this https://github.com/yanndegat/irobox this morning, do you think it could be useful somehow ?

PR #325, opened 2021-11-19
change "/bin" refs in scripts to relative path

yanndegat:


dtantsur:
Do we even need any path? `/bin` should be on the default `PATH`, maybe just use script names?

kashifest:
/hold
Until the default branch is updated to main.

kashifest:
/hold cancel
Default branch is now main. Please rebase on top of main for future reference.

yanndegat:
hi @namnx228 

it seems to me that the integration tests are failing due to an issue unrelated with this PR. do you confirm ?

here's an extract of the error i found in the jenkins job
```
Installing cert-manager Version="v1.5.3"
Fetching File="cert-manager.yaml" Provider="cert-manager" Type="" Version="v1.5.3"
Retrying with backoff Cause="failed to get cert-manager object /, Kind=, /: Object 'Kind' is missing in 'unstructured object has no kind'"
Retrying with backoff Cause="failed to get cert-manager object /, Kind=, /: Object 'Kind' is missing in 'unstructured object has no kind'"
Retrying with backoff Cause="failed to get cert-manager object /, Kind=, /: Object 'Kind' is missing in 'unstructured object has no kind'"
Retrying with backoff Cause="failed to get cert-manager object /, Kind=, /: Object 'Kind' is missing in 'unstructured object has no kind'"
Retrying with backoff Cause="failed to get cert-manager object /, Kind=, /: Object 'Kind' is missing in 'unstructured object has no kind'"
Retrying with backoff Cause="failed to get cert-manager object /, Kind=, /: Object 'Kind' is missing in 'unstructured object has no kind'"
Retrying with backoff Cause="failed to get cert-manager object /, Kind=, /: Object 'Kind' is missing in 'unstructured object has no kind'"
Retrying with backoff Cause="failed to get cert-manager object /, Kind=, /: Object 'Kind' is missing in 'unstructured object has no kind'"
Retrying with backoff Cause="failed to get cert-manager object /, Kind=, /: Object 'Kind' is missing in 'unstructured object has no kind'"
Error: action failed after 10 attempts: failed to get cert-manager object /, Kind=, /: Object 'Kind' is missing in 'unstructured object has no kind'
```

elfosardo (changes requested):
I believe you're forgetting that all the scripts are copied under /bin/ during build time
https://github.com/metal3-io/ironic-image/blob/main/Dockerfile#L32

yanndegat:
Hi @elfosardo 

yes, the needs is to be able to mount a volume which contains patched scripts. 
but as it's /bin, you can't mount a source volume of the patched scripts without having to include all other system binaries

PR #327, opened 2021-12-02
Move RUN commands to prepare-image

elfosardo:
Move change ownership of /var/log/ironic
Aggregate file removals
Remove file removal duplicate


namnx228 (commented):
/lgtm

PR #330, opened 2021-12-08
Add a combined Ironic launcher

dtantsur:
This series of commits adds a new entrypoint `/bin/runironic` that uses the new combined `ironic` executable to start both API and conductor in one process. TLS is expected to be handled by a reverse proxy setup, similar to Inspector. The default for the new executable is to be backed by SQLite, not MariaDB for a smaller footprint.


dtantsur:
Note for testing: this patch works, but requires an Ironic change that hasn't made it to the RDO repositories yet. Use source build if you want to try it.

elfosardo (approved):

namnx228 (commented):
/lgtm

PR #332, opened 2021-12-09
Do not remove pip in the source build

dtantsur:
For whatever reason, Python itself depends on it in CentOS.

### DIFF
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -20,9 +20,10 @@ fi
 
 # SOURCE install #
 if [[ $INSTALL_TYPE == "source" ]]; then
-    BUILD_DEPS="python3-pip python3-devel gcc git-core"
+    BUILD_DEPS="python3-devel gcc git-core"
     dnf upgrade -y
-    dnf install -y $BUILD_DEPS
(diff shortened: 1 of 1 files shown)


elfosardo:
/approve
this is unfortunately true :/

prepare-image.sh
```diff
@@ -20,9 +20,10 @@ fi
 
 # SOURCE install #
 if [[ $INSTALL_TYPE == "source" ]]; then
-    BUILD_DEPS="python3-pip python3-devel gcc git-core"
+    BUILD_DEPS="python3-devel gcc git-core"
     dnf upgrade -y
-    dnf install -y $BUILD_DEPS
+    # NOTE(dtantsur): pip is a requirement of python3 in CentOS
+    dnf install -y python3-pip $BUILD_DEPS
```
hardys, line 26: Not directly related to this PR, but I wonder do we still need the pinned `21.3.1` version below, or can we just use the packaged version?

elfosardo: we could, although packaged version is different between distributions and in certain cases not recent enough to deal with some recent builds.
Since we're using source code for most of the ironic software/libraries/dependencies, I'd rather use also a fixed recent pip version.

PR #338, opened 2021-12-15
Use IPv6-friendly URLs for CoreOS

zaneb:


zaneb:
(This was an accidental dup of #337)

PR #344, opened 2022-01-14
Fix authentication with Ironic reverse proxy

dtantsur:
1) The authentication code assumes /v1 prefix, but Ironic has redirects,
   e.g. /drivers -> /v1/drivers
2) The lookup/heartbeat endpoint rules are overridden by the generic
   rule, so they don't have effect.

While here, fix some whitespace.


elfosardo (approved):

iurygregory (approved):
/lgtm

PR #346, opened 2022-01-18
Call configure-coreos-ipa in httpd to create the Ignition file

zhouhao3:
Fixed the issue of missing Ignition file during IPI deployment.
configure-coreos-ipa is not called in httpd and the IGNITION_FILE is not created,
so the address of IGNITION_FILE is not added to the boot kernel parameters.

Fixes: https://github.com/openshift/installer/issues/5542

Signed-off-by: Zhou Hao <zhouhao@fujitsu.com>

### DIFF
--- a/scripts/runhttpd
+++ b/scripts/runhttpd
@@ -4,6 +4,8 @@
 
 . /bin/ironic-common.sh
 
+. /bin/configure-coreos-ipa
+
 HTTP_PORT=${HTTP_PORT:-"80"}
 export VMEDIA_TLS_PORT=${VMEDIA_TLS_PORT:-8083}
 


dtantsur:
/cc @zaneb 

I don't think this is correct. The configure-coreos-ipa script is unused in the openshift context, ignition is appended in image-customization-controller, not here.

This approach *may* make sense for the upstream ironic-image, but then you need to make configure-coreos-ipa handle the absence of the `IRONIC_AGENT_IMAGE` variable (since it's absent now and has no default).

zaneb:
I agree with Dmitry.

I'm tempted to say we should delete the configure-coreos-ipa script altogether, but I guess somebody upstream may want to use it to deploy using Fedora CoreOS without an image controller? (OTOH it's not tested anywhere and will probably break soon.)

In any event, I don't think this makes sense to roll into the runhttpd script. When we were using it for development testing we ran it as a separate init container with its own environment, and I think that's the way to do it.

/close

PR #356, opened 2022-02-23
Clean packages after install and added some env vars to sushy-tools container

e-minguez:


resources/sushy-tools/Dockerfile
```diff
@@ -1,7 +1,15 @@
...
 RUN apt update && \
     apt install -y libvirt-dev && \
-    pip3 install sushy-tools==0.18.1 libvirt-python
+    apt clean && \
+    pip3 install --no-cache-dir sushy-tools==${SUSHY_TOOLS_VERSION} libvirt-python
+
+ENV SUSHY_EMULATOR_CONFIG="/root/sushy/conf.py"
+ENV SUSHY_EMULATOR_LISTEN_IP="::"
+ENV SUSHY_EMULATOR_LISTEN_PORT="8000"
+ENV SUSHY_EXTRA_FLAGS="--debug"
```
dtantsur, line 13: Should be `SUSHY_EMULATOR_EXTRA_FLAGS` for consistency

resources/sushy-tools/Dockerfile
```diff
@@ -1,7 +1,15 @@
...
 
+ENV SUSHY_TOOLS_VERSION="0.18.1"
+
 RUN apt update && \
     apt install -y libvirt-dev && \
-    pip3 install sushy-tools==0.18.1 libvirt-python
+    apt clean && \
+    pip3 install --no-cache-dir sushy-tools==${SUSHY_TOOLS_VERSION} libvirt-python
+
+ENV SUSHY_EMULATOR_CONFIG="/root/sushy/conf.py"
```
dtantsur, line 10: Is this really useful given that it's inside a container?

e-minguez: The thing is if you want to specify a custom config you need to overwrite the CMD as:

```
sudo podman run -d --privileged --rm --name sushy-tools -v /home/core/sushy-tools/sushy-emulator.conf:/etc/sushy/sushy-emulator.conf:Z -v /var/run/libvirt:/var/run/libvirt:z -e SUSHY_EMULATOR_CONFIG=/etc/sushy/sushy-emulator.conf -p 8000:8000 quay.io/metal3-io/sushy-tools:latest sushy-emulator
```

otherwise it will use the default `/root/sushy/conf.py` file instead and will fail:

```
Traceback (most recent call last):
  File "/usr/local/bin/sushy-emulator", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/sushy_tools/emulator/main.py", line 773, in main
    app.configure(config_file=args.config)
  File "/usr/local/lib/python3.9/site-packages/sushy_tools/emulator/main.py", line 85, in configure
    self.config.from_pyfile(config_file)
  File "/usr/local/lib/python3.9/site-packages/flask/config.py", line 119, in from_pyfile
    with open(filename, mode="rb") as config_file:
FileNotFoundError: [Errno 2] Unable to load configuration file (No such file or directory): '/root/sushy/conf.py'
```

dtantsur: Cannot you mount if over the existing path? Anyway, it's not an objection, I'm just asking.

resources/sushy-tools/Dockerfile
e-minguez: To be fair, that's not a sushy-emulator env variable, that's why I choose to _rename_ it:

```
sushy-emulator -h
usage: sushy-emulator [-h] [--config CONFIG] [--debug] [-i INTERFACE] [-p PORT] [--ssl-certificate SSL_CERTIFICATE]
                      [--ssl-key SSL_KEY] [--os-cloud OS_CLOUD | --libvirt-uri LIBVIRT_URI]

optional arguments:
  -h, --help            show this help message and exit
  --config CONFIG       Config file path. Can also be set via environment variable SUSHY_EMULATOR_CONFIG.
  --debug               Enables debug mode when running sushy-emulator.
  -i INTERFACE, --interface INTERFACE
                        IP address of the local interface to listen at. Can also be set via config variable
                        SUSHY_EMULATOR_LISTEN_IP. Default is all local interfaces.
  -p PORT, --port PORT  TCP port to bind the server to. Can also be set via config variable SUSHY_EMULATOR_LISTEN_PORT. Default is
                        8000.
  --ssl-certificate SSL_CERTIFICATE
                        SSL certificate to use for HTTPS. Can also be set via config variable SUSHY_EMULATOR_SSL_CERT.
  --ssl-key SSL_KEY     SSL key to use for HTTPS. Can also be setvia config variable SUSHY_EMULATOR_SSL_KEY.
  --os-cloud OS_CLOUD   OpenStack cloud name. Can also be set via environment variable OS_CLOUD or config variable
                        SUSHY_EMULATOR_OS_CLOUD.
  --libvirt-uri LIBVIRT_URI
                        The libvirt URI. Can also be set via environment variable SUSHY_EMULATOR_LIBVIRT_URI. Default is
                        qemu:///system
```

But I can modify it `SUSHY_EMULATOR_EXTRA_FLAGS`, sure

resources/sushy-tools/Dockerfile
e-minguez: You are right, this works:

```
sudo podman run -d --privileged --rm --name sushy-tools -v /home/core/sushy-tools/sushy-emulator.conf:/root/sushy/conf.py:Z -v /var/run/libvirt:/var/run/libvirt:z -p 8000:8000 quay.io/metal3-io/sushy-tools:latest
```

e-minguez: So I guess it would be better then to have something like:

```
ENV SUSHY_EMULATOR_LISTEN_IP="::"
ENV SUSHY_EMULATOR_LISTEN_PORT="8000"
ENV SUSHY_EMULATOR_CONFIG="/root/sushy/conf.py"
ENV SUSHY_EMULATOR_EXTRA_FLAGS="--debug"

ENTRYPOINT /usr/local/bin/sushy-emulator
CMD ${SUSHY_EXTRA_FLAGS}
```

Or should I get rid of all those flags and keep the CMD as it is now?

Thanks!

resources/sushy-tools/Dockerfile
```diff
@@ -1,7 +1,14 @@
...
-    pip3 install sushy-tools==0.18.1 libvirt-python
+    apt clean && \
+    pip3 install --no-cache-dir sushy-tools==${SUSHY_TOOLS_VERSION} libvirt-python
+
+ENV SUSHY_EMULATOR_LISTEN_IP="::"
+ENV SUSHY_EMULATOR_LISTEN_PORT="8000"
+ENV SUSHY_EMULATOR_EXTRA_FLAGS="--debug"
 
-CMD sushy-emulator -i :: -p 8000 --config /root/sushy/conf.py --debug
+CMD sushy-emulator ${SUSHY_EXTRA_FLAGS}
```
dtantsur, line 14: I think you need to return `--config /root/sushy/conf.py` now.

resources/sushy-tools/Dockerfile
```diff
@@ -1,7 +1,14 @@
...
+ENV SUSHY_TOOLS_VERSION="0.18.1"
+
 RUN apt update && \
     apt install -y libvirt-dev && \
-    pip3 install sushy-tools==0.18.1 libvirt-python
+    apt clean && \
+    pip3 install --no-cache-dir \
+        sushy-tools==${SUSHY_TOOLS_VERSION} libvirt-python
+
+ENV SUSHY_EMULATOR_EXTRA_FLAGS="--debug"
```
elfosardo, line 11: let's move this under the first ENV and compress to a single ENV

resources/sushy-tools/Dockerfile
```diff
@@ -1,7 +1,14 @@
 FROM registry.hub.docker.com/library/python:3.9
 
+ENV SUSHY_TOOLS_VERSION="0.18.1"
```
elfosardo, line 3: for a followup, would be probably interesting to convert both ENVs to ARGs to give flexibility during build time

elfosardo (changes requested):

resources/sushy-tools/Dockerfile
e-minguez: If I [read correctly](https://docs.docker.com/engine/reference/builder/#arg), ARGs are used for builds only and ENVs for build+run. This one definitely is an ARG type as it is required in the build phase, but the other one is definitely a 'run' TMK.
WDYT?

resources/sushy-tools/Dockerfile
e-minguez: If finally using `SUSHY_TOOLS_VERSION` as ARG (see the other comment), I guess it wouldn't make sense to have them both as ENV.

resources/sushy-tools/Dockerfile
elfosardo: you can set an ENV to be the value of an ARG to give more flexibility during build time, I should have probably started with this :)

resources/sushy-tools/Dockerfile
elfosardo: of course not, this comment is about a change that should be applied now, considering that both are ENVs
the other comment was about a possible follow-up, to not add complexity to this change
all considered "--debug" does not need to be an ENV in this change

elfosardo (changes requested):

resources/sushy-tools/Dockerfile
e-minguez: I think I got it. I've created an ARG for the `sushy-tools` version and an ENV for the extra flags. Do you mind to take a look again? Thanks!

resources/sushy-tools/Dockerfile
e-minguez: > you can set an ENV to be the value of an ARG to give more flexibility during build time, I should have probably started with this :)

TIL, thanks!

resources/sushy-tools/Dockerfile
elfosardo: the SUSHY_TOOLS_VERSION ARG is ok, but what I was really saying is that I'm not sure SUSHY_EMULATOR_EXTRA_FLAGS should be an ENV, you're creating one more layer for a thing that could be included in the CMD and does not need to persist in other parts of the Dockerfile

elfosardo (changes requested):

resources/sushy-tools/Dockerfile
e-minguez: ack

PR #360, opened 2022-03-15
Add option to choose TLS version

namnx228:


ironic-config/apache2-vmedia.conf.j2
```diff
@@ -7,6 +7,7 @@ Listen {{ env.VMEDIA_TLS_PORT }}
 
     ServerName {{ env.IRONIC_IP }}
     SSLEngine on
+    SSLProtocol {{ env.IRONIC_SSL_PROTOCOL }}
```
dtantsur, line 10: I have concerns about this one. It is used to expose files to BMCs, and BMCs may have older crypto. I'd leave this alone OR create a new option defaulting to "ALL".

PR #361, opened 2022-03-16
Consolidate efi related code into prepare-efi script

elfosardo:


prepare-efi.sh
```diff
@@ -14,12 +14,29 @@ elif [ "$ARCH" = "aarch64" ]; then
     PACKAGES=grub2-efi-aa64
     BOOTEFI=BOOTAA64.EFI
     GRUBEFI=grubaa64.efi
+    ARCH=$(echo $ARCH | sed 's/aarch/arm/')
```
dtantsur, line 17: nit:

 ARCH=arm64

:)

prepare-efi.sh
```diff
@@ -14,12 +14,29 @@ elif [ "$ARCH" = "aarch64" ]; then
...
 else
     echo "WARNING: don't know how to build an EFI image on $ARCH"
     touch "$DEST"
     exit 0
 fi
 
+if [ "$OS" = "centos" ]; then
```
dtantsur, line 24: If you want to have this in a script, let's create `build-ipxe` or something. This is not EFI-specific.

prepare-efi.sh
elfosardo: too easy! :)

prepare-efi.sh
elfosardo: you mean a separate script?
my idea is that having all into one script would simplify downstream backport

derekhiggins: This block creates files that are referenced in Dockerfile causing it to fail if $OS != "centos",
I guess this is ok as downstream it will be used by another Dockerfile? if so can you add a comment to explain

derekhiggins (commented):
When I try to build the Dockerfile locally dnf is being killed, have you seen this elsewhere?
> $ podman build -f Dockerfile
> ...
>  Installing       : python3-oslo-messaging-12.11.1-0.20211220094843.   149/166xargs: dnf: terminated by signal 11
> Error: error building at STEP "RUN prepare-image.sh &&   rm -f /bin/prepare-image.sh": error while running runtime: exit status 125

prepare-efi.sh
derekhiggins: This block creates files that are referenced in Dockerfile causing it to fail if $OS != "centos",
I guess this is ok as downstream it will be used by another Dockerfile? if so can you add a comment to explain

elfosardo:
> dnf: terminated by signal 11

@derekhiggins that looks like a segmentation fault, but it's in a part not touched by this patch
anyway, I'll run a test and see if I can reproduce

prepare-efi.sh
dtantsur: > my idea is that having all into one script would simplify downstream backport

Then we need to rename this script? Dunno, I'd not mix two unrelated things.

elfosardo:
/hold
need some rework if we still want to unify the code in one single script

PR #362, opened 2022-03-17
Fix missing semi-colon in runironic-inspector script

elfosardo:
Also add verbosity and exit on error


furkatgofurov7 (approved):
/approve

PR #364, opened 2022-03-21
Move python3-pip install into patch-image script

elfosardo:
It's not needed if not for the patch-image script, so install it only
when we run that script.


furkatgofurov7 (approved):
This should also work fine, thanks @elfosardo!

PR #372, opened 2022-05-24
Add option to to listen on a single inspector interface

mboukhalfa:
When running ironic without TLS inspector is running without reverse proxy and it always listening on all interfaces, since we have already introduced an option to listen on a single interface #371 this PR add the option in the case we run ironic without TLS


Rozzii (approved):
lgtm

mboukhalfa:
cc @elfosardo @zaneb

PR #376, opened 2022-07-28
Pin Werkzeug to 2.1.2 until newer version of sushy-tools is out

elfosardo:
Werkzeug 2.2.0 broke behavior of sushy-tools until version 0.20.0
is out.


dtantsur:
/approve
/lgtm

I can see the right version installed in the logs. Merging with a single review to unblock everyone using sushy-tools.

PR #381, opened 2022-08-19
Inspector: Set server name from variable

lentzi90:
ServerName was hard coded. Not quite sure if we should use IRONIC_IP here or perhaps IRONIC_URL_HOST or something?


dtantsur:
/lgtm

I think IRONIC_IP is the closest thing we can provide

PR #388, opened 2022-10-14
Change variable to environment

shibaPuppy:
```
cat /shared/html/inspector.ipxe  |grep callback
```

```
cat /shared/html/inspector.ipxe  |grep ipa
kernel --timeout 60000 http://x:6180/images/ironic-python-agent.kernel ipa-insecure=1 ipa-inspection-collectors=default,extra-hardware,logs systemd.journald.forward_to_console=yes BOOTIF=${mac} ipa-debug=1 ipa-enable-vlan-interfaces= ipa-inspection-dhcp-all-interfaces=1 ipa-collect-lldp=1  initrd=ironic-python-agent.initramfs console=ttyS0 ipa-ntp-server=x || goto retry_boot
```

ipa-inspection-callback-url value is empty.

### DIFF
--- a/scripts/configure-httpd-ipa.sh
+++ b/scripts/configure-httpd-ipa.sh
@@ -20,6 +20,7 @@ INSPECTOR_EXTRA_ARGS=" ipa-inspection-callback-url=${IRONIC_BASE_URL}:${IRONIC_I
 if [[ $IRONIC_FAST_TRACK == true ]]; then
     INSPECTOR_EXTRA_ARGS+=" ipa-api-url=${IRONIC_BASE_URL}:${IRONIC_ACCESS_PORT}"
 fi
+export INSPECTOR_EXTRA_ARGS
 
 # Copy files to shared mount
 render_j2_config /tmp/inspector.ipxe.j2 /shared/html/inspector.ipxe
--- a/scripts/runhttpd
+++ b/scripts/runhttpd
@@ -26,6 +26,7 @@ INSPECTOR_EXTRA_ARGS=" ipa-inspection-callback-url=${IRONIC_BASE_URL}:${IRONIC_I
 if [[ $IRONIC_FAST_TRACK == true ]]; then
     INSPECTOR_EXTRA_ARGS+=" ipa-api-url=${IRONIC_BASE_URL}:${IRONIC_ACCESS_PORT}"
 fi
+export INSPECTOR_EXTRA_ARGS
 
 . /bin/coreos-ipa-common.sh
 


scripts/configure-httpd-ipa.sh
```diff
@@ -16,7 +16,7 @@ else
     IRONIC_BASE_URL="http://${IRONIC_URL_HOST}"
 fi
 
-INSPECTOR_EXTRA_ARGS=" ipa-inspection-callback-url=${IRONIC_BASE_URL}:${IRONIC_INSPECTOR_ACCESS_PORT}/v1/continue"
+export INSPECTOR_EXTRA_ARGS=" ipa-inspection-callback-url=${IRONIC_BASE_URL}:${IRONIC_INSPECTOR_ACCESS_PORT}/v1/continue"
```
elfosardo, line 19: let's leave this untouched and instead export the variable after the if-then block, in L23 for example

scripts/runhttpd
```diff
@@ -22,7 +22,7 @@ chmod 0777 /shared/html
 
 IRONIC_BASE_URL="${IRONIC_SCHEME}://${IRONIC_URL_HOST}"
 
-INSPECTOR_EXTRA_ARGS=" ipa-inspection-callback-url=${IRONIC_BASE_URL}:${IRONIC_INSPECTOR_ACCESS_PORT}/v1/continue"
+export INSPECTOR_EXTRA_ARGS=" ipa-inspection-callback-url=${IRONIC_BASE_URL}:${IRONIC_INSPECTOR_ACCESS_PORT}/v1/continue"
```
elfosardo, line 25: same here, let's leave this untouched and export the variable  after the if-then block in L29

elfosardo (changes requested):

scripts/configure-httpd-ipa.sh
shibaPuppy: @elfosardo 
I don't understand why you have to move to L23.

elfosardo: because we're modifyinf the variable in L21

elfosardo (changes requested):

scripts/configure-httpd-ipa.sh
shibaPuppy: @elfosardo 
ok, i changed it.

shibaPuppy: https://github.com/metal3-io/ironic-image/pull/388/files#diff-ae3e5b5ce83ded18509d59dd21be8cf99921107ef4ab4d571dff31dd76c581caR23

elfosardo (approved):

PR #390, opened 2022-10-24
Allow customizing building from source

elfosardo:
This patch introduces some variables and improvements that can help
customizing the ironic-image when building it from source.
In particular, it adds the option to install ironic, ironic-inspector,
or the sushy library from a gerrit refs or a commit hash.


ironic-source-list.j2
```diff
@@ -0,0 +1,18 @@
+crudini
+gunicorn
+{% if env.IRONIC_SOURCE_HASH %}
+ironic @ git+https://opendev.org/openstack/ironic@{{ env.IRONIC_SOURCE_HASH }}
+{% elif env.IRONIC_SOURCE_GERRIT %}
+ironic @ git+https://opendev.org/openstack/ironic@{{ env.IRONIC_SOURCE_GERRIT }}
```
dtantsur, line 6: Looks identical to the previous case, maybe use the same option?

ironic-source-list.j2
```diff
@@ -0,0 +1,18 @@
...
+{% else %}
+ironic @ git+https://opendev.org/openstack/ironic
+{% endif %}
+ironic-inspector @ git+https://opendev.org/openstack/ironic-inspector
```
dtantsur, line 10: Will likely need the same treatment.

ironic-source-list.j2
```diff
@@ -0,0 +1,18 @@
...
+ironic-prometheus-exporter
+proliantutils
+PyMySQL>=0.8.0
+python-dracclient
+python-ibmcclient
+python-scciclient
+sushy
```
dtantsur, line 17: Ideally here as well

ironic-source-list.j2
elfosardo: yes, I will use the same option
it will be one option for each custom source

ironic-source-list.j2
elfosardo: yeah, in the end I'll try to cover as much packages/libraries as possible

elfosardo:
weird resolution issue for https://raw.githubusercontent.com

PR #393, opened 2022-11-07
add SECURITY_CONTACTS file

Rozzii:


iurygregory:
Do we need test-ubuntu-integration-main for this?
I can apply the lgtm now if we don't need.

Rozzii:
@iurygregory  we don't need the ubuntu integration test, it is failing because of unrelated CI issues. Your lgtm would be much appreciated !

dtantsur:
Force-merging since there is no point to run the CI.

PR #395, opened 2022-11-09
Correctly support IPv6 in DNS_IP and GATEWAY_IP

dtantsur:


ironic-config/dnsmasq.conf.j2
```diff
@@ -16,13 +16,13 @@ dhcp-range={{ env.DHCP_RANGE }}
 # Disable default router(s)
 dhcp-option=3
 {% else %}
-dhcp-option=3,{{ env["GATEWAY_IP"] }}
+dhcp-option=option{% if ":" in env["GATEWAY_IP"] %}6{% endif %}:router,{{ env["GATEWAY_IP"] }}
```
shibaPuppy, line 19: @dtantsur 
https://github.com/metal3-io/ironic-image/pull/395/files#diff-3800b4db123aac6066002eb43c6eaaa0ef456540527d88128cd8d5bacdc5213eR45-R68

It seems to be an available option in "dhcp-option=option6" .
the link above seems to be moved to ipv6.

[dnsmasq man page](https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html) 
```
IPv6 options are specified using the option6: keyword, followed by the option number or option name. The IPv6 option name space is disjoint from the IPv4 option name space. IPv6 addresses in options must be bracketed with square brackets, eg. --dhcp-option=option6:ntp-server,[1234::56] For IPv6, [::] means "the global address of the machine running dnsmasq", whilst [fd00::] is replaced with the ULA, if it exists, and [fe80::] with the link-local address.


```

dtantsur: Sorry, I don't quite get the comment. I'm doing exactly what is described there, no?

shibaPuppy: @dtantsur 
sorry. i look checked it wrong.

PR #396, opened 2022-11-10
Use local directory to build from source

elfosardo:


Dockerfile
```diff
@@ -31,6 +31,10 @@ ARG IRONIC_SOURCE
 ARG IRONIC_INSPECTOR_SOURCE
 ARG SUSHY_SOURCE
 
+COPY sources /sources/
```
dtantsur, line 34: I wonder if this can be an ARG?

ironic-source-list
```diff
@@ -1,12 +1,20 @@
 crudini
 gunicorn
 {% if env.IRONIC_SOURCE %}
+    {% if path.isdir('/sources/' + env.IRONIC_SOURCE) %}
```
dtantsur, line 4: Can we just assume `/sources/ironic` in this case to avoid setting a dummy variable?

Dockerfile
elfosardo: I'd rather keep all the sources under a standard path
it needs to exist (otherwise the copy fails) and it needs to be under the same context as the Dockerfile, so I don't see a lot of margin for customization

ironic-source-list
elfosardo: I'm not sure I follow, IRONIC_SOURCE needs to be set to allow customizing the source from a local dir or a remote location
if we assume /sources/ironic do we want to force the local ironic source dir to always be 'ironic' ?

Dockerfile
dtantsur: Same context, right.. This is unfortunate, but not too bad.

ironic-source-list
dtantsur: > if we assume /sources/ironic do we want to force the local ironic source dir to always be 'ironic' ?

Yeah, that's what I meant. I don't think we're likely to have anything else :) but it's not a blocking issue.

ironic-source-list
```diff
@@ -1,12 +1,20 @@
...
+git+file:///sources/{{ env.IRONIC_SOURCE }}
+    {% else %}
 ironic @ git+https://opendev.org/openstack/ironic@{{ env.IRONIC_SOURCE }}
+    {% endif %}
 {% else %}
 ironic @ git+https://opendev.org/openstack/ironic
 {% endif %}
 {% if env.IRONIC_INSPECTOR_SOURCE %}
+    {% if if path.isdir('/sources/' + env.IRONIC_INSPECTOR_SOURCE) %}
```
furkatgofurov7, line 13: Am I seeing double `if` and that is on purpose?

ironic-source-list
```diff
@@ -17,7 +25,11 @@ python-dracclient
 python-ibmcclient
 python-scciclient
 {% if env.SUSHY_SOURCE %}
+    {% if if path.isdir('/sources/' + env.SUSHY_SOURCE) %}
```
furkatgofurov7, line 28: same

furkatgofurov7:
Oh, one more thing I forgot to mention, I assume https://metal3io.netlify.app/ironic/ironic-container-images.html needs an updated as well?

ironic-source-list
elfosardo: oh snap, thanks for spotting this!

elfosardo:
> Oh, one more thing I forgot to mention, I assume https://metal3io.netlify.app/ironic/ironic-container-images.html needs an update after this?

@furkatgofurov7 yes, I was planning to update that in one PR once this was merged
I will add the change as a WIP for the time being

ironic-source-list
elfosardo: this is disturbing though, I tested the change and it works even with double if :/
anyway, I'll fix it

furkatgofurov7: strange, but anyways IMO it is safer if we fix it right away. Thanks

furkatgofurov7:
> > Oh, one more thing I forgot to mention, I assume https://metal3io.netlify.app/ironic/ironic-container-images.html needs an update after this?
> 
> @furkatgofurov7 yes, I was planning to update that in one PR once this was merged I will add the change as a WIP for the time being

😀 no hurry just wanted to remind it, thank you @elfosardo

furkatgofurov7:
There is an issue (VM creation being stuck) we are seeing recently with the CI recently, and we have a support ticket open, meanwhile, re-triggering should help. 

/test-ubuntu-integration-main

furkatgofurov7 (approved):
/lgtm

PR #397, opened 2022-11-23
Fix source build if building sushy from source

elfosardo:
The sushy library is present in upper-consraints blocking the installation of any newer version (like dev versions). When installing sushy from source we need to remove it from the upper-constraints file.

### DIFF
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -32,6 +32,12 @@ if [[ $INSTALL_TYPE == "source" ]]; then
 
     python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ, path=os.path))' < "${IRONIC_PKG_LIST}" > "${IRONIC_PKG_LIST_FINAL}"
 
+    if [ -n $SUSHY_SOURCE ]; then
+        curl -L ${UPPER_CONSTRAINTS_FILE:-"https://releases.openstack.org/constraints/upper/master"} -o /tmp/sushy-constraints.txt
+        UPPER_CONSTRAINTS_FILE="/tmp/sushy-constraints.txt"
+        sed -i '/^sushy===/d' $UPPER_CONSTRAINTS_FILE
+    fi
+
     python3 -m pip install --ignore-installed --prefix /usr -r $IRONIC_PKG_LIST_FINAL -c ${UPPER_CONSTRAINTS_FILE:-"https://releases.openstack.org/constraints/upper/master"}
 
     # ironic and ironic-inspector system configuration


prepare-image.sh
```diff
@@ -32,6 +32,12 @@ if [[ $INSTALL_TYPE == "source" ]]; then
 
     python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ, path=os.path))' < "${IRONIC_PKG_LIST}" > "${IRONIC_PKG_LIST_FINAL}"
 
+    if [ -n $SUSHY_SOURCE ]; then
+        UPPER_CONSTRAINTS_FILE="/tmp/sushy-constraints.txt"
+        curl -L https://releases.openstack.org/constraints/upper/master -o $UPPER_CONSTRAINTS_FILE
```
dtantsur, line 37: The existing code accounts for an UPPER_CONSTRAINTS_FILE provided by a user. Maybe do it like this:

```bash
curl -L ${UPPER_CONSTRAINTS_FILE:-"https://releases.openstack.org/constraints/upper/master"} -o /tmp/sushy-constraints.txt
UPPER_CONSTRAINTS_FILE="/tmp/sushy-constraints.txt"
```

prepare-image.sh
```diff
@@ -32,6 +32,12 @@ if [[ $INSTALL_TYPE == "source" ]]; then
...
+        sed -i '/sushy===/d' $UPPER_CONSTRAINTS_FILE
```
dtantsur, line 38: I'd use `/^sushy===/d` to be really sure we're not touching any other packages ending with "sushy"

prepare-image.sh
elfosardo: yeah, looks like a good solution for that, going to update the change
thanks!

elfosardo:
/test-ubuntu-integration-main
timeout on cs9 repo

iurygregory (approved):

prepare-image.sh
```diff
@@ -32,6 +32,12 @@ if [[ $INSTALL_TYPE == "source" ]]; then
 
     python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ, path=os.path))' < "${IRONIC_PKG_LIST}" > "${IRONIC_PKG_LIST_FINAL}"
 
+    if [ -n $SUSHY_SOURCE ]; then
+        curl -L ${UPPER_CONSTRAINTS_FILE:-"https://releases.openstack.org/constraints/upper/master"} -o /tmp/sushy-constraints.txt
+        UPPER_CONSTRAINTS_FILE="/tmp/sushy-constraints.txt"
```
derekhiggins, line 37: If you put this on line up you could use it twice

dtantsur: Not really, then user-provided UPPER_CONSTRAINTS_FILE won't work.

derekhiggins: ack, I missed that

PR #370, opened 2022-04-19
Avoid unbound variables and other surprises in ironic-common

elfosardo:

PR #398, opened 2022-11-25
🐛  Fix centos httpd

mboukhalfa:
httpd-infra is not running correctly on centos because of `PROVISIONING_MACS unbound value` after inroducing bash `set -u` in https://github.com/metal3-io/ironic-image/pull/370


kashifest:
@bfournie,@derekhiggins, @dtantsur, @iurygregory we need lgtm here, the CI is stuck because of this.

furkatgofurov7 (commented):
LGTM

lentzi90 (approved):

PR #400, opened 2022-11-28
Dockerfiles: use apt-get instead of apt

tuminoid:
Apt gives out warnings that it doesn't have stable CLI for script usage. Replace apt with apt-get in Dockerfiles.

Also, it is a best practise to define DEBIAN_FRONTEND=noninteractive so the build will not ask you questions, but accepts the defaults, in case it needs any confirmation from user. Setting this avoids stalling builds in case there is a conflict, or configuration option added to the installed packages.


lentzi90 (approved):

kashifest (commented):
lgtm

furkatgofurov7 (approved):

PR #404, opened 2022-12-27
Use sushytools port from the config file

mboukhalfa:
Sushy tools port is hard-coded in the Dockerfile this PR removes it since the default port is 8000 and allows parameterize the port from the config file e.g `SUSHY_EMULATOR_LISTEN_PORT = 8000`


Rozzii:
/ test-ubuntu-integration-main

mboukhalfa:
/assign @derekhiggins
cc @elfosardo @dtantsur

PR #409, opened 2023-03-02
set docker registry url to docker.io

tuminoid:
This is the last instance of registry.hub.docker.com to align in main branches of actively developed Metal3 repositories.


lentzi90 (commented):
/lgtm

PR #410, opened 2023-03-10
make ironic-image runnable as non-root

tuminoid:
BMO ironic has no reason to run as root. Make it run as "ironic" user.
    
dnsmasq requires elevated capabiities. k8s is missing the feature of ambient capabilities, so it requires us to setcap the binaries with expected capabilities and container must be running with "allowPrivilegeEscalation: true" in the manifest to allow elevation.
    
Read the ambient capabilities KEP for more details: https://github.com/kubernetes/enhancements/blob/master/keps/sig-security/2763-ambient-capabilities/README.md
    
Most of the changed here are making runtime configuration possible by changing file and directory ownership from root:root to root:ironic and applying 775/664 perms to allow modification/creation of files.

### DIFF
--- a/configure-nonroot.sh
+++ b/configure-nonroot.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/bash
+
+# This script changes permissions to allow Ironic container to run as non-root
+# user. As the same image is used to run ironic, ironic-httpd, ironic-dsnmasq,
+# ironic-inspector and ironic-log-watch via BMO's ironic k8s manifest, it has
+# to be configured to work with multiple different users and groups, while they
+# share files via bind mounts (/shared, /certs/*), which can only get one
+# group id as "fsGroup". Additionally, dnsmasq needs three capabilities to run
+# which we provide via "setcap", and "allowPrivilegeEscalation: true" in
+# manifest.
+
+set -eux
+
+# user and group are from ironic rpms (uid 997, gid 994)
+IRONIC_USER="ironic"
+IRONIC_GROUP="ironic"
+INSPECTOR_GROUP="ironic-inspector"
+
+# we'll bind mount shared ca and ironic/inspector certificate dirs here
+# that need to have correct ownership as the entire ironic in BMO
+# deployment shares a single fsGroup in manifest's securityContext
+mkdir -p /certs
+chown "${IRONIC_USER}":"${INSPECTOR_GROUP}" /certs
+chmod 2775 /certs
+
+# ironic, inspector and httpd related changes
+chown -R root:"${IRONIC_GROUP}" /etc/ironic /etc/httpd/conf /etc/httpd/conf.d
+chown -R "${IRONIC_USER}":"${INSPECTOR_GROUP}" /etc/ironic-inspector
+chmod 2775 /etc/ironic /etc/ironic-inspector /etc/httpd/conf /etc/httpd/conf.d
+chmod 664 /etc/ironic/* /etc/ironic-inspector/* /etc/httpd/conf/* /etc/httpd/conf.d/*
+
+chown -R root:"${IRONIC_GROUP}" /var/lib/ironic
+chown -R root:"${INSPECTOR_GROUP}" /var/lib/ironic-inspector
+chmod 2775 /var/lib/ironic /var/lib/ironic-inspector
+chmod 664 /var/lib/ironic/ironic.db /var/lib/ironic-inspector/ironic-inspector.db
+
+# dnsmasq, and the capabilities required to run it as non-root user
+chown -R root:"${IRONIC_GROUP}" /etc/dnsmasq.conf /var/lib/dnsmasq
+chmod 2775 /var/lib/dnsmasq
+touch /var/lib/dnsmasq/dnsmasq.leases
+chmod 664 /etc/dnsmasq.conf /var/lib/dnsmasq/dnsmasq.leases
+
+setcap "cap_net_raw,cap_net_admin,cap_net_bind_service=+eip" /usr/sbin/dnsmasq
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -21,6 +21,11 @@ fi
 
 # SOURCE install #
 if [[ $INSTALL_TYPE == "source" ]]; then
+    # emulate uid/gid configuration to match rpm install
+    IRONIC_UID=997
+    IRONIC_GID=994
+    INSPECTOR_UID=996
+    INSPECTOR_GID=993
     BUILD_DEPS="python3-devel gcc git-core python3-setuptools python3-jinja2"
     dnf upgrade -y
     # NOTE(dtantsur): pip is a requirement of python3 in CentOS
@@ -32,7 +37,7 @@ if [[ $INSTALL_TYPE == "source" ]]; then
 
     python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ, path=os.path))' < "${IRONIC_PKG_LIST}" > "${IRONIC_PKG_LIST_FINAL}"
 
-    if [ -n $SUSHY_SOURCE ]; then
+    if [[ -n $SUSHY_SOURCE ]]; then
         curl -L ${UPPER_CONSTRAINTS_FILE:-"https://releases.openstack.org/constraints/upper/master"} -o /tmp/sushy-constraints.txt
         UPPER_CONSTRAINTS_FILE="/tmp/sushy-constraints.txt"
         sed -i '/^sushy===/d' $UPPER_CONSTRAINTS_FILE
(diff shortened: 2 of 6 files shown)


tuminoid:
/test-ubuntu-integration-main 

To get any benefit from this change, the BMO PR need to be applied, but as standalone, this should be non-invasive and should pass testing. So let's run it for smoke test.

tuminoid:
Yeah, it passes smoke, nice. Now to figure out the TODOs and how to test in CI.

tuminoid:
/cc @elfosardo @dtantsur 

This is still WIP, but I'd appreciate early comments.

tuminoid:
> @tuminoid thanks, two things for now:
> 
> * please run centos job as well
> * let's not add too many things to Dockerfile, I'm pretty sure all the changes odne there can be moved to existing scripts

The things I’ve changed in/to Dockerfile are there because they need root privileges, and during build we are root. When were running the container, we are nonroot and cannot do any of them.

elfosardo:
> > @tuminoid thanks, two things for now:
> > 
> > * please run centos job as well
> > * let's not add too many things to Dockerfile, I'm pretty sure all the changes odne there can be moved to existing scripts
> 
> The things I’ve changed in/to Dockerfile are there because they need root privileges, and during build we are root. When were running the container, we are nonroot and cannot do any of them.

@tuminoid we run scripts in the Dockerfile to avoid that being too long and create too many layers
please follow the same example

tuminoid:
> 

OK, now I got you. Sure, this is just POC at this point, I'll put all the non-root related configuration to single script when I'm done with the FIXME's and TODO's.

tuminoid:
/test-centos-integration-main
/test-ubuntu-integration-main 

Combined tests are running at https://github.com/metal3-io/metal3-dev-env/pull/1172 but this should pass standalone as well.

configure-nonroot.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/bash
+
+set -eux
+
+NONROOT_USER="nonroot"
+NONROOT_GROUP="nonroot"
```
dtantsur, line 6: I'm quite sure the RPMs already set up an "ironic" user and group, why not use them? "nonroot" looks confusing to me.

scripts/rundnsmasq
```diff
@@ -20,10 +25,14 @@ mkdir -p /shared/html/pxelinux.cfg
...
-python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ))' </etc/dnsmasq.conf.j2 >/etc/dnsmasq.conf
+# we template via /tmp as sed otherwise creates temp files in /etc directory
+# where we can't write
+python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ))' </etc/dnsmasq.conf.j2 >/tmp/dnsmasq.conf
 
 for iface in $( echo "$DNSMASQ_EXCEPT_INTERFACE" | tr ',' ' '); do
-    sed -i -e "/^interface=.*/ a\except-interface=${iface}" /etc/dnsmasq.conf
+    sed -i -e "/^interface=.*/ a\except-interface=${iface}" /tmp/dnsmasq.conf
 done
+cat /tmp/dnsmasq.conf > /etc/dnsmasq.conf
```
dtantsur, line 35: nit: why not mv?

dtantsur:
Tuomo, this is a great idea, but I wonder if it would be easier to exclude dnsmasq initially. It's pretty natural for it to run as root.

/cc @zaneb 

Zane, I know you were interested in such a thing.

configure-nonroot.sh
tuminoid: `nonroot` is standard user in k8s rootless. I can change to `ironic`, no problem.

prepare-image.sh
```diff
@@ -74,11 +74,15 @@ rm -f /etc/httpd/conf.d/ssl.conf /etc/httpd/conf.d/autoindex.conf /etc/httpd/con
 # RDO-provided configuration forces creating log files
 rm -f /usr/share/ironic/ironic-dist.conf /etc/ironic-inspector/inspector-dist.conf
 
+# add ironic and ironic-inspector to apache group
+usermod -aG ironic apache
+usermod -aG ironic-inspector apache
```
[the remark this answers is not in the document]
tuminoid: These are just moved from runtime scripts to build time scripts, no change in functionality. Apache group is already configured for some of the directories for httpd.

scripts/runhttpd
```diff
@@ -78,6 +74,12 @@ sed -i -e 's|\(^[[:space:]]*\)\(DocumentRoot\)\(.*\)|\1\2 "/shared/html"|' \
 sed -i -e 's%^ \+CustomLog.*%    CustomLog /dev/stderr combined%g' /etc/httpd/conf/httpd.conf
 sed -i -e 's%^ErrorLog.*%ErrorLog /dev/stderr%g' /etc/httpd/conf/httpd.conf
 
+# FIXME: put pidfile to somewhere else than /tmp, originally at /run/httpd.pid
+# but /run is not permanent dir so can't fix perms for that
```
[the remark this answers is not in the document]
tuminoid: This is still a FIXME so I have not decided on the directory yet, but `/var/tmp` is a candidate.

tuminoid:
/test-centos-integration-main
/test-ubuntu-integration-main

Fixed one left-over comment, otherwise the same.

@dtantsur @zaneb I've addressed all comments. PTAL.

tuminoid:
> @tuminoid can we also run the source CI job please? it looks like it might conflict with the user/group configuration we use there

1. I don't know what exactly you mean by "source CI job"? Please trigger it as you see necessary.
2. I'm using pre-existing user from Ironic RPM, just as you @dtantsur asked, and not creating any new ones. How can it conflict? Can you point to where it conflicts?

elfosardo:
@tuminoid the ironic user/group is valid for the ironic processes and configuration, but ironic-inspector uses different user/group, that's where the conflict is
for the  source configuration you can check the prepare-image.sh script at L22 to L52

tuminoid:
> @tuminoid the ironic user/group is valid for the ironic processes and configuration, but ironic-inspector uses different user/group, that's where the conflict is for the source configuration you can check the prepare-image.sh script at L22 to L52

OK, I made the users/groups match the rpm install image. I also changed inspector related files to inspector, and on the BMO side made the inspector container run as inspector.

/test-ubuntu-integration-main-ironic-source
/test-centos-integration-main

elfosardo:
@tuminoid thanks!
I beleive the name of the source CI job changed and I missed it!
I'll ask during the meeting today

tuminoid:
> @tuminoid thanks! I beleive the name of the source CI job changed and I missed it! I'll ask during the meeting today

I found out that they're the regular job names plus `-ironic-source` postfix.

/test-ubuntu-integration-main-ironic-source
/test-centos-integration-main

furkatgofurov7:
> test-ubuntu-integration-main-ironic-source

this is correct, even though you run ^ , jenkins would not show it with that name in the list of running jobs, rather with usual name `test-ubuntu-integration-main` but taking into account the postfix specified: https://jenkins.nordix.org/view/Metal3/job/metal3_ironic_image_main_integration_test_ubuntu_ironic_from_source/

configure-nonroot.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -eux
+
+# user and group are from ironic rpms (uid 997, gid 994)
+IRONIC_USER="ironic"
+IRONIC_GROUP="ironic"
+INSPECTOR_GROUP="ironic-inspector"
+
+# certificates (must be writable by both ironic and ironic-inspector)
+mkdir -p /certs
```
zaneb, line 11: IIRC this is somewhere that we expect to bind-mount something in to the container. Does creating a directory in the same location and setting the permissions actually have the right/any effect?

zaneb (commented):
This is great, thank you :+1:

configure-nonroot.sh
tuminoid: Thanks for the comment @zaneb !

Bind mounts are actually done for sub-directories under `/certs`, not  `/certs` itself. We are then creating directories in `tls-common.sh` under `/certs` to those bind mount locations, and those mkdir's are failing without these lines.

Two ways of handling this are:

- I can either keep-as is, or 
- Drop the two lines in `configure-nonroot.sh` and also drop L20-24 in https://github.com/metal3-io/ironic-image/blob/ba6890c7b69ef16ed3f362a707a0fdd48a07e6a8/scripts/tls-common.sh#L20 (the failing mkdir's). 

Latter seems to work, but the resulting directory ownerships are bit different. In `configure-nonroot.sh` we assign them to `ironic:ironic-inspector`, but the bind mount created directories and files will be `root:ironic` per the `fsGroup: ...` in the deployment manifest. The sub-directories under `/certs` are only created by the bind mounts, I don't know if that would be an issue in some use-case where mounting does not occur?

I would be more comfortable with the current implementation, but what do you think?

zaneb: Ah, OK, it sounds like we're creating dirs in there so that they exist whether or not somebody actually bind-mounts something at that point. Presumably that is required to not break something else.
I'm comfortable with this implementation. Thanks for the explanation! (I wouldn't be sad if some of this made it's way into a code comment.)

tuminoid: I have put much of the explanation into code comments.

tuminoid:
OK, tests are passing here as well as https://github.com/metal3-io/metal3-dev-env/pull/1172, and all the comments are addressed.

Can I get LGTM and approval? @zaneb @dtantsur @elfosardo

PR #412, opened 2023-04-05
add shellcheck and fix complaints

tuminoid:
Add shellcheck. Fix all the shellcheck complaints, and make shell code consistent Bash.
    
Another PR in project-infra will activate the check for PRs.

Markdownlint is separated to its own PR #413

### DIFF
--- a/prepare-efi.sh
+++ b/prepare-efi.sh
@@ -2,15 +2,15 @@
 
 set -euxo pipefail
 
-ARCH=$(uname -m)
-DEST=${2:-/tmp/esp.img}
 OS=${1:-centos}
+DEST=${2:-/tmp/esp.img}
+ARCH=$(uname -m)
 
-if [ "$ARCH" = "x86_64" ]; then
+if [[ "$ARCH" == "x86_64" ]]; then
     PACKAGES=grub2-efi-x64
     BOOTEFI=BOOTX64.EFI
     GRUBEFI=grubx64.efi
-elif [ "$ARCH" = "aarch64" ]; then
+elif [[ "$ARCH" == "aarch64" ]]; then
     PACKAGES=grub2-efi-aa64
     BOOTEFI=BOOTAA64.EFI
     GRUBEFI=grubaa64.efi
@@ -28,17 +28,18 @@ fi
 # ``Error converting to codepage 850 Invalid argument``
 # ``Cannot initialize '::'``
 # This is due to the conversion table missing codepage 850, included in glibc-gconv-extra
+# shellcheck disable=SC2086
 dnf install -y grub2 shim dosfstools mtools glibc-gconv-extra $PACKAGES
 
 ## TODO(TheJulia): At some point we may want to try and make the size
 ## of the ESP image file to be sized smaller for the files that need to
 ## be copied in, however that requires more advanced scripting beyond
 ## an MVP.
-dd bs=1024 count=6400 if=/dev/zero of=$DEST
-mkfs.msdos -F 12 -n 'ESP_IMAGE' $DEST
+dd bs=1024 count=6400 if=/dev/zero of="$DEST"
+mkfs.msdos -F 12 -n 'ESP_IMAGE' "$DEST"
 
-mmd -i $DEST EFI
-mmd -i $DEST EFI/BOOT
-mcopy -i $DEST -v /boot/efi/EFI/BOOT/$BOOTEFI ::EFI/BOOT
-mcopy -i $DEST -v /boot/efi/EFI/$OS/$GRUBEFI ::EFI/BOOT
-mdir -i $DEST ::EFI/BOOT;
+mmd -i "$DEST" EFI
+mmd -i "$DEST" EFI/BOOT
+mcopy -i "$DEST" -v "/boot/efi/EFI/BOOT/$BOOTEFI" ::EFI/BOOT
+mcopy -i "$DEST" -v "/boot/efi/EFI/$OS/$GRUBEFI" ::EFI/BOOT
+mdir -i "$DEST" ::EFI/BOOT
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -2,7 +2,7 @@
 
 set -euxo pipefail
 
-IRONIC_PKG_LIST=/tmp/ironic-${INSTALL_TYPE}-list
+IRONIC_PKG_LIST="/tmp/ironic-${INSTALL_TYPE}-list"
 
 echo "install_weak_deps=False" >> /etc/dnf/dnf.conf
 # Tell RPM to skip installing documentation
@@ -11,16 +11,16 @@ echo "tsflags=nodocs" >> /etc/dnf/dnf.conf
 dnf install -y python3 python3-requests 'dnf-command(config-manager)'
 
 # RPM install #
-if [[ $INSTALL_TYPE == "rpm" ]]; then
+if [[ "$INSTALL_TYPE" == "rpm" ]]; then
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/plugins/module_utils/tripleo_repos/main.py | python3 - -b master current-tripleo
     # NOTE(elfosardo): enable CRB repo for more python3 dependencies
     dnf config-manager --set-enabled crb
     dnf upgrade -y
-    xargs -rtd'\n' dnf install -y < $IRONIC_PKG_LIST
+    xargs -rtd'\n' dnf install -y < "$IRONIC_PKG_LIST"
 fi
 
 # SOURCE install #
-if [[ $INSTALL_TYPE == "source" ]]; then
+if [[ "$INSTALL_TYPE" == "source" ]]; then
     # emulate uid/gid configuration to match rpm install
     IRONIC_UID=997
     IRONIC_GID=994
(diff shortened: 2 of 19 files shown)


kashifest (dismissed):

prepare-efi.sh
```diff
@@ -28,17 +28,17 @@ fi
 # ``Error converting to codepage 850 Invalid argument``
 # ``Cannot initialize '::'``
 # This is due to the conversion table missing codepage 850, included in glibc-gconv-extra
-dnf install -y grub2 shim dosfstools mtools glibc-gconv-extra $PACKAGES
+dnf install -y grub2 shim dosfstools mtools glibc-gconv-extra "${PACKAGES}"
```
elfosardo, line 31: maybe add disable=SC2086 here?
the PACKAGES list will be considered as a single string, afraid this will break the build

prepare-image.sh
```diff
@@ -11,16 +11,16 @@ echo "tsflags=nodocs" >> /etc/dnf/dnf.conf
...
 
 # RPM install #
-if [[ $INSTALL_TYPE == "rpm" ]]; then
+if [[ "${INSTALL_TYPE}" == "rpm" ]]; then
     curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/plugins/module_utils/tripleo_repos/main.py | python3 - -b master current-tripleo
     # NOTE(elfosardo): enable CRB repo for more python3 dependencies
     dnf config-manager --set-enabled crb
     dnf upgrade -y
-    xargs -rtd'\n' dnf install -y < $IRONIC_PKG_LIST
+    xargs -rtd'\n' dnf install -y < "${IRONIC_PKG_LIST}"
```
elfosardo, line 19: same here, probably better add disable=SC2086

prepare-image.sh
```diff
@@ -29,38 +29,37 @@ if [[ $INSTALL_TYPE == "source" ]]; then
     BUILD_DEPS="python3-devel gcc git-core python3-setuptools python3-jinja2"
     dnf upgrade -y
     # NOTE(dtantsur): pip is a requirement of python3 in CentOS
-    dnf install -y python3-pip $BUILD_DEPS
+    dnf install -y python3-pip "${BUILD_DEPS}"
```
elfosardo, line 32: ditto

prepare-image.sh
```diff
@@ -29,38 +29,37 @@ if [[ $INSTALL_TYPE == "source" ]]; then
...
-    getent group ironic-inspector >/dev/null || groupadd -r ironic-inspector -g "${INSPECTOR_GID}"
-    getent passwd ironic-inspector >/dev/null || useradd -r -g ironic-inspector -u "${INSPECTOR_UID}" -s /sbin/nologin ironic-inspector -d /var/lib/ironic-inspector
+    getent group ironic > /dev/null || groupadd -r ironic -g "${IRONIC_GID}"
+    getent passwd ironic > /dev/null || useradd -r -g ironic -u "${IRONIC_UID}" -s /sbin/nologin ironic -d /var/lib/ironic
+    getent group ironic-inspector > /dev/null || groupadd -r ironic-inspector -g "${INSPECTOR_GID}"
+    getent passwd ironic-inspector > /dev/null || useradd -r -g ironic-inspector -u "${INSPECTOR_UID}" -s /sbin/nologin ironic-inspector -d /var/lib/ironic-inspector
 
     # clean installed build dependencies
-    dnf remove -y $BUILD_DEPS
+    dnf remove -y "${BUILD_DEPS}"
```
elfosardo, line 55: ditto

elfosardo (changes requested):

prepare-image.sh
tuminoid: IRONIC_PKG_LIST is a file name and its read by `<`, so here quotes are correct.

prepare-efi.sh
tuminoid: PACKAGES is defined above to be a single word `grub2-efi-x86`, so it will not break. Variable naming is misleading.

prepare-image.sh
tuminoid: I'll fix that.

dtantsur (changes requested):
I absolutely like the markdown verification, but the diff in the scripts seems mostly useless. There is no point to enforce ${variables} over $variables except that in a few cases.

Landing this will also turn our life downstream into a nightmare, because no backports will apply cleanly (if you do backporting, you'll have the same problem too).

scripts/configure-httpd-ipa.sh
```diff
@@ -1,23 +1,24 @@
...
-if [ -f "$IRONIC_CERT_FILE" ]; then
+if [ -f "${IRONIC_CERT_FILE}" ]; then
     IRONIC_BASE_URL="https://${IRONIC_URL_HOST}"
 else
     IRONIC_BASE_URL="http://${IRONIC_URL_HOST}"
 fi
 
 INSPECTOR_EXTRA_ARGS=" ipa-inspection-callback-url=${IRONIC_BASE_URL}:${IRONIC_INSPECTOR_ACCESS_PORT}/v1/continue"
-if [[ $IRONIC_FAST_TRACK == true ]]; then
+if [[ $IRONIC_FAST_TRACK = true ]]; then
```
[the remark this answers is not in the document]
tuminoid: `==` and `=` are effecively the same thing, except `==` is Bash-only and `=` is POSIX compatible.

That said, the usage of `=` and `==` is  very mixed in the scripts, and I just went consistently for `=`. I can change the other way around, if we end up taking this PR in.

dtantsur:
As a note, the Google style guide is pretty opinionated and may not reflect how most people write scripts. Nor does it reflect the intention of $ vs ${}, which is: ${} is used when the variable name is followed by characters that are valid in a variable name, e.g. `${foo}_bar` (cannot be `$foo_bar`).

tuminoid:
> > Stay consistent with what you find

There is no consistency like I said already, so nothing to stick by.

tuminoid:
/hold

Holding this so it doesn't get merged by drive-by lgtm. @elfosardo and @dtantsur please settle if you are willing to take this in. I'm happy to make adjustments if required, but as stated before, I'm going for consistency, maintainability and readability, preferably organization wide. If this does not work for you, I can drop this, and you can implement it in a way that works for you.

scripts/configure-httpd-ipa.sh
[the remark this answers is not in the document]
tuminoid: Fixed all of the `=` to `==` as requested. Also made all `[ .. ]` to `[[ .. ]]` as `[ ` is POSIX and `==` is Bash-only and their combination is not supported (yes, Bash doesn't explode, but officially its undefined combination).

tuminoid:
@dtantsur @elfosardo Its been a week. Have you guys discussed this?

tuminoid:
> @dtantsur @elfosardo Its been a week. Have you guys discussed this?

It's now been two weeks. Can you guys discuss your mutual stance on this?

The question in the end is rather simple: Do you want to take the PR in or not? We have bunch of shell code that is a bit messy, not linted, not consistent. We have here PR cleaning it up for you, maybe a bit opinionated PR but guess what, most of the languages/ecossytems have rather opinionated best practices/style guides these days (gofmt, prettier, k8s and so on), because its effective. They're easy for anyone to work in, you can automate your editors and your CI.

> I have no visibiility into your internal development practices, and in this repo there are no branches. I'm trying to have all repositories under metal3-io to have consistent coding style, basic quality checks in place etc so they are in the long run easier to maintain, developers have safety nets in place and faster feedback cycle.
> 
> If you feel this change is not worth of merging or causes too much trouble for you, then it is your right to decline it, but I'd hope the opposite, that you possibly take this into your backports, and after that there is no more conflicts. But like I said, I got no idea what your setup looks like and if that is feasible.

This still stands as well.

tuminoid:
Took away all curly brace changes etc. @dtantsur PTAL.

/test-ubuntu-integration-main

scripts/coreos-ipa-common.sh
```diff
@@ -4,18 +4,24 @@ ROOTFS_FILE=${ROOTFS_FILE:-/shared/html/images/ironic-python-agent.rootfs}
...
+    if [[ -f "$ROOTFS_FILE" ]]; then
+        return 0
+    fi
+    return 1
 }
 
 if use_coreos_ipa; then
-    export IRONIC_KERNEL_PARAMS="${IRONIC_KERNEL_PARAMS:-} $(coreos_kernel_params)"
+    IRONIC_KERNEL_PARAMS="${IRONIC_KERNEL_PARAMS:-} $(coreos_kernel_params)"
+    export IRONIC_KERNEL_PARAMS
```
dtantsur, line 26: I wonder why the split is required...

scripts/runhttpd
```diff
@@ -1,16 +1,18 @@
 #!/usr/bin/bash
 
+# shellcheck disable=SC1091
```
dtantsur, line 3: Cannot we follow the recommendation in https://www.shellcheck.net/wiki/SC1090 instead of disabling it?

scripts/runhttpd
```diff
@@ -23,28 +25,29 @@ chmod 0777 /shared/html
 IRONIC_BASE_URL="${IRONIC_SCHEME}://${IRONIC_URL_HOST}"
 
 INSPECTOR_EXTRA_ARGS=" ipa-inspection-callback-url=${IRONIC_BASE_URL}:${IRONIC_INSPECTOR_ACCESS_PORT}/v1/continue"
-if [[ $IRONIC_FAST_TRACK == true ]]; then
+if [[ "$IRONIC_FAST_TRACK" == "true" ]]; then
     INSPECTOR_EXTRA_ARGS+=" ipa-api-url=${IRONIC_BASE_URL}:${IRONIC_ACCESS_PORT}"
 fi
 export INSPECTOR_EXTRA_ARGS
 
+# shellcheck disable=SC1091
```
dtantsur, line 33: same

dtantsur:
Much better, thanks! I'm not a fan of some changes, but I understand the consistency argument here. Just one question inline.

scripts/coreos-ipa-common.sh
tuminoid: `$(coreos_kernel_params)` is subshell call and exporting it on the same line masks the possible error from it.

scripts/configure-ironic.sh
```diff
@@ -11,16 +11,19 @@ IRONIC_EXTERNAL_IP="${IRONIC_EXTERNAL_IP:-}"
 #   <interface.vlan> - a particular VLAN on an interface, not relying on LLDP
 export IRONIC_INSPECTOR_VLAN_INTERFACES=${IRONIC_INSPECTOR_VLAN_INTERFACES:-all}
 
+# shellcheck disable=SC1091
```
elfosardo, line 14: you could use curly brackets to group multiple lines and remove the second disable statement, I'm ok either way

scripts/configure-ironic.sh
```diff
@@ -40,17 +43,18 @@ wait_for_interface_or_ip
 export IRONIC_BASE_URL="${IRONIC_SCHEME}://${IRONIC_URL_HOST}:${IRONIC_ACCESS_PORT}"
 export IRONIC_INSPECTOR_BASE_URL="${IRONIC_INSPECTOR_SCHEME}://${IRONIC_URL_HOST}:${IRONIC_INSPECTOR_ACCESS_PORT}"
 
-if [ -n "$IRONIC_EXTERNAL_IP" ]; then
-    export IRONIC_EXTERNAL_CALLBACK_URL="${IRONIC_SCHEME}://${IRONIC_EXTERNAL_IP}:${IRONIC_ACCESS_PORT}"
-    if [ "$IRONIC_VMEDIA_TLS_SETUP" = "true" ]; then
+if [[ -n "$IRONIC_EXTERNAL_IP" ]]; then
+    IRONIC_EXTERNAL_CALLBACK_URL="${IRONIC_SCHEME}://${IRONIC_EXTERNAL_IP}:${IRONIC_ACCESS_PORT}"
+    export IRONIC_EXTERNAL_CALLBACK_URL
```
elfosardo, line 48: is the split necessary here?

scripts/runhttpd
tuminoid: `disable=SC1091` needs to be there, since shellcheck won't follow `source=xxx` as `/bin/xxx` is considered external source by shellcheck. We could use `SC1090` recommendation, if it was pointing to non-absolute path.

scripts/runhttpd
tuminoid: Same as above.

scripts/configure-ironic.sh
tuminoid: True. IMO it is easier to understand this way, while being explicit about the ignores.

scripts/configure-ironic.sh
tuminoid: No its not, good catch. Fixed.

tuminoid:
I took some more unnecessary changes off, and I think I addressed all comments.

/test-centos-integration-main-ironic-source
/test-ubuntu-integration-main

tuminoid:
Cleura is busted but one more try for today.

/test-centos-integration-main-ironic-source
/test-ubuntu-integration-main

tuminoid:
/cc @dtantsur 

Good for LGTM now?

lentzi90 (commented):
/lgtm

tuminoid:
@dtantsur can you PTAL and remove changes requested so we can merge this? All comments should be addressed now.

dtantsur (approved):

PR #413, opened 2023-04-25
add markdown linter and fix complaints

tuminoid:
Add markdownlint. Fix all the markdownlint complaints.

Another PR in project-infra will activate the check for PRs.

This is markdownlint part separated from #412


dtantsur:
/approve

Thanks!

PR #416, opened 2023-05-23
Add customization, which are needed for multi-conductor scenario

mquhuy:
As titled. In details:
- `IRONIC_BASE_URL` and `IRONIC_INSPECTOR_URL`, up until now, are both inferred from `IRONIC_URL_HOST`, which rendered them inseparable (we cannot config ironic to work with an inspector on another interface). In multi-ironic scenario, there's only 1 inspector and multiple ironics, so every ironic should be able to point to the same inspector, regardless of its ip.
- Adding `IRONIC_URL_HOST` to ironic and inspector httpd config is needed so that we can config multiple httpd containers, each work on its own interface with its own ironic. Having httpd only listening on 1 interface and not all, is actually the expected behavior in case `LISTEN_ALL_INTERFACES` is not True, regardless of the multi-ironic requirements.


ironic-config/apache2-ironic-api.conf.j2
```diff
@@ -10,7 +10,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-Listen {{ env.IRONIC_LISTEN_PORT }}
+Listen {{ env.IRONIC_URL_HOST }}:{{ env.IRONIC_LISTEN_PORT }}
```
dtantsur, line 13: I think you need to respect LISTEN_ALL_INTERFACES here

mquhuy: Thanks, that's a good point. I was under the impression that it was not important as the previous setting didn't care about that. Making change

ironic-inspector-config/inspector-apache.conf.j2
```diff
@@ -11,7 +11,7 @@
 # under the License.
 
 
-Listen {{ env.IRONIC_INSPECTOR_LISTEN_PORT }}
+Listen {{ env.IRONIC_URL_HOST }}:{{ env.IRONIC_INSPECTOR_LISTEN_PORT }}
```
dtantsur, line 14: Same here

mboukhalfa:
> _ No description provided. _

Looks good to me, I would like to have brief description about the need for those changes:
listening on `IRONIC_URL_HOST` and defaulting the other vars like `IRONIC_BASE_URL`.

cc @elfosardo

elfosardo:
> > _ No description provided. _
> 
> Looks good to me, I would like to have brief description and context about the need for those changes: listening on `IRONIC_URL_HOST` and defaulting the other vars like `IRONIC_BASE_URL`.
> 
> cc @elfosardo

I agree, either we do it here or in a follow up
@mquhuy up to you

mquhuy:
> > _ No description provided. _
> 
> Looks good to me, I would like to have brief description and context about the need for those changes: listening on `IRONIC_URL_HOST` and defaulting the other vars like `IRONIC_BASE_URL`.
> 
> cc @elfosardo



> > > _ No description provided. _
> > 
> > 
> > Looks good to me, I would like to have brief description and context about the need for those changes: listening on `IRONIC_URL_HOST` and defaulting the other vars like `IRONIC_BASE_URL`.
> > cc @elfosardo
> 
> I agree, either we do it here or in a follow up @mquhuy up to you

Thanks guys, I added the description with some explanation. Please tell me if it's not clear :D

dtantsur:
You may be also interested in https://github.com/metal3-io/ironic-image/pull/421

mquhuy:
> You may be also interested in #421

Thank you. I will note that.

mquhuy:
@elfosardo Could you please check if we can merge this now?

PR #420, opened 2023-06-09
Fix dnf clean command position

elfosardo:
The dnf clean and data removal should be the last command used to be sure to remove all the content.


elfosardo:
cc @iurygregory @lentzi90 @Rozzii

lentzi90 (commented):
/lgtm

PR #423, opened 2023-06-22
Stop using extra-hardware by default

dtantsur:
After https://github.com/metal3-io/baremetal-operator/pull/1297,
it is no longer needed. Upstream ramdisks do not support it.


elfosardo:
/approve
/hold
BMO change should merge soon

PR #427, opened 2023-07-17
Start pinning ironic with commit hash

elfosardo:


elfosardo:
/hold
we need to discuss about this first

elfosardo:
we've decided to go in a different direction

PR #428, opened 2023-08-04
reorganize the internal structure

Rozzii:
This commit:
  - provides a more strict access controll for the httpd ditectories
  - blocks access to the ftp directories via http paths
  - removes unused legacy code (mainly related to ignition and coreOS)
  - where possible 'sed' commands have been removed and replaced with jinja templates
  - httpd configuration has been modifed to better utilize the modular nature of apache configuration files
  - environment variable controlled switch has been introduced to the allow/block serving /shared/html/images when "runhttpd" is in use
  
  This PR is also a preparatory work before image-server basic auth and iPXE hardening is introduced.


Dockerfile
```diff
@@ -46,13 +46,16 @@ COPY --from=ironic-builder /tmp/ipxe/src/bin/undionly.kpxe /tmp/ipxe/src/bin-x86
...
+COPY ironic-config/apache2-ironic-api.conf.j2 /etc/httpd-ironic-api.conf.j2
+
+# DNSMASQ
 COPY ironic-config/dnsmasq.conf.j2 /etc/
-COPY ironic-config/inspector.ipxe.j2 ironic-config/ironic-python-agent.ign.j2 /tmp/
 
 # Custom httpd config, removes all but the bare minimum needed modules
-COPY ironic-config/httpd.conf /etc/httpd/conf.d/
+COPY ironic-config/httpd.conf.j2 /etc/httpd/conf/
+RUN mv /etc/httpd/conf/httpd.conf /etc/httpd/conf/httpd.conf.example
```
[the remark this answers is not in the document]
Rozzii: I thought to keep the example for folks who might be working on the repo and as reference. I am not opinionated in this regard I can get rid of this if the RUN is too expensive to run in some build envs.

Dockerfile
```diff
@@ -46,13 +46,16 @@ COPY --from=ironic-builder /tmp/ipxe/src/bin/undionly.kpxe /tmp/ipxe/src/bin-x86
 COPY --from=ironic-builder /tmp/esp.img /tmp/uefi_esp.img
 
 COPY ironic-config/ironic.conf.j2 /etc/ironic/
+COPY ironic-config/inspector.ipxe.j2 /tmp/
+COPY ironic-config/apache2-ironic-api.conf.j2 /etc/httpd-ironic-api.conf.j2
```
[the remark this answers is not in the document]
Rozzii: sure, I will change it then

Dockerfile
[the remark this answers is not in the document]
elfosardo: let's meet in the middle :)
maybe we can remove this RUN command and instead add the mv command in the scripts/runhttpd file, before we run the jinja command, for example at L41 ?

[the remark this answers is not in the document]
Rozzii: Sure, I will do so

README.md
```diff
@@ -52,6 +52,9 @@ functionality:
 - `DNSMASQ_EXCEPT_INTERFACE` - interfaces to exclude when providing DHCP address
   (default `lo`)
 - `HTTP_PORT` - port used by http server (default `80`)
+- `HTTPD_SERVE_NODE_IMAGES` - used by runhttpd script, controlls access
+   to the `/shared/html/images` directory via the default virtual host
+   `(HTTP_PORT)`.  (defaullt `true`)
```
dtantsur, line 57: nit: default

ironic-config/httpd.conf.j2
```diff
@@ -0,0 +1,82 @@
...
+Listen [::]:{{ env.HTTP_PORT }}
+{% else %}
+Listen {{ env.IRONIC_URL_HOST }}:{{ env.HTTP_PORT }}
+{% endif %}
+Include conf.modules.d/*.conf
+User apache
+Group apache
+
+ServerAdmin root@localhost
+ServerName www.example.com:80
```
dtantsur, line 12: I don't think we should be doing this? Maybe use IRONIC_IP and the actual port here?

ironic-config/httpd.conf.j2
```diff
@@ -0,0 +1,82 @@
...
+
+<Files ".ht*">
+    Require all denied
+</Files>
+
+ErrorLog "/dev/stderr"
+
+LogLevel warn
+
+<IfModule log_config_module>
```
dtantsur, line 47: Everything below this line seems to be repeating the defaults or at least stuff that is not metal3-specific. Do we really need it?

Dockerfile
```diff
@@ -46,13 +46,14 @@ COPY --from=ironic-builder /tmp/ipxe/src/bin/undionly.kpxe /tmp/ipxe/src/bin-x86
...
+
+# DNSMASQ
 COPY ironic-config/dnsmasq.conf.j2 /etc/
-COPY ironic-config/inspector.ipxe.j2 ironic-config/ironic-python-agent.ign.j2 /tmp/
 
 # Custom httpd config, removes all but the bare minimum needed modules
-COPY ironic-config/httpd.conf /etc/httpd/conf.d/
+COPY ironic-config/httpd.conf.j2 /etc/httpd/conf/
 COPY ironic-config/httpd-modules.conf /etc/httpd/conf.modules.d/
-COPY ironic-config/apache2-ironic-api.conf.j2 /etc/httpd-ironic-api.conf.j2
```
kashifest, line 55: so this configuration file is not needed anymore?

kashifest (commented):
lgtm
One question inline.

README.md
```diff
@@ -52,6 +52,9 @@ functionality:
 - `DNSMASQ_EXCEPT_INTERFACE` - interfaces to exclude when providing DHCP address
   (default `lo`)
 - `HTTP_PORT` - port used by http server (default `80`)
+- `HTTPD_SERVE_NODE_IMAGES` - used by runhttpd script, controlls access
```
tuminoid, line 55:
```suggestion
- `HTTPD_SERVE_NODE_IMAGES` - used by runhttpd script, controls access
```

tuminoid (commented):
One typo, plus the servername pointed out earlier is still to be fixed.

Dockerfile
Rozzii: it has been moved up a bit to L49 https://github.com/metal3-io/ironic-image/pull/428/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R49

ironic-config/httpd.conf.j2
Rozzii: I was looking into these some of these are not hard coded defaults,  I think neither the logging nor the mime types are the hard coded defaults.  

When I have more time I could go through this file and remove options that match the hard coded defaults. I will also turn  EnableSendfile on configurable in a followup anyways.

But for now I would like to merge this and not delay the merger further by introducing test failures that might result from deleting something here.

lentzi90 (commented):
/lgtm

PR #429, opened 2023-08-08
introduction of custom iPXE firmware support

Rozzii:
This PR adds the following, :

      - Support for detecting iPXE certificates
      - When iPXE certs are detected, the iPXE services
        and conigs will be advertised with https addresses
      - Enforce chainloading of the iPXE firmware provided by
        the user even if the client machine comes with an iPXE
        firmware by default
      - Adds modular iPXE apache2 config file to enable separate
        iPXE specific virtual host (port) where TLS is enforced
      - Adds ipxe builder script that can be run as an entry point
        to standalone container or as a K8s pod init container
      - custom iPXE config file template has been added thus Ironic can
        generate node specific iPXE config files that use the custom iPXE
        firmware
      - a set of environment variable that control the new iPXE related
        features

This PR builds on top of https://github.com/metal3-io/ironic-image/pull/428 so it will be on hold until that is merged.

IPXE building functionality was originally planned to be hosted in this repository but after extensive community discussion I have decided to move the iPXE firmware builder to it's own repository.


ironic-config/httpd.conf.j2
```diff
@@ -0,0 +1,82 @@
...
+
+AddDefaultCharset UTF-8
+
+<IfModule mime_magic_module>
+    MIMEMagicFile conf/magic
+</IfModule>
+
+PidFile /var/tmp/httpd.pid
+
+EnableSendfile on
```
tuminoid, line 72: I guess you decided to leave EnableSendFile on despite the possible compatiblity issues?

scripts/buildipxe
```diff
@@ -0,0 +1,85 @@
...
+ARCH=$(uname -m | sed 's/aarch/arm/')
+# NOTE(elfosardo): warning should not be treated as errors by default
+cd "/tmp/ipxe-source/src"
+if [[ "${IPXE_ENABLE_IPV6}" == "true" ]]; then
+    sed -i 's/^\/\/#define[ \t]NET_PROTO_IPV6/#define\tNET_PROTO_IPV6/g' \
+        "config/general.h"
+fi
+if [[ "${IPXE_ENABLE_TLS}" == "true" ]]; then
+    if [[ ! -r "${IPXE_CERT_FILE}" ]]; then
+        >&2 echo "ERROR: iPXE TLS support is enabled but cert is missing!"
```
tuminoid, line 63:
```suggestion
        echo >&2 "ERROR: iPXE TLS support is enabled but cert is missing!"
```

scripts/buildipxe
```diff
@@ -0,0 +1,83 @@
...
+        exit 1
+    fi
+# In case building ipxe firmware from upstream git repo directly
+# Requires Internet access!
+elif [[ "${IPXE_BUILD_FROM_CACHE}" == "false" ]] \
+         && [[ "${IPXE_BUILD_FROM_REPO}" == "true" ]]; then
+    git clone --depth 1 --branch "${IPXE_RELEASE_BRANCH}" \
+        "https://github.com/ipxe/ipxe.git" \
+        "/tmp/ipxe-source"
+fi
```
lentzi90, line 44: I wonder if it would make sense to validate the variables before this? My understanding is that we require either `IPXE_BUILD_FROM_CACHE=true && IPXE_BUILD_FROM_REPO=false` or the other way around so we could check that.

Alternatively, we can make this a bit easier on the user by doing more like a fall through. If user asks to build from cache, then try it. If it fails fall back to repo. If we go with this alternative I would say we do not need an explicit variable for `IPXE_BUILD_FROM_REPO`.

lentzi90 (commented):
Really nice work from what I can understand. I must confess this is a bit outside of my expertise though

scripts/buildipxe
Rozzii: I see your point, but my logic here was to make cache based building take precedence, and I would like to avoid fall through logic as much as possible, my idea was to have explicit config with explicit precedence and to fail hard in case pxe buildig was needed but not configured.

Rozzii: And the reason that cache based building takes precedence is that I would like to avoid reaching for external git repos a secondary option, I consider local cache a safer default choice.

lentzi90: Ok I'm fine with no fallthrough. However, I still think some validation could be helpful here instead of ending up below at 

> ERROR: The ipxe firmware source is missing, check the env vars!

It may not be obvious what variables to check.

Rozzii: Do I understand correctly, that you are proposing an extra check that would check both IPXE_BUILD_FROM_CACHE and IPXE_BUILD_FROM_REPO and if both of these are false it would say that these are and the build can't be executed ? That is fine by me  I can add that, I agree that checking /tmp/ipxe-source is not the best approach.

lentzi90: Exactly, it would be helpful to have this check I think

Rozzii: done.

lentzi90 (commented):
I don't understand all of it, but it looks solid. Hope more people will review
/lgtm

prepare-image.sh
```diff
@@ -8,7 +8,7 @@ echo "install_weak_deps=False" >> /etc/dnf/dnf.conf
 # Tell RPM to skip installing documentation
 echo "tsflags=nodocs" >> /etc/dnf/dnf.conf
 
-dnf install -y python3 python3-requests 'dnf-command(config-manager)'
+dnf install -y python3 python3-requests 'dnf-command(config-manager)' gcc make git-core perl xz-devel
```
Rozzii, line 11: Changes to this file might be a topic of debate or topic of deeper discussion because this affects the size of the OCI image.

But I would really like to support this functionality so your input is very welcome @elfosardo @dtantsur

dtantsur: As discussed on the meeting, could you please consider creating a new container image (akin to ipa-downloader) to build the iPXE firmware? These new dependencies are quite heavy, and in any case we should avoid piling more functionality to ironic-image that is not about launching ironic or something absolutely required for it.

hardys: +1 as we discussed in the meeting if this is going to be an optional initContainer IMO it would be better to create a new container image, particularly given the additional dependencies aren't particularly small

Rozzii: - I have no objections to a separate container image, your arguments make sense to me.

- What do you think should the separate container live in this repo, or should it have it's own repo?

- Adding an additional repo that has to be versioned, branched, tested separately and it would be considerable work for this relatively small image. But on the other hand ipa-downloader is even simpler and smaller and it has it's own repo so I am not against that either. @tuminoid @kashifest @lentzi90  your input would be appreciated also.

lentzi90: Separate container image sounds good to me. Separate repositories for every image sounds a bit much to me though. I would prefer to merge at least some of them. Something like a "ironic-image-utils" repo?

kashifest: +1 to @lentzi90's proposal. I also don't see a need for separate repository, specially since we are already struggling to manage versioning/branching etc for existing images like ironic-image, mariadb-image or do we already have clear versioning schemes in mind for all these exisinting repos and those that we propose to come in future?

Rozzii: Okay, thanks for the responses, I was also thinking about this, and I would propose to have this ipxe builder as part of the ironic-image repository as a separate image but also I would like to bring the IPA-downloader and the ironic-client images under the same roof in the ironic-image repository. I believe it would be beneficial if all this component/utility images would be versioned together and live on the same release branches together with the ironic-imagein the future (when we will have those) as they are all related in some way.

tuminoid: Yes, I agree with @Rozzii  latest comment. We should group these so they're more manageable.

hardys: +1 I agree it doesn't necessarily need to be a new repository provided we can build the images independently - we do already have some utility images in this repo in the [resources subdir](https://github.com/metal3-io/ironic-image/tree/main/resources) although those are only test related, not runtime dependencies

Rozzii:
PTAL: @dtantsur @hardys @elfosardo

lentzi90 (commented):
/lgtm

kashifest:
@dtantsur @elfosardo can we move forward with this, this is hanging here absent review for a month now

elfosardo (changes requested):

ironic-config/apache2-ipxe.conf.j2
```diff
@@ -0,0 +1,35 @@
...
+        Allow from all
+    </Directory>
+    <Directory "/shared/html/(redfish|ilo|images)/">
+        Order Deny,Allow
+        Deny from all
+    </Directory>
+</VirtualHost>
+
+<Location ~ "^/grub.*/">
+    SSLRequireSSL
```
dtantsur, line 24: nit: Why not just everywhere?

ironic-config/dnsmasq.conf.j2
```diff
@@ -29,13 +29,23 @@ dhcp-option=option{% if ":" in env["DNS_IP"] %}6{% endif %}:dns-server,{{ env["D
...
+{% endif %}
 
 # Note: Need to test EFI booting
 dhcp-match=set:efi,option:client-arch,7
 dhcp-match=set:efi,option:client-arch,9
 dhcp-match=set:efi,option:client-arch,11
-# Client is PXE booting over EFI without iPXE ROM; send EFI version of iPXE chainloader
+# Client is PXE booting over EFI without iPXE ROM; send EFI version of iPXE chainloader do the same also if iPXE ROM boots but TLS is enabled
+{%- if env.IPXE_TLS_SETUP == "true"  %}
+dhcp-boot=tag:efi,tag:ipxe,snponly.efi
```
dtantsur, line 47: This does not look right. You're saying "if iPXE on UEFI, boot iPXE".

ironic-config/dnsmasq.conf.j2
```diff
@@ -29,13 +29,23 @@ dhcp-option=option{% if ":" in env["DNS_IP"] %}6{% endif %}:dns-server,{{ env["D
 # IPv4 Configuration:
 dhcp-match=ipxe,175
 # Client is already running iPXE; move to next stage of chainloading
+{%- if env.IPXE_TLS_SETUP == "true"  %}
+# iPXE with (U)EFI
+dhcp-boot=tag:efi,tag:ipxe,http://{{ env.IRONIC_URL_HOST }}:{{ env.HTTP_PORT }}/custom-ipxe/snponly.efi
```
dtantsur, line 34: "http://" here sorta defeats the whole point.. if the attacker can intersect the connection, they can provide a different image?

ironic-config/ipxe_config.template
```diff
@@ -0,0 +1,81 @@
+#!ipxe
```
dtantsur, line 1: I know Ironic allows copying this file, but I'd much rather prefer we don't do it. You're going to miss any upstream fixes, and reconciling will be very hard. Can we add the missing functionality in Ironic?

ironic-config/ironic.conf.j2
```diff
@@ -116,6 +116,8 @@ ramdisk_image_download_source = {{ env.IRONIC_BOOT_ISO_SOURCE }}
 {% endif %}
 {% if env.IRONIC_EXTERNAL_HTTP_URL %}
 external_http_url = {{ env.IRONIC_EXTERNAL_HTTP_URL }}
+{% elif env.IPXE_TLS_SETUP == "true" %}
+external_http_url = https://{{ env.IRONIC_URL_HOST }}:{{ env.IPXE_TLS_PORT }}
```
dtantsur, line 120: This option is unrelated to iPXE, I'm not quite sure why you change it.

ironic-config/apache2-ipxe.conf.j2
Rozzii: I have a few things in min when I have done this
- I want to use the modular functionality of the Apache config files, so config files that are related to a type of boot interface should only restrict files that are also related to the same Ironic boot interface/boot method.
- The other thing is that, at the moment  firmware loading is not secured by TLS only the image download this I will explain in detail under your other comments.

ironic-config/dnsmasq.conf.j2
Rozzii: The goal here is to always force the chain loading of our iPXE firmware and not whatever is running on the machine by default. Our firmware will have the ability to exit this loop.

ironic-config/dnsmasq.conf.j2
Rozzii: You are right this will load the firmware via http, at the moment the TLS support is only introduced for the later part of the chain loading .There are a simple reason for this, we don't know what will be running on a machine by default, it could be and iPXE firmware that is unable to use https and first we need to provide a firmware that we can use with TLS.

The topic of man in the middle during firmware loading is very important I agree but that will be a followup feature based on all the things that are getting introduced in the ipxe-builder image and here.  Securing the firmware loading won't be as easy as turning on TLS as I have mentioned we don't know what the host machine is capable at the initial stage of the loading. I plan to go into the direction of "trusted firmware" which would mean that we would only allow a firmware to get to our images if it can identify itself as a trusted image.

ironic-config/ipxe_config.template
Rozzii: I would like to get this in first, but for sure I can later implement changes from the Ironic side also I agree eventually a natively supported approach should be the way to do it.

ironic-config/dnsmasq.conf.j2
dtantsur: Okay, so it will only work for your custom firmware? :-/

Rozzii: Yes because the firmware has to contain the correct certificate, although by default iPXE supports the Mozzilla cert boundle but in our case we have our own for Ironic.

I could add https address support for loading loading the boot.ipxe directly  but then that would exclude all machines that doesn't have a TLS enabled IPXE because I can't enable the two addresses at the same time because the dhcp "tag:"s have to match.

Rozzii: I have looked into the possibility to differentiate our custom firmware from a regular TLS enabled iPXE via a custom dhcp tag but I have not found any documented ways to do it . I am not sure in general whether a custom DHCP tag is a thing or not.

ironic-config/ironic.conf.j2
Rozzii: Thanks this was a nice catch, I have removed these two lines, these were leftovers from the version where I wanted to do the basic auth differently not via image_info as it has been implemented now.

Rozzii: Or maybe from test I don't really remember.

elfosardo (approved):

Rozzii:
@lentzi90  could you please lgtm this? it has the "hold" only because we are not sure this should go in the first ironic-image release or not

Rozzii:
/hold cancel
Now that the first release is done I will merge this as agreed @elfosardo .

PR #431, opened 2023-09-15
Switch from current-tripleo to puppet-passed-ci

derekhiggins:
current-triple is no longer maintained, puppet-passed-ci has at least gone through some RDO packstack scenario ci jobs.

### DIFF
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -8,13 +8,15 @@ echo "install_weak_deps=False" >> /etc/dnf/dnf.conf
 # Tell RPM to skip installing documentation
 echo "tsflags=nodocs" >> /etc/dnf/dnf.conf
 
-dnf install -y python3 python3-requests 'dnf-command(config-manager)'
+dnf install -y 'dnf-command(config-manager)'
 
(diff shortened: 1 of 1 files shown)


prepare-image.sh
```diff
@@ -8,13 +8,15 @@ echo "install_weak_deps=False" >> /etc/dnf/dnf.conf
 # Tell RPM to skip installing documentation
 echo "tsflags=nodocs" >> /etc/dnf/dnf.conf
 
-dnf install -y python3 python3-requests 'dnf-command(config-manager)'
+dnf install -y python3 'dnf-command(config-manager)'
```
hardys, line 11: nit but I guess we don't need the python3 here anymore given it's no longer run after the curl, and will get pulled in via deps from IRONIC_PKG_LIST?

derekhiggins: Howdy Steve o/, Updated the PR.

elfosardo:
thanks @derekhiggins 
we switched to building the container from source by default a couple of months ago so this change won't be really visible as rpm method has been deprecated, but I'm going to approve it as it's useful in case someone still wants to build the container using packages

PR #433, opened 2023-10-04
Remove unused ironic-python-agent.ign.j2

elfosardo:
The file is not used anymore and it's absent from the Dockerfile or any script


hardys:
lgtm - seems this one was missed when the other related pieces were removed in https://github.com/metal3-io/ironic-image/pull/428

hardys:
lgtm

PR #435, opened 2023-10-16
add ironic-image branching and releasing procedure

elfosardo:


docs/releasing.md
```diff
@@ -0,0 +1,128 @@
...
+Release documentation in release branches may be outdated.
+
+## Before making a release
+
+Things you should check before making a release:
+
+- Check the
+  [Metal3 release process](https://github.com/metal3-io/metal3-docs/blob/main/processes/releasing.md)
+  for high-level process and possible follow-up actions
+- Verify the latest bgufix branches in ironic and ironic-inspector upstream repositories;
```
mquhuy, line 17: typo

docs/releasing.md
```diff
@@ -0,0 +1,128 @@
...
+  ## Permissions
+
+Creating a release requires repository `write` permissions for:
+
+- Tag pushing
+- Branch creation
+- GitHub Release publishing
+
+These permissions are implicit for the org admins and repository admins.
+Release team member gets his/her permissions via `metal3-release-team`
```
mquhuy, line 34: "A release team member"?

docs/releasing.md
```diff
@@ -0,0 +1,128 @@
...
+membership. This GitHub team has the required permissions in each repository
+required to release CAPM3. Adding person to the team gives him/her the necessary
```
Rozzii, line 36:
```suggestion
required to release CAPM3. Adding person to the team provides the necessary
```
Just to make sure we don't exclude anyone.

docs/releasing.md
```diff
@@ -0,0 +1,128 @@
+# Releasing ironic-image
+
+This document explains how to prepare and publish a release for ironic-image, and
+how often a new release branch is created.
+
+**NOTE**: Always follow
+[release documentation from the main branch](https://github.com/metal3-io/ironic-image/blob/main/docs/releasing.md).
+Release documentation in release branches may be outdated.
+
+## Before making a release
```
smoshiur1237, line 10:
```suggestion
## Pre-release actions
```

smoshiur1237 (commented):
Thanks for the release doc. Few improvement comments added here.

docs/releasing.md
```diff
@@ -0,0 +1,128 @@
...
+visible.
+
+<!-- No example PR yet. To be added when first release from branch is made -->
+
+## Additional actions outside this repository
+
+Further additional actions may be required in the Metal3 project after
+ironic-image release.
+For that, please continue following the instructions provided in
+[Metal3 release process](https://github.com/metal3-io/metal3-docs/blob/main/processes/releasing.md)
```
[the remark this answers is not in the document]
dtantsur: I don't think sushy-tools/vbmc should be versioned together with Ironic, we expect them to work with any versions. Not sure about MariaDB either, but I can live with that/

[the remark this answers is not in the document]
dtantsur: ipa-downloader is also interesting since right now it always downloads IPA master. So it makes little sense to version or even release it. We should fix that though.

docs/releasing.md
```diff
@@ -0,0 +1,128 @@
...
+
+Clone the repository:
+`git clone git@github.com:metal3-io/ironic-image`
+
+or if using existing repository, verify your intended remote is set to
+`metal3-io`: `git remote -v`. For this document, we assume it is `origin`.
+
+- If creating a new release branch, identify the commit you wish to create the
+  branch from, and create a branch `release-X.x` where `X.x` corresponds to the
+  ironic upstream bugfix version, for example for `bugfix-22.1` it's `release-22.1`:
```
dtantsur, line 54: nit: depending on the exact moment, it can also be a stable branch

dtantsur: I suggest we don't do things like `release-<codename>` in this case and keep the number (corresponding to Ironic's release number)

docs/releasing.md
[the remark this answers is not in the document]
elfosardo: During the weekly meeting discussion we've decided to leave things as they are at the moment, it may makes sense for ipa-downloader to be versioned though.
We have the capability to build the ipa ramdisk in metal3 directly, which will help if we decide to do that.

Rozzii (dismissed):
/lgtm

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
...
+  branch from, and create a branch `release-X.x` where `X.x` corresponds to the
+  ironic upstream release version, for example for `ironic 22.1` it's `release-22.1`:
+  `git checkout <sha> -b release-X.x` and push it to remote:
+  `git push origin release-X.x` to create it
+- If creating a new patch release, use existing branch `release-X.x`:
+  `git checkout origin/release-X.x`
+- In the newly created release branch:
+   - Add a copy of upper-constraints.txt from [Openstack Requirements](https://opendev.org/openstack/requirements)
+     to replace the placeholder
+   - Pin ironic and ironic-inspector to match the corresponding bugfix or
```
jayofdoom, line 64: Ironic only tests upgrades to/from major 6 month releases. In addition, the stable 6 month releases are supported for a long life cycle -- guaranteed for 18 months -- where bugfix branches have a much shorter lifecycle.

Additionally, there is minimal value in metal3 using Ironic bugfix branches as it's relatively rare that Ironic has a new feature that metal3 immediately consumes. 

What is the primary driver behind tracking bugfix releases in Ironic?

elfosardo: @jayofdoom thanks for the comment!
we're not going to always tie the release/branch of the ironic-image to bugfix branches, we'll use either bugffix or stable based on which is the closest one as a general tule of thumb
the main reason why we can use the bugfix branches, considering their short life, is that the release/branches of the ironic-image will also be supported for no longer than 6 months, except very special cases
we aim to branch 3-4 times a year, meaning every 3-4 months (shorter than lifetime of a bugfix branch), and release maybe even more often

dtantsur: > it's relatively rare that Ironic has a new feature that metal3 immediately consumes.

This is not quite true, by the way. We have several features developed specifically for Metal3 consumption. This includes JSON RPC or all-in-one Ironic executable in the past, firmware upgrades and non-bootable ISO presently.

jayofdoom: Ack. I think part of why this isn't ideal is that bugfix releases are not handled entirely via automation on the Ironic side, so it makes me nervous to have another OSS downstream consumer of it -- but frankly, you both know I've felt that way about them since the beginning.

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
...
+These permissions are implicit for the org admins and repository admins.
+A release team member gets permissions via `metal3-release-team`
+membership. This GitHub team has the required permissions in each repository
+to make a release. Adding person to the team provides the necessary
+rights  in all relevant repositories in the organization. Individual persons
+should not be given permissions directly.
+
+## Process
+
+ironic-image uses [semantic versioning](https://semver.org). For version `vX.x.y`:
```
tuminoid, line 44: Why `vX.x.y`? Semver is specific of the format being `vX.Y.Z`

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
...
+### Repository setup
+
+Clone the repository:
+`git clone git@github.com:metal3-io/ironic-image`
+
+or if using existing repository, verify your intended remote is set to
+`metal3-io`: `git remote -v`. For this document, we assume it is `origin`.
+
+- If creating a new release branch, identify the commit you wish to create the
+  branch from, and create a branch `release-X.x` where `X.x` corresponds to the
```
tuminoid, line 55: Same for all references of `X.x`  -> `X.Y`

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
...
+  ironic upstream release version, for example for `ironic 22.1` it's `release-22.1`:
+  `git checkout <sha> -b release-X.x` and push it to remote:
+  `git push origin release-X.x` to create it
+- If creating a new patch release, use existing branch `release-X.x`:
+  `git checkout origin/release-X.x`
+- In the newly created release branch:
```
tuminoid, line 61: This is a bit problematic timing wise, since we need to add CI support for the newly created branch before you can merge PRs to it. All of the other repos do the uplifts etc in main first, and then branch out.

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
...
+   - If needed, pin versions of ironic-lib and sushy to grant compatibility with
+     ironic and ironic-inspector; this can probably be avoided considering that
+     we're using upper-constraints
+
+### Tags
+
+First we create a primary release tag, that triggers release note creation and
+image building processes.
+
+- Create a signed, annotated tag with: `git tag -s -a vX.x.y -m vX.x.y`
```
tuminoid, line 75: Same as before

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
...
+Branch protection rules need to be applied to the new release branch. Copy the
+settings after the previous release branch, with the exception of
+`Required tests` selection. Required tests can only be selected after new
+keywords are implemented in Jenkins JJB, and project-infra, and have been run at
+least once in the PR targeting the branch in question.
+
+### Update README.md and build badges
+
+Update `README.md` with release specific information, both on `main` and in the
+new `release-X.x` branch as necessary.
```
tuminoid, line 116: Same as before

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
...
+
+<!-- No example PR yet. To be added when first release from branch is made -->
+
+In the `release-X.x` branch, update the build badges in the `README.md` to point
+to correct Jenkins jobs, so the build statuses of the release branch are
+visible.
+
+<!-- No example PR yet. To be added when first release from branch is made -->
+
+### Branches lifecycle
```
tuminoid, line 126: This section should probably be elsewhere, like `CONTRIBUTING.md`, like it is in CAPM3, IPAM and BMO.

docs/releasing.md
elfosardo: main does not need to have an upper-constraints file
Also I think setting up CI first is not a problem, we'll wait for that to be up and then update upper-constraints
I'll specify it in this proposal

docs/releasing.md
elfosardo: I can move it, but honestly it does make sense to me to keep it here since it's the main place where we talk about branches and release strategy

docs/releasing.md
tuminoid: I know, hence mentioned the way its done elsewhere is problematic. Docs update is all I wanted, thanks for fixing.

docs/releasing.md
tuminoid: Sure, no strong opinion. We can always move later if feeling like it.

tuminoid (approved):
LGTM, thanks for putting this forward.
Getting ironic-image separately released with pinned dependencies is super important for the project stability and rebuildability.

Rozzii (approved):
/lgtm

docs/releasing.md
dtantsur: To close the loop here: Riccardo is looking into improving the automation.

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
...
+  and ironic-inspector upstream repositories;
+- Verify compatibility with latest sushy and ironic-lib upstream releases or master
+  branches based on ironic and ironic-inspector bugfix or stable requirements
+  and constraints
+- Verify openstack upper-constraints compatibility with ironic and ironic-inspector
+  bugfix or stable branches
+- Verify any other direct or indirect dependency is uplifted to close any public
+  vulnerabilities
+
+  ## Permissions
```
dtantsur, line 27: Weird indentation here, did you mean to use ###?

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
+# Releasing ironic-image
+
+This document explains how to prepare and publish a release for ironic-image, and
+how often a new release branch is created.
```
dtantsur, line 4: The "How often" part could use move love. The way I read the document, we should aim for a release shortly after a corresponding Ironic stable or bugfix branch is created, otherwise the newly created ironic-image branch will effectively go back in time once you pin the projects.

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
...
+- If creating a new release branch, identify the commit you wish to create the
+  branch from, and create a branch `release-X.Y` where `X.Y` corresponds to the
+  ironic upstream release version, for example for `ironic 22.1` it's `release-22.1`:
+  `git checkout <sha> -b release-X.Y` and push it to remote:
+  `git push origin release-X.Y` to create it
+- If creating a new patch release, use existing branch `release-X.Y`:
+  `git checkout origin/release-X.Y`
+- Wait for the CI support to be configured, then in the newly created release branch:
+   - Add a copy of upper-constraints.txt from [Openstack Requirements](https://opendev.org/openstack/requirements)
+     to replace the placeholder
```
dtantsur, line 63: A few more words here probably? Like, if we pin ironic to a stable branch, the requirements should also come from the same stable branch. For bugfix branches... just whatever the current version is?

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
...
+   - Pin ironic and ironic-inspector to match the corresponding bugfix or
+     stable branches
+   - If needed, pin versions of ironic-lib and sushy to grant compatibility with
+     ironic and ironic-inspector; this can probably be avoided considering that
+     we're using upper-constraints
```
dtantsur, line 68: Yes, this point can be removed for this specific reason.

docs/releasing.md
```diff
@@ -0,0 +1,138 @@
...
+- Check for duplicates, reverts, and incorrect classifications of PRs, and
+  whatever release creation tagged to be manually checked.
+- For any superseded PRs (like same dependency uplifted multiple times, or
+  commit revertions) that provide no value to the release, move them to
+  Superseded section. This way the changes are acknowledged to be part of the
+  release, but not overwhelming the important changes contained by the release.
+- If the release you're making is not a new major release, new minor release, or
+  a new patch release from the latest release branch, uncheck the box for latest
+  release.
+- If it is a release candidate (RC) or a pre-release, tick pre-release box.
```
dtantsur, line 98: We're not going to have pre-releases here, are we?

docs/releasing.md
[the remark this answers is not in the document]
dtantsur: Let's write it somewhere that we will use master IPA for now with a goal of moving towards custom builds eventually?

docs/releasing.md
elfosardo: I've added some words in the Branches lifecycle paragraph

docs/releasing.md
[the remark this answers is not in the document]
elfosardo: I've added another paragraph below

Rozzii (approved):
/lgtm

PR #436, opened 2023-10-18
Use bash process substitution instead of pipe

elfosardo:
Provides more debugging info and can avoid situation where the pipe hides errors.


Rozzii (approved):
/lgtm

PR #439, opened 2023-10-25
Make sure /certs/ca is created in configure-nonroot

dtantsur:
Otherwise, it cannot be created later if something is mounted under it.


tuminoid (commented):
/lgtm

lentzi90 (commented):
/lgtm

PR #440, opened 2023-11-02
python-ironicclient debug container image

jklippel:
The Metal³ ironic container image does not include an ironic-client installation. Furthermore there is no longer a package in centos that provides the ironic-client via dnf.

This commit adds a Dockerfile for a docker container image which contains an installation of ironic-pythonclient. The docker image is configured such that it can be used as debug container alongside the Metal³ ironic container, providing an interactive shell environment which can be used to interact with Metal³ ironic using the "baremetal" cli.

### DIFF
--- a/resources/ironicclient/Dockerfile
+++ b/resources/ironicclient/Dockerfile
@@ -1,0 +1,15 @@
+ARG BASE_IMAGE=quay.io/centos/centos:stream9
+FROM $BASE_IMAGE
+
+COPY ironic.profile /etc/profile.d/ironic.sh
+
+RUN set -eux ; \
+    export LANG=C.UTF-8; \
+    BUILD_PACKAGES="gcc python3-pip" ; \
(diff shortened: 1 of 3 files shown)


tuminoid:
We have https://github.com/metal3-io/ironic-client How is this different from that?

jklippel:
> Hi @jklippel! We have https://github.com/metal3-io/ironic-client How is this different from that?

Hi @tuminoid, sorry for my ignorance, I simply did not know about it. If at all I guess I should open a pull request there.
Thanks for the info and sorry for the fuss.

tuminoid:
> > Hi @jklippel! We have https://github.com/metal3-io/ironic-client How is this different from that?
> 
> Hi @tuminoid, sorry for my ignorance, I simply did not know about it. If at all I guess I should open a pull request there. Thanks for the info and sorry for the fuss.

No problem. Thank you for the effort!

PR #442, opened 2023-11-07
Remove ServerName from httpd.conf

elfosardo:
The ServerName value is not mandatory and it's not really needed as apache can get that from /etc/hosts.
Also this configuration will never work in an ipv6 environemnt as ipv6 ips are not allowed as values for ServerName.


elfosardo:
looks like CI issue still ongoing

elfosardo:
\o/

PR #447, opened 2023-11-15
Create /shared in configure-nonroot.sh

dtantsur:
Currently, all containers mount it, but it's actually not required for
dnsmasq since it only needs a few bits of static configuration.


tuminoid:
/test-centos-integration-mIain
Ironic in Ubuntui is not running in k8s, so centos is relevant test.

tuminoid (commented):
/lgtm

PR #448, opened 2023-11-15
Allow guessing PROVISIONING_INTERFACE by PROVISIONING_IP

dtantsur:
This simplifies the case where PROVISIONING_IP is known but
the exact interface name is not.


iurygregory (commented):
/lgtm

PR #451, opened 2023-11-24
sushy-tools: make listen interface customizable

dtantsur:
Hardcoding :: makes it impossible to override the listen IP, which can
by handy in some testing scenarios.

### DIFF
--- a/resources/sushy-tools/redfish-emulator.sh
+++ b/resources/sushy-tools/redfish-emulator.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -eux -o pipefail
+
+CONFIG="${SUSHY_TOOLS_CONFIG:-/root/sushy/conf.py}"
+ARGS=
+
+if [[ -f "${CONFIG}" ]]; then
+    ARGS="${ARGS} --config ${CONFIG}"
+fi
+
+if [[ ! -f "${CONFIG}" ]] || ! grep -q "^SUSHY_EMULATOR_LISTEN_IP =" "${CONFIG}"; then
+    # Listen on all interfaces unless explicitly configured otherwise.
+    ARGS="${ARGS} --interface ::"
+fi
+
+# shellcheck disable=SC2086
+exec /usr/local/bin/sushy-emulator --debug $ARGS
--- a/resources/sushy-tools/Dockerfile
+++ b/resources/sushy-tools/Dockerfile
@@ -9,5 +9,6 @@ RUN apt-get update && \
     pip3 install --no-cache-dir \
         sushy-tools==${SUSHY_TOOLS_VERSION} libvirt-python
 
-CMD /usr/local/bin/sushy-emulator -i :: \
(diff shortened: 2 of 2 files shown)


resources/sushy-tools/redfish-emulator.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -eux -o pipefail
+
+CONFIG="${SUSHY_TOOLS_CONFIG:-/root/sushy/conf.py}"
+ARGS=
+
+if [[ -f "${CONFIG}" ]]; then
```
tuminoid, line 8:
```suggestion
if [[ -r "${CONFIG}" ]]; then
```
Maybe `-r` so the file exists and is readable too? Doesn't matter right now as its running as`root` now, so feel free to ignore too.

resources/sushy-tools/redfish-emulator.sh
```diff
@@ -0,0 +1,18 @@
...
+    ARGS="${ARGS} --config ${CONFIG}"
+fi
+
+if [[ ! -f "${CONFIG}" ]] || ! grep -q "^SUSHY_EMULATOR_LISTEN_IP =" "${CONFIG}"; then
```
tuminoid, line 12:
```suggestion
elif ! grep -q "^SUSHY_EMULATOR_LISTEN_IP =" "${CONFIG}"; then
```
Avoids repetive check about config existing.

resources/sushy-tools/redfish-emulator.sh
```diff
@@ -0,0 +1,18 @@
...
+    # Listen on all interfaces unless explicitly configured otherwise.
+    ARGS="${ARGS} --interface ::"
+fi
+
+# shellcheck disable=SC2086
+exec /usr/local/bin/sushy-emulator --debug $ARGS
```
tuminoid, line 18:
```suggestion
exec /usr/local/bin/sushy-emulator --debug ${ARGS}
```

tuminoid (commented):
Some nits

resources/sushy-tools/redfish-emulator.sh
dtantsur: I did this on purpose, I want the service to fail if the rights are not set up correctly (it can happen even with root if selinux is enforcing).

resources/sushy-tools/redfish-emulator.sh
dtantsur: I think you're wrong, there an "or", not an "and" here.

resources/sushy-tools/redfish-emulator.sh
tuminoid: If we know it is a real possibility to have file existing but not readable, why not explicitly fail and exit here already? Did not check what sushy-tools does in such case, but in general, services may exit gracefully, dump stack or even just ignore the failing config file.

tuminoid (commented):
/lgtm

Nits are just nits.

PR #452, opened 2023-11-27
Detect default images downloaded by ipa-downloader

dtantsur:
In the future, this will simplify BMO configuration by no longer
requiring deploy kernel/ramdisk to be provided via environment
variables.


tuminoid:
Since the filesystem sharing is different on k8s side, let's run this as well. LGTM in general.

/test-centos-integration-main

tuminoid:
Btw, should we make the `test-centos-integration-main` required test for ironic-image, like it is in BMO etc?
@kashifest @dtantsur @elfosardo

dtantsur:
@tuminoid if there is an actual difference - yes, definitely

/test-ubuntu-integration-main
/test-centos-integration-main

elfosardo:
> Btw, should we make the `test-centos-integration-main` required test for ironic-image, like it is in BMO etc? @kashifest @dtantsur @elfosardo

yes, we should

kashifest:
> > Btw, should we make the `test-centos-integration-main` required test for ironic-image, like it is in BMO etc? @kashifest @dtantsur @elfosardo
> 
> yes, we should

Done

tuminoid (commented):
/lgtm

PR #454, opened 2023-11-30
Don't install ironic-lib from git when it's not requested

dtantsur:
When IRONIC_LIB_SOURCE is not set, we should use pip. Otherwise,
we can a conflict with upper-constraints.


Rozzii (approved):
/lgtm

iurygregory (approved):

dtantsur:
I suspect nothing will pass until we get https://review.opendev.org/c/openstack/ironic-python-agent/+/902231 in...

PR #457, opened 2023-12-10
Add openstacksdk to sushy-tools container image

steveb:
This will allow sushy-tools containers to use the nova driver, as well as the libvirt driver.


tuminoid (commented):
/lgtm

PR #463, opened 2023-12-18
add condition for ipxe_tls template usage

Rozzii:
Previously custom ironic ipxe template was taken into use in ironic config even when the TLS support was not enabled for the ironic-image.

This commit:
  - Adds  a condition to exclude/include referencing custom ipxe template based on whether TLS is enabled for the ipxe.


lentzi90 (commented):
/lgtm

lentzi90 (commented):
/lgtm

Rozzii:
/override test-ubuntu-integration-main
ubuntu test was expected to fail because of the CI issue that this PR might be able to fix but not until it is merged.

PR #470, opened 2024-01-05
[WIP] 🌱 Backward compable callback endpoint when Inspector is off

dtantsur:
Some PreprovisioningImage controller, like the one we have downstream,
hardcode {IP}:5050/v1/continue in their IPA configuration. To avoid
updating them in a lockstep with disabling Inspector, provide a small
virtual host that proxies :5050/v1/continue to :6385/v1/continue_inspection.


dtantsur:
I have alternative ideas and probably won't pursue this path.

PR #473, opened 2024-01-10
🌱 Use bookworm slim to build sushy-tools image

elfosardo:
It should reduce the final size of the image from 1.2GB to 500MB

### DIFF
--- a/resources/sushy-tools/Dockerfile
+++ b/resources/sushy-tools/Dockerfile
@@ -1,13 +1,14 @@
-FROM docker.io/library/python:3.9
+FROM docker.io/library/python:3.9.18-slim-bookworm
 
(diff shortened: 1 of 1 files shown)


tuminoid (commented):
/lgtm

From 1.18G to 0.57G, good finding.

elfosardo:
if it works :D
CI doesn't seem very happy

elfosardo:
removing gcc at the end saves us even more bringing the image down to 350 MB

tuminoid:
> removing gcc at the end saves us even more bringing the image down to 350 MB

Seems to work without it indeed. 342MB on my machine, even better.

kashifest (commented):
/lgtm

Rozzii (approved):
/lgtm

PR #475, opened 2024-02-06
[DNM] test ci

elfosardo:

Fixes #

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,8 @@ RUN git clone --depth 1 --branch v1.21.1 https://github.com/ipxe/ipxe.git && \
 COPY prepare-efi.sh /bin/
 RUN prepare-efi.sh centos
 
+### TEST
+
 FROM $BASE_IMAGE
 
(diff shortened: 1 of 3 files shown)


tuminoid:
/test-ubuntu-integration-main
//test-centos-integration-main

kashifest:
the triggers doesnt look correct, we using main branch test trigger here, perhaps we should use the correct branch name. Let me check if everything is up to date in gerrit repo

kashifest:
ok it seems like we do not have the logic to test the release branches yet , cause for that we have to decide which branch of  BMO we should test for a given ironic release branch, currently we run these integration tests based on CAPM3 branches and CAPM3 and BMO branches are coupled together for the integration tests (we should change this), we can also use BMO e2e tests here but then again we need to know for which branch of BMO we should test which branch of ironic!!! the 
`ironic-image-build` pipeline is another thing that we can leverage, currently its only building ironic image and that also from rpm for some legacy reason, we can perhaps improve this and add options to test integration of different ironic image branches with BMO/CAPM3.

PR #480, opened 2024-02-29
:bug: Fix vbmc container build after tripleo deprecation

elfosardo:
The tripleo repo has been archived as tripleo is not maintained anymore.
Let's switch to RDO repos for vbmc to avoid breakage.


dtantsur:
/override test-ubuntu-integration-main 

We need to fix the vBMC build, and the ubuntu job is pretty hopeless at this point.

PR #482, opened 2024-03-05
🐛 Stop reading ironic and inspector HTPASSWD from environment variables

MahnoorAsghar:
Security baselines such as CIS do not recommend using secrets as environment variables, but using files instead. Therefore, the IRONIC_HTPASSWD and INSPECTOR_HTPASSWD will now be populated from files instead of environment variables.

### DIFF
--- a/scripts/auth-common.sh
+++ b/scripts/auth-common.sh
@@ -2,8 +2,6 @@
 
 set -euxo pipefail
 
-export IRONIC_HTPASSWD=${IRONIC_HTPASSWD:-${HTTP_BASIC_HTPASSWD:-}}
-export INSPECTOR_HTPASSWD=${INSPECTOR_HTPASSWD:-${HTTP_BASIC_HTPASSWD:-}}
 export IRONIC_REVERSE_PROXY_SETUP=${IRONIC_REVERSE_PROXY_SETUP:-false}
 export INSPECTOR_REVERSE_PROXY_SETUP=${INSPECTOR_REVERSE_PROXY_SETUP:-false}
 
@@ -14,8 +12,35 @@ else
     export IRONIC_EXPOSE_JSON_RPC="${IRONIC_EXPOSE_JSON_RPC:-false}"
 fi
 
+set +x
 IRONIC_HTPASSWD_FILE=/etc/ironic/htpasswd
+if [[ -f "/auth/ironic/username" ]]; then
+    read -r IRONIC_HTPASSWD_USERNAME<"/auth/ironic/username"
+fi
+IRONIC_HTPASSWD_USERNAME=${IRONIC_HTPASSWD_USERNAME:-}
+if [[ -f "/auth/ironic/password" ]]; then
+    read -r IRONIC_HTPASSWD_PASSWORD<"/auth/ironic/password"
+fi
+IRONIC_HTPASSWD_PASSWORD=${IRONIC_HTPASSWD_PASSWORD:-}
+if [[ -n "${IRONIC_HTPASSWD_USERNAME}" ]]; then
+    IRONIC_HTPASSWD="$(htpasswd -n -b -B "${IRONIC_HTPASSWD_USERNAME}" "${IRONIC_HTPASSWD_PASSWORD}")"
+fi
+export IRONIC_HTPASSWD=${IRONIC_HTPASSWD:-${HTTP_BASIC_HTPASSWD:-}}
+
 INSPECTOR_HTPASSWD_FILE=/etc/ironic-inspector/htpasswd
+if [[ -f "/auth/inspector/username" ]]; then
+    read -r INSPECTOR_HTPASSWD_USERNAME<"/auth/inspector/username"
+fi
+INSPECTOR_HTPASSWD_USERNAME=${INSPECTOR_HTPASSWD_USERNAME:-}
+if [[ -f "/auth/inspector/password" ]]; then
+    read -r INSPECTOR_HTPASSWD_PASSWORD<"/auth/inspector/password"
+fi
+INSPECTOR_HTPASSWD_PASSWORD=${INSPECTOR_HTPASSWD_PASSWORD:-}
+if [[ -n "${INSPECTOR_HTPASSWD_USERNAME}" ]]; then
+    INSPECTOR_HTPASSWD="$(htpasswd -n -b -B "${INSPECTOR_HTPASSWD_USERNAME}" "${INSPECTOR_HTPASSWD_PASSWORD}")"
+fi
+export INSPECTOR_HTPASSWD=${INSPECTOR_HTPASSWD:-${HTTP_BASIC_HTPASSWD:-}}
+set -x
 
 configure_client_basic_auth()
 {


scripts/auth-common.sh
```diff
@@ -2,15 +2,18 @@
...
-export INSPECTOR_HTPASSWD=${INSPECTOR_HTPASSWD:-${HTTP_BASIC_HTPASSWD:-}}
 export IRONIC_DEPLOYMENT="${IRONIC_DEPLOYMENT:-}"
 export IRONIC_REVERSE_PROXY_SETUP=${IRONIC_REVERSE_PROXY_SETUP:-false}
 export INSPECTOR_REVERSE_PROXY_SETUP=${INSPECTOR_REVERSE_PROXY_SETUP:-false}
 
 IRONIC_HTPASSWD_FILE=/etc/ironic/htpasswd
 INSPECTOR_HTPASSWD_FILE=/etc/ironic-inspector/htpasswd
-
+if [ -s "${IRONIC_HTPASSWD_FILE}" ]; then
+        export IRONIC_HTPASSWD=$(<$IRONIC_HTPASSWD_FILE)
```
elfosardo, line 12: better define the variable first, then export it 
I think shellcheck is also complaining about this

scripts/auth-common.sh
```diff
@@ -2,15 +2,18 @@
...
+fi
+if [ -s "${INSPECTOR_HTPASSWD_FILE}" ]; then
+        export INSPECTOR_HTPASSWD=$(<$INSPECTOR_HTPASSWD_FILE)
```
elfosardo, line 15: ditto

tuminoid (commented):
I'm not following: who and where are the files populated now?

Some nits too.

MahnoorAsghar:
> I'm not following: who and where are the files populated now?
> 
> Some nits too.

The files are populated by cluster-baremetal-operator. I have a PR there which I will test alongside this one: https://github.com/openshift/cluster-baremetal-operator/pull/408

tuminoid:
> > I'm not following: who and where are the files populated now?
> > Some nits too.
> 
> The files are populated by cluster-baremetal-operator. I have a PR there which I will test alongside this one: [openshift/cluster-baremetal-operator#408](https://github.com/openshift/cluster-baremetal-operator/pull/408)

That is not part of Metal3. Who is populating/creating files in Metal3?

MahnoorAsghar:
> > > I'm not following: who and where are the files populated now?
> > > Some nits too.
> > 
> > 
> > The files are populated by cluster-baremetal-operator. I have a PR there which I will test alongside this one: [openshift/cluster-baremetal-operator#408](https://github.com/openshift/cluster-baremetal-operator/pull/408)
> 
> That is not part of Metal3. Who is populating/creating files in Metal3?

Good point. I will come up with a change for metal3 as well.

dtantsur:
This is going to break virtually all consumers: the files are not there when the container starts, we populate them.

My earlier plan was to eventually start accepting a mounted secret instead of HTPASSWD variables. Then the consumers, such as CBO, will be able to fix the problem for themselves. Maybe you could go down the same path? It's harder to backport, of course, but nor is this change really backportable.

It's also not the worst vulnerability you can imagine (the password is hashed), so I'm fine if we only fix it on the main branch and only for consumers that switch to mounting the secret.

scripts/auth-common.sh
```diff
@@ -16,6 +14,24 @@ fi
...
 INSPECTOR_HTPASSWD_FILE=/etc/ironic-inspector/htpasswd
+yaml() {
+        python3 -c "import yaml;
+try:
+    print(yaml.safe_load(open('$1'))$2)
+except:
+    print('')"
+}
+
+IRONIC_HTPASSWD=$(yaml ${IRONIC_HTPASSWD_FILE} "['data']['htpasswd']")
```
dtantsur, line 25: You're trying to read the file you will be writing (also, it's not YAML).

If you mount a secret to, say, `/auth/ironic`, you'll get a directory with files corresponding to every secret key. In you case, you'll get `/auth/ironic/username` and `/auth/ironic/password`. You need to take these and make htpasswd out of them (check the man pages of the htpasswd tool).

scripts/auth-common.sh
```diff
@@ -43,7 +59,6 @@ configure_ironic_auth()
     local config=/etc/ironic/ironic.conf
     # Configure HTTP basic auth for API server
     if [[ -n "${IRONIC_HTPASSWD}" ]]; then
-        printf "%s\n" "${IRONIC_HTPASSWD}" > "${IRONIC_HTPASSWD_FILE}"
```
dtantsur, line 46: If you don't write this file, authentication will be disabled

scripts/runhttpd
```diff
@@ -68,8 +68,6 @@ else
     export IRONIC_REVERSE_PROXY_SETUP="false" # If TLS is not used, we have no reason to use the reverse proxy
 fi
 
-write_htpasswd_files
```
dtantsur, line 71: Same here, you're disabling authentication.

scripts/auth-common.sh
dtantsur: Reference https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#create-a-pod-that-has-access-to-the-secret-data-through-a-volume

scripts/auth-common.sh
```diff
@@ -15,7 +13,38 @@ else
...
 IRONIC_HTPASSWD_FILE=/etc/ironic/htpasswd
+if [[ -f "/auth/ironic/username" ]]; then
+    read -r IRONIC_HTPASSWD_USERNAME<"/auth/ironic/username"
+fi
+IRONIC_HTPASSWD_USERNAME=${IRONIC_HTPASSWD_USERNAME:-}
+if [[ -f "/auth/ironic/password" ]]; then
+    read -r IRONIC_HTPASSWD_PASSWORD<"/auth/ironic/password"
+fi
+IRONIC_HTPASSWD_PASSWORD=${IRONIC_HTPASSWD_PASSWORD:-}
+if [[ -n "${IRONIC_HTPASSWD_USERNAME}" ]]; then
```
dtantsur, line 24: I wonder if it's going to complain about undefined variables. Maybe you need "${IRONIC_HTPASSWD_USERNAME:-}"

scripts/auth-common.sh
```diff
@@ -15,7 +13,38 @@ else
...
+    IRONIC_HTPASSWD="$(htpasswd -n -b -B "${IRONIC_HTPASSWD_USERNAME}" "${IRONIC_HTPASSWD_PASSWORD}")"
+fi
+IRONIC_HTPASSWD=${IRONIC_HTPASSWD:-${HTTP_BASIC_HTPASSWD:-}}
```
dtantsur, line 27: nit: you combine this with `export`

scripts/auth-common.sh
```diff
@@ -15,7 +13,38 @@ else
...
+if [[ -n "${IRONIC_HTPASSWD}" ]]; then
+    export IRONIC_HTPASSWD
+fi
+
 INSPECTOR_HTPASSWD_FILE=/etc/ironic-inspector/htpasswd
+if [[ -f "/auth/inspector/username" ]]; then
```
dtantsur, line 33: A bit of code duplicate here, but since inspector is going away, I don't insist on refactoring.

MahnoorAsghar: Yeah, I figured its going away soon so it should be okay hopefully.

scripts/auth-common.sh
MahnoorAsghar: Its there on line 19 :)

scripts/auth-common.sh
```diff
@@ -71,4 +96,4 @@ write_htpasswd_files()
     if [[ -n "${INSPECTOR_HTPASSWD:-}" ]]; then
         printf "%s\n" "${INSPECTOR_HTPASSWD}" > "${INSPECTOR_HTPASSWD_FILE}"
     fi
-}
+}
```
tuminoid, line 99: Can yout put the linefeed back?

tuminoid (commented):
A tiny nit left.

scripts/auth-common.sh
MahnoorAsghar: Apologies! Its back now :)

dtantsur:
/hold

Has any integration jobs run here? I don't see any results, let me check with the folks.

dtantsur:
/hold cancel

Looking good, please proceed with reviews.

PR #487, opened 2024-03-15
⚠️ Remove support for API/conductor split

dtantsur:
Running API and conductor separately mostly makes sense in the context of large OpenStack clouds. We used to use it before the all-in-one executable appeared, but now it's untested and potentially broken.

While here, enforce authentication for JSON RPC and clean up related options.


lentzi90 (commented):
/lgtm

PR #501, opened 2024-04-26
WIP: 🌱 remove world-writable socket

tuminoid:
We don't need a world-writable socket anymore. Group write is enough as we have users created in configure-nonroot.sh to share GIDs, and on top, Inspector is now removed.


tuminoid:
Since Centos passed and Ubuntu didn't, I guess we have issues still with the Ironic sockets in local Ironic installation, as mounts are different than in Centos/K8s installation.

/close

PR #506, opened 2024-05-08
🌱 Read Ironic htpasswd from file or environment variable

MahnoorAsghar:
This PR reverts #482 and #498, because it is not secure to mount the ironic username and password as plain text files into
ironic-image. The PR adds the ability to mount the ironic htpasswd as a file (to introduce a better way than just using environment variable for this purpose).

### DIFF
--- a/scripts/auth-common.sh
+++ b/scripts/auth-common.sh
@@ -11,21 +11,11 @@ else
     export IRONIC_EXPOSE_JSON_RPC="${IRONIC_EXPOSE_JSON_RPC:-false}"
 fi
 
-set +x
 IRONIC_HTPASSWD_FILE=/etc/ironic/htpasswd
-if [[ -f "/auth/ironic/username" ]]; then
-    IRONIC_HTPASSWD_USERNAME=$(</auth/ironic/username)
-fi
-IRONIC_HTPASSWD_USERNAME=${IRONIC_HTPASSWD_USERNAME:-}
-if [[ -f "/auth/ironic/password" ]]; then
-    IRONIC_HTPASSWD_PASSWORD=$(</auth/ironic/password)
-fi
-IRONIC_HTPASSWD_PASSWORD=${IRONIC_HTPASSWD_PASSWORD:-}
-if [[ -n "${IRONIC_HTPASSWD_USERNAME}" ]]; then
-    IRONIC_HTPASSWD="$(htpasswd -n -b -B "${IRONIC_HTPASSWD_USERNAME}" "${IRONIC_HTPASSWD_PASSWORD}")"
+if [[ -f "/auth/ironic/htpasswd" ]]; then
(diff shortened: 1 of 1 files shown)


Rozzii:
I was not following this "username:password" topic for some time but I have also found interesting when I have noticed that the username and the password is in plain text so I agree with this PR, thanks!
/lgtm

lentzi90 (commented):
The test failure seems to be just a timeout with 1 out of 11 succeeded. Retriggering
/test metal3-centos-e2e-integration-test-main
/approve

Rozzii:
Need an approver
/cc @elfosardo

tuminoid:
/retest
CI should be better now.

scripts/auth-common.sh
```diff
@@ -13,16 +13,8 @@ fi
 
 set +x
```
tuminoid, line 14: We don't need this since we're not using plain password anymore.

scripts/auth-common.sh
```diff
@@ -13,16 +13,8 @@ fi
...
-    IRONIC_HTPASSWD_PASSWORD=$(</auth/ironic/password)
-fi
-IRONIC_HTPASSWD_PASSWORD=${IRONIC_HTPASSWD_PASSWORD:-}
-if [[ -n "${IRONIC_HTPASSWD_USERNAME}" ]]; then
-    IRONIC_HTPASSWD="$(htpasswd -n -b -B "${IRONIC_HTPASSWD_USERNAME}" "${IRONIC_HTPASSWD_PASSWORD}")"
+if [[ -f "/auth/ironic/htpasswd" ]]; then
+    IRONIC_HTPASSWD=$(</auth/ironic/htpasswd)
 fi
 export IRONIC_HTPASSWD=${IRONIC_HTPASSWD:-${HTTP_BASIC_HTPASSWD:-}}
 set -x
```
tuminoid, line 20: Same here.

dtantsur: Yeah, I don't see a huge problem in logging the htpasswd string.

Rozzii: But is this a problem? I mean I get that it is an overkill but I don't think it is computationally expensive so can't we just leave it like that ? I would like to leave it as it is.

tuminoid: It might make debugging harder as you miss part of the flow without any logging.

elfosardo: this should probably be changed now, CI needs to rerun anyway

scripts/auth-common.sh
MahnoorAsghar: Removed.

scripts/auth-common.sh
MahnoorAsghar: Removed.

PR #507, opened 2024-05-10
🌱 enable https download in ipxe

defo89:
I am trying to use TLS in every part of the deployment chain and it seems that by default [HTTPS is disabled](https://github.com/ipxe/ipxe/blob/master/src/config/general.h#L59) in iPXE image.

image

As per https://ipxe.org/buildcfg/download_proto_https this commit would enable support (tested this in our environment).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,8 @@ WORKDIR /tmp
 
 RUN git clone --depth 1 --branch v1.21.1 https://github.com/ipxe/ipxe.git && \
       cd ipxe/src && \
+      # Enable HTTPS support
+      sed -i 's/#undef\tDOWNLOAD_PROTO_HTTPS/#define\tDOWNLOAD_PROTO_HTTPS/' config/general.h && \
       ARCH=$(uname -m | sed 's/aarch/arm/') && \
       # NOTE(elfosardo): warning should not be treated as errors by default
       NO_WERROR=1 make bin/undionly.kpxe "bin-$ARCH-efi/snponly.efi"


tuminoid:
/retest
CI should be better now.

tuminoid (commented):
LGTM, but 
/cc @Rozzii

defo89:
> I would recommend to deploy the ipxebuilder as an init container of the Ironic pod or just patch the Dockerfile downstream
We use this in production - problem with this approach would be delayed pod readiness on reschedule/failure of the node previously running this pod.

@Rozzii Would you consider an option to override the IPXE image location per env var like its done with `IPA_BASEURI` env? E.g. I could provide them via volume. My issue atm is that `ironic-image` provides all means to leverage TLS (I could do it for ironic and httpd) and only this bit is missing.

Rozzii:
> > I would recommend to deploy the ipxebuilder as an init container of the Ironic pod or just patch the Dockerfile downstream
> > We use this in production - problem with this approach would be delayed pod readiness on reschedule/failure of the node previously running this pod.
> 
> @Rozzii Would you consider an option to override the IPXE image location per env var like its done with `IPA_BASEURI` env? E.g. I could provide them via volume. My issue atm is that `ironic-image` provides all means to leverage TLS (I could do it for ironic and httpd) and only this bit is missing.

I plan to document this topic a bit more (or a lot more) in the future but I think we have such variables:

`IPXE_CUSTOM_FIRMWARE_DIR` can be used to specify where should the dnsmasq container find the custom firmware files
`IPXE_SSL_PROTOCOL` is should set the SSL/TLS versions that
`IPXE_CERT_FILE` the should be the same as the one embedded in the firmware
`IPXE_KEY_FILE` same as above
`IPXE_TLS_PORT` by default 8084 this is the port that the pxe firmware will use to call "home" and download the pxe files and the ramdisk

Rozzii:
Would it be okay to close this ticket @defo89 ?

defo89:
@Rozzii sure, we can close it. Thanks!

PR #519, opened 2024-06-11
🌱 increase ironic-image build timeout to 3600s

tuminoid:
Ironic-image build locally, with a fast disk, fast CPU, and fast net takes 500-600s, depending on pre-pulls, repository speeds etc. New cloud has none of those reliably, so it times out at 1000s far too often. Also scheduling time is considered when images are built in Jenkins, and often scheduling takes a long time. Thanks to #505  we get notifications of these failures now.

- Ironic image: 1000s -> 3600s (build time 500-600s, less than 2x safety)
- Sushy-tools: keep 1000s (build time 150s, 6x safety)
- VMBC: keep 1000s (build time 100s, 10x safety)

I will cherry-pick this to release branches as well.

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -31,7 +31,7 @@ jobs:
           {
             "BUILD_CONTAINER_IMAGE_GIT_REFERENCE": "${{ github.ref }}"
           }
-        job_timeout: "1000"
+        job_timeout: "3600"
     - name: build sushy-tools image
       uses: toptal/jenkins-job-trigger-action@137fff703dd260b52b53d3ba1960396415abc568 # 1.0.2
       with:


tuminoid:
/override metal3-centos-e2e-integration-test-main  metal3-ubuntu-e2e-integration-test-main 
GH action.

NymanRobin (approved):
The change is good, just to note there has been some problems scheduling the VMs also for the jobs, like this took 15 hour when you account for scheduling
https://jenkins.nordix.org/job/metal3_ironic_container_image_building/62/

Maybe these jobs could be moved from Jenkins to github action runners in the future to improve reliability

Rozzii:
> The change is good. However, please note that there have been some issues with VM scheduling for the jobs. For instance, this job took 15 hours when scheduling time is included: [Build #62](https://jenkins.nordix.org/job/metal3_ironic_container_image_building/62/).
> 
> Maybe these jobs could be moved from Jenkins to github action runners in the future to improve reliability

WOW that is not expected IMO, anyways yeah I would guess container image building could be a github action, I think that was considered originally too.

tuminoid:
> The change is good. However, please note that there have been some issues with VM scheduling for the jobs. For instance, this job took 15 hours when scheduling time is included: [Build #62](https://jenkins.nordix.org/job/metal3_ironic_container_image_building/62/).
> 
> Maybe these jobs could be moved from Jenkins to github action runners in the future to improve reliability

Yes agreed. We can see that after scheduling, the build itself took almost 21minutes (~1250 sec), which would've timeouted as well, so this fix is needed. It would then leave 30mins for scheduling as it is counted as well. Similarly for sushy/vmbc, the safety allows ~10 minutes of scheduling. We should be able to accomplish that in Jenkins or with runners. Jenkins scheduling we can fix ourselves, but runners would be at mercy of CNCF and other runner users.

PR #525, opened 2024-06-21
:seedling: Update Makefile

elfosardo:
Introduce variable to set build container engine.
Build with podman by default.


Rozzii (commented):
/lgtm

PR #532, opened 2024-07-23
🐛 Update httpd.conf.j2 

### LINKED ISSUE #531: ironic-httpd config does not work on systems with ipv6 disabled
The current `Listen` directive for all interfaces does not work on systems that have ipv6 disabled. Here is an example log extracted from a deployment I've tried to get working:

```bash
+ exec /usr/sbin/httpd -DFOREGROUND
[Tue Jul 23 08:29:14.110334 2024] [core:crit] [pid 1:tid 1] (97)Address family not supported by protocol: AH00078: alloc_listener: failed to get a socket for ::
AH00526: Syntax error on line 2 of /etc/httpd/conf/httpd.conf:
Listen setup failed
```

owenthomas17:
The current `Listen` directive does not work on systems that have ipv6 disabled. The Apache Listen directive supports `Listen <port>` syntax at which point it seems to be able to listen correctly on both dual stack and ipv4 only systems. See https://httpd.apache.org/docs/2.4/bind.html for more details.

Fixes #531

### DIFF
--- a/ironic-config/httpd.conf.j2
+++ b/ironic-config/httpd.conf.j2
@@ -1,6 +1,6 @@
 ServerRoot "/etc/httpd"
 {%- if env.LISTEN_ALL_INTERFACES | lower == "true" %}
-Listen [::]:{{ env.HTTP_PORT }}
+Listen {{ env.HTTP_PORT }}
 {% else %}
 Listen {{ env.IRONIC_URL_HOST }}:{{ env.HTTP_PORT }}
 {% endif %}


ironic-config/httpd.conf.j2
```diff
@@ -1,6 +1,6 @@
 ServerRoot "/etc/httpd"
 {%- if env.LISTEN_ALL_INTERFACES | lower == "true" %}
-Listen [::]:{{ env.HTTP_PORT }}
+Listen {{ env.HTTP_PORT }}
```
mboukhalfa, line 3: Is this still listening on all interfaces ?

mboukhalfa: looks like from the Apache docs

owenthomas17: Yes, according to the apache documentation:
image

mboukhalfa:
Can you please sign the commit to fix the DCO ? and run tests ?

mboukhalfa:
failed because of http issue when installing packages 

mboukhalfa:
@iurygregory Can you please re-lgtm I do not have this permission on this repo ?
cc @Rozzii

PR #535, opened 2024-08-06
:sparkles: Enable FIPS mode for IPA if system is in FIPS mode

elfosardo:
If FIPS is enabled in the hosts we should also run IPA in FIPS mode. It is possible to enable FIPS directly at kernel level using the fips option, determining the FIPS status for example from the cryptographic module and specifically the
/proc/sys/crypto/fips_enabled file; if the file contains 1 then the system is in FIPS mode, if it contains 0 the FIPS algorithms are disabled.
Therefore the value of the fips kernel option is 0 (default) if FIPS is disabled, or 1 if enabled.

### DIFF
--- a/scripts/configure-ironic.sh
+++ b/scripts/configure-ironic.sh
@@ -86,6 +86,11 @@ mkdir -p /shared/ironic_prometheus_exporter
 
 configure_json_rpc_auth
 
+if [[ -f /proc/sys/crypto/fips_enabled ]]; then
+    ENABLE_FIPS_IPA=$(cat /proc/sys/crypto/fips_enabled)
+    export ENABLE_FIPS_IPA
+fi
+
 # The original ironic.conf is empty, and can be found in ironic.conf_orig
 render_j2_config /etc/ironic/ironic.conf.j2 /etc/ironic/ironic.conf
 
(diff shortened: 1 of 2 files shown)


scripts/configure-ironic.sh
```diff
@@ -86,6 +86,9 @@ mkdir -p /shared/ironic_prometheus_exporter
 
 configure_json_rpc_auth
 
+ENABLE_FIPS_IPA=$(cat /proc/sys/crypto/fips_enabled)
```
tuminoid, line 89: Is this file guaranteed to exist on Centos? It does not exist on my Ubuntu machine, which would result in empty variable, instead of 0 or 1.

elfosardo: this should be guarantee existing in FIPS enabled kernels, including ubuntu, since at least 4 years
I don't recall exactly the kernel version, but I'm quite sure the file exists since at least RHEL/CENTOS 7.6 and Ubuntu 16.0.4

elfosardo: found it, a bit older than 4 years :)
https://www.kernelconfig.io/config_crypto_fips

tuminoid: It still isn't present on Ubuntu 22.04 by default tho. Obviously my desktop kernel isn't FIPS enabled.

Whats the behavior of the flag if it isn't 0 or 1 but empty? `simple_strtol` used by fips.c isn't well documented.

elfosardo:
some issue with CI?

scripts/configure-ironic.sh
elfosardo: probably only very latest kernels are all compiled with fips support by default
I guess I can add a test for the file, just to be on the safe side

tuminoid (commented):
/lgtm

PR #536, opened 2024-08-07
:seedling: Disable pip cache

elfosardo:
We don't really need to keep pip cache as we don't install more packages with pip after the initial build.
It should reduce the size of the final image of about 50MB.

### DIFF
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -33,7 +33,7 @@ if [[ "$INSTALL_TYPE" == "source" ]]; then
     # NOTE(dtantsur): pip is a requirement of python3 in CentOS
     # shellcheck disable=SC2086
     dnf install -y python3-pip $BUILD_DEPS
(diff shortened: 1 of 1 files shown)


prepare-image.sh
```diff
@@ -104,5 +104,8 @@ if [[ -n "${PATCH_LIST:-}" ]]; then
 fi
 rm -f /bin/patch-image.sh
 
+# purge pip cache
+python3 -m pip cache purge
```
tuminoid, line 108: How many places do we call pip in? If not many, it would be even better to call pip install with `--no-cache-dir` to not create any caches to begin with.

elfosardo: yeah, I was actually going to change it to that as we run pip only twice and only when we install ironic related packages

tuminoid (commented):
Good improvement on the image size!

/lgtm

PR #539, opened 2024-08-19
🌱 Fix syntax in build images github workflow

NymanRobin:
The syntax error in build images workflow caused all jobs to fail. These could probably be automatically caught in the future


tuminoid (commented):
/lgtm

tuminoid:
/override metal3-centos-e2e-integration-test-main  metal3-ubuntu-e2e-integration-test-main

PR #540, opened 2024-08-26
🐛runlogwatch.sh fix CPU usage and partial files

steveb:
Currently runlogwatch.sh runs in a tight loop when there are no files,
consuming CPU. Also there have been reports of truncated logs being
outputted which would be explained by ironic not writing these files
atomically [1]. This change improves this script by:
- moving the sleep to the top of the loop so it is called on every
  iteration
- change the file glob to a `find` command which includes `-mmin +0.1`
  so the file is only processed when it is at least 6 seconds old

### DIFF
--- a/scripts/runlogwatch.sh
+++ b/scripts/runlogwatch.sh
@@ -4,15 +4,14 @@
 LOG_DIR="/shared/log/ironic/deploy"
 
 while :; do
-    if ! ls "${LOG_DIR}"/*.tar.gz 1> /dev/null 2>&1; then
-        continue
-    fi
+    sleep 5
 
-    for fn in "${LOG_DIR}"/*.tar.gz; do
+    while read -r fn; do
+        echo
         echo "************ Contents of $fn ramdisk log file bundle **************"
         tar -xOzvvf "$fn" | sed -e "s/^/$(basename "$fn"): /"
         rm -f "$fn"
-    done
+    # find all *.tar.gz files which are older than six seconds
+    done < <(find "${LOG_DIR}" -mmin +0.1 -type f -name "*.tar.gz")
 
-    sleep 5
 done


tuminoid:
```console
[2024-08-27T07:15:52.536Z] dnf install -y dnsmasq dosfstools httpd httpd-tools inotify-tools ipcalc ipmitool iproute mod_ssl procps python3-jinja2 qemu-img sqlite syslinux-nonlinux util-linux xorriso

[2024-08-27T07:15:52.914Z] Last metadata expiration check: 0:02:06 ago on Tue Aug 27 07:13:46 2024.
```
It doesn't seem to be able to install the package you added.

main-packages-list.txt
```diff
@@ -2,6 +2,7 @@ dnsmasq
 dosfstools
 httpd
 httpd-tools
+inotify-tools
```
elfosardo, line 5: inotify-tools is already installed from epel

elfosardo (dismissed):

steveb:
Actually I'm going to go back to a polling approach, but fix the cpu churn issue, and also add some mitigation for partially written tar files

scripts/runlogwatch.sh
```diff
@@ -4,15 +4,13 @@
...
+    sleep 5
 
-    for fn in "${LOG_DIR}"/*.tar.gz; do
+    while read -r fn; do
         echo "************ Contents of $fn ramdisk log file bundle **************"
         tar -xOzvvf "$fn" | sed -e "s/^/$(basename "$fn"): /"
         rm -f "$fn"
-    done
+    # find all *.tar.gz files which are older than six seconds
+    done < <(find ${LOG_DIR} -mmin +0.1 -type f -name "*.tar.gz")
```
tuminoid, line 14:
```suggestion
    done < <(find "${LOG_DIR}" -mmin +0.1 -type f -name "*.tar.gz")
```

tuminoid (commented):
Nit

scripts/runlogwatch.sh
steveb: Done

scripts/runlogwatch.sh
```diff
@@ -4,15 +4,13 @@
...
 
 while :; do
-    if ! ls "${LOG_DIR}"/*.tar.gz 1> /dev/null 2>&1; then
-        continue
-    fi
+    sleep 5
 
-    for fn in "${LOG_DIR}"/*.tar.gz; do
+    while read -r fn; do
         echo "************ Contents of $fn ramdisk log file bundle **************"
```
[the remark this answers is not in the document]
steveb: Done

blallau:
Hi Steve,
I was working on the same issue without seeing that you already work on it.
In order to be sure that archive file is fully ready, I made use of the inotify close_write event.
You can take a look at https://github.com/metal3-io/ironic-image/pull/548
regards,

steveb:
https://github.com/metal3-io/ironic-image/pull/548 looks nice and clean. Seems like a good alternative to this change

tuminoid:
/hold
I'm holding this as concensus seems to be to use inotifywait anyways?

steveb:
Closing my PR to make it clear which approach I think we should use :)

PR #548, opened 2024-08-30
🐛Fix CPU burning on ironic-log-watch when no ramdisk file is present

### LINKED ISSUE #547: ironic-log-watch: CPU at full capacity in case of no ramdisk file present
In the ironic-log-watch container when no ramdisk file is present in /shared/log/ironic/deploy directory an infinite loop lead cpu to be at full capacity all the time.

blallau:
Use inotifywait instead of an infinite loop.

Fixes #547

### DIFF
--- a/scripts/runlogwatch.sh
+++ b/scripts/runlogwatch.sh
@@ -3,16 +3,9 @@
 # Ramdisk logs path
 LOG_DIR="/shared/log/ironic/deploy"
 
-while :; do
-    if ! ls "${LOG_DIR}"/*.tar.gz 1> /dev/null 2>&1; then
-        continue
-    fi
-
-    for fn in "${LOG_DIR}"/*.tar.gz; do
-        echo "************ Contents of $fn ramdisk log file bundle **************"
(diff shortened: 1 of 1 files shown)


tuminoid:
We have https://github.com/metal3-io/ironic-image/pull/540 already addressing similar issue, can you collaborate on that PR?

tuminoid:
Please fix shellcheck issues, preferably use curlys around variables as well, and when linting is green, trigger integration tests.

tuminoid:
> /retest

Retest won't run tests that are not run in the first place. 
/test metal3-centos-e2e-integration-test-main  metal3-ubuntu-e2e-integration-test-main

tuminoid:
Can you rebase, so the fix we have merged is included. Otherwise CI won't work.

tuminoid:
> Can you rebase, so the fix we have merged is included. Otherwise CI won't work.

The official fix is in, so this should work now.
/retest

tuminoid (commented):
/lgtm

derekhiggins:
/approve
thanks

PR #569, opened 2024-09-24
:bug: Pin setuptools and pin upper-constraints

elfosardo:
Usually the setuptools installed by the system is stable but
a bit too old compared to the packages we use for ironic,
so pinning a more recent version helps avoiding
errors and incompatibilities.

Also pin upper-constraints since we forgot!


tuminoid:
Oh we had forgotten to pin the requirements, ouch :)

tuminoid (commented):
/lgtm

Also mention of upper constrain pinning in the PR title.

elfosardo:
@dtantsur  can you please have a look?

PR #570, opened 2024-09-24
:bug: Pin setuptools during image build

elfosardo:
Usually the setuptools installed by the system is stable but a bit too old compared to the packages we use for ironic, so pinning a more recent version helps avoiding errors and incompatibilities.


elfosardo:
I guess I'll cherry-pick this manually for 25.0 and 24.1

PR #572, opened 2024-09-25
:bug: Pin setuptools during image build

elfosardo:
Usually the setuptools installed by the system is stable but a bit too old compared to the packages we use for ironic, so pinning a more recent version helps avoiding errors and incompatibilities.

(cherry picked from commit 82692ba7683ec1d48a67aaaaf6d8bf5496177c1d)


tuminoid (commented):
/lgtm

Issue #529, opened 2024-07-17
Support CentOS Stream 10


elfosardo:
CentOS Stream 10 is not officially out yet but bits are already available https://mirror.stream.centos.org/10-stream/ and can be used for testing.

Important changes include:

kernel 6.10 or more recent
Python 3.12

This heavily impacts ironic and, as a consequence, the ironic-image

Before moving forward with this some changes are needed in the ironic project and the entire Openstack ecosystem.
For example while Python 3.12 is already being tested, it's not officially supported yet, and there's still no plan to start using CS10 in CI, although discussions on migrating some important components are ongoing since a while, for example removing eventlet https://review.opendev.org/c/openstack/governance/+/902585

This issue should be used to track the progression on the CS10 migration for ironic-image, while keeping an eye on the progress in the ironic project itself.

/kind feature

elfosardo:
The Epoxy development cycle for Openstack has just started and includes support for Python 3.12, which is now actively tested

elfosardo:
I've opened a bug to ask inotify-tools build for EPEL10 https://bugzilla.redhat.com/show_bug.cgi?id=2331998

elfosardo:
ironic upstream CI is still based on CS9 and AFAIK there is no will to migrate it during this development cycle, but it could change depending on the priorities
my suggestion is to hold on to this until we have some movement in ironic upstream

elfosardo:
community has decided to move forward and test both CS9 and CS10 in parallel
I'll propose a configuration for the ironic-image that could work with both versions

tuminoid:
> @tuminoid I'd rather keep this open until we actually switch to CS10 by default

Can we make another issue about making it the default? It is easier to track and reference things when they aren't used for many things over long span of time.

Rozzii:
I am in favor of closing this and creating a new Issue related to making it default.

elfosardo:
https://github.com/metal3-io/ironic-image/issues/697

PR #581, opened 2024-10-18
:sparkles: Install shim-pkg explicitly and add support for CentOS Stream 10

elfosardo:
This simple change allows us to build ironic containers using
CS10 base images, adding that as supported OS.
This has been tested locally and in CI.

related issue https://github.com/metal3-io/ironic-image/issues/529

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 This repo contains the files needed to build the Ironic images used by Metal3.
 
+Supported base images are CentOS Stream 9 and CentOS Stream 10.
+
 ## Build Status
 
 [![CLOMonitor](https://img.shields.io/endpoint?url=https://clomonitor.io/api/projects/cncf/metal3-io/badge)](https://clomonitor.io/projects/cncf/metal3-io)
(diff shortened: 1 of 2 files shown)


elfosardo:
this will stay on hold until adoption in ironic upstream that should happen next year

tuminoid (commented):
It builds fine via --build-arg.

How are we planning to test this? Do we need to add a build jobs for both stream9 and 10 so it does not break? How about e2e?

elfosardo:
@tuminoid we discussed this a couple of times during the weekly meeting and I'm going to bring it up again tomorrow, but the consensus was to have build jobs for both CS9 and CS10 (really just verify that the image builds), plus periodic e2e jobs for both, at least until we support both versions, which I expect it will be until at least early next year
we could also add another centos integration job to run for each PR, but maybe it's too much?

tuminoid:
> @tuminoid we discussed this a couple of times during the weekly meeting and I'm going to bring it up again tomorrow, but the consensus was to have build jobs for both CS9 and CS10 (really just verify that the image builds), plus periodic e2e jobs for both, at least until we support both versions, which I expect it will be until at least early next year we could also add another centos integration job to run for each PR, but maybe it's too much?

OK, yeah OK, now I recall. Let's discuss more tomorrow and finalize the plan. We have plenty of hardware resources now so running more e2e is not a problem, if we just can keep the tests stable.

Rozzii (commented):
/lgtm

PR #584, opened 2024-10-31
:bug: improve upper constraint handling

Rozzii:
This commit:


Rozzii:
If this gets merged I will cherry pick this to other branches.

tuminoid (commented):
/lgtm

This must be backported to release branches as well, where it actually matters.

tuminoid:
@Rozzii will make manual cherry-picks to all maintained release branches.

We also need to consider should we make patch release after fix is merged, as this might/will change the Python dependencies quite drastically. @elfosardo thoughts?

PR #588, opened 2024-11-06
🐛 improve upper constraints handling

Rozzii:
This commit:


tuminoid:
For reference: this is manual cherry-pick of https://github.com/metal3-io/ironic-image/pull/584

tuminoid (commented):
/lgtm

PR #589, opened 2024-11-06
🐛 improve upper constraints handling

Rozzii:
This commit:


tuminoid:
For reference: this is manual cherry-pick of https://github.com/metal3-io/ironic-image/pull/584

tuminoid (commented):
/lgtm

tuminoid:
I think the deps have rot already for this version:

```console
[2024-11-08T09:06:46.846Z] INFO: pip is looking at multiple versions of <Python from Requires-Python> to determine which version is compatible with other requirements. This could take a while.
[2024-11-08T09:06:46.846Z] INFO: pip is looking at multiple versions of gunicorn to determine which version is compatible with other requirements. This could take a while.
[2024-11-08T09:06:46.846Z] INFO: pip is looking at multiple versions of crudini to determine which version is compatible with other requirements. This could take a while.
[2024-11-08T09:06:46.846Z] Collecting crudini
[2024-11-08T09:06:46.846Z]   Downloading crudini-0.9.4-py2.py3-none-any.whl (21 kB)
[2024-11-08T09:06:46.846Z] ERROR: Cannot install -r /tmp/ironic-source-list-final (line 5) and sushy because these package versions have conflicting dependencies.
[2024-11-08T09:06:46.846Z] 
[2024-11-08T09:06:46.846Z] The conflict is caused by:
[2024-11-08T09:06:46.846Z]     The user requested sushy
[2024-11-08T09:06:46.846Z]     ironic 24.0.1.dev10 depends on sushy>=4.7.0
[2024-11-08T09:06:46.846Z]     The user requested (constraint) sushy===4.5.2
[2024-11-08T09:06:46.846Z] 
[2024-11-08T09:06:46.846Z] To fix this you could try to:
[2024-11-08T09:06:46.846Z] 1. loosen the range of package versions you've specified
[2024-11-08T09:06:46.846Z] 2. remove package versions to allow pip attempt to solve the dependency conflict
[2024-11-08T09:06:46.846Z] 
[2024-11-08T09:06:46.846Z] ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
[2024-11-08T09:06:49.219Z] Error: building at STEP "RUN prepare-image.sh &&   rm -f /bin/prepare-image.sh": while running runtime: exit status 1
```

Rozzii:
@elfosardo  is it okay if I uplift the constraint ?

tuminoid:
> @elfosardo is it okay if I uplift the constraint ?

Can we do quick check if its already been uplifted in the upstream? It might've changed since the contsraints file was saved here.

Rozzii:
@tuminoid at least  AFAIK in the upstream branch the requirement is clear https://opendev.org/openstack/ironic/src/branch/bugfix/24.0/requirements.txt#L49 so there is not much question about it IMO. It is not possible to build it under 4.7.

Rozzii:
> @tuminoid at least AFAIK in the upstream branch the requirement is clear https://opendev.org/openstack/ironic/src/branch/bugfix/24.0/requirements.txt#L49 so there is not much question about it IMO. It is not possible to build it under 4.7.

Now I have checked the original release tag too and even at that time it was 4.7 https://opendev.org/openstack/ironic/src/tag/24.0.0/requirements.txt so I think setting it lower was a mistake initially too just the bug served as a "workaround".

tuminoid (commented):
/lgtm

Rozzii:
Seems it will be multiple of the dependencies, I will push the fixes.

tuminoid:
/hold
More fixes needed.

Rozzii:
Should be fine now.
/hold cancel

elfosardo:
@Rozzii  sorry for ignoring this! It got completely  out of my radar after I approved it.
my understanding is that the initial UC pinning was correct, is that right?
Also please let me know if you need help troubleshooting

Rozzii:
> @Rozzii sorry for ignoring this! It got completely out of my radar after I approved it. my understanding is that the initial UC pinning was correct, is that right? Also please let me know if you need help troubleshooting

AFAIK originally this has worked, I assume the change of the UC pinning is related to 24.1 release.

tuminoid (commented):
Good job finding working combination!

/lgtm

PR #591, opened 2024-11-10
🌱 Update Dockerfile

Mionsz:

Update Dockerfile, add basic labels and speed-up the build process:
- Dockerfile compliant first line metadata
- use `curl` instead of `git` for repository download
- use build-time bash shell with `-exo pipefail` for error catching
- add proper labeling of resulting image
- further improve image size by reducing layers number using one RUN command.
- missing `ENTRYPOINT` added
- missing `USER` added
- other minor improvements

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
-ARG BASE_IMAGE=quay.io/centos/centos:stream9
+# syntax=docker/dockerfile:1
+
+ARG BASE_IMAGE=quay.io/centos/centos:stream9@sha256:e5fdd83894773a25f22fbdf0b5253c63677d0cbaf8d3a8366b165a3ef5902964
 
 ## Build iPXE w/ IPv6 Support
 ## Note: we are pinning to a specific commit for reproducible builds.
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -61,11 +68,17 @@ COPY ironic-config/httpd-modules.conf /etc/httpd/conf.modules.d/
...
-RUN configure-nonroot.sh && \
-  rm -f /bin/configure-nonroot.sh
+  configure-nonroot.sh && \
+  dnf remove -y sqlite && \
+  dnf clean all && \
+  rm -rf /bin/configure-nonroot.sh /bin/prepare-image.sh /var/cache/{yum,dnf}/*
+
+USER ironic
+WORKDIR /
+ENTRYPOINT ["/bin/bash"]
```
dtantsur, line 84: The entrypoint does not look right.

Dockerfile
```diff
@@ -7,25 +9,33 @@ ARG BASE_IMAGE=quay.io/centos/centos:stream9
...
 RUN prepare-efi.sh centos
 
 FROM $BASE_IMAGE
 
-ENV PKGS_LIST=main-packages-list.txt
+LABEL org.opencontainers.image.url="https://github.com/metal3-io/ironic-image"
+LABEL org.opencontainers.image.title="Metal3 Ironic Container"
+LABEL org.opencontainers.image.description="Container image to run OpenStack Ironic as part of Metal³"
+LABEL org.opencontainers.image.documentation="https://github.com/metal3-io/ironic-image/blob/main/README.md"
+LABEL org.opencontainers.image.version="v26.0.1"
```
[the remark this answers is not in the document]
tuminoid: We should not have anything but latest in main tho

Dockerfile
```diff
@@ -7,25 +9,33 @@ ARG BASE_IMAGE=quay.io/centos/centos:stream9
 FROM $BASE_IMAGE AS ironic-builder
 
 ARG IPXE_COMMIT_HASH=119c415ee47aaef2717104fea493377aa9a65874
+ARG MAKEFLAGS="-j100"
```
tuminoid, line 12: A 100 jobs, I don't think thats reasonable default.

Dockerfile
```diff
@@ -7,25 +9,33 @@ ARG BASE_IMAGE=quay.io/centos/centos:stream9
...
 COPY prepare-efi.sh /bin/
 RUN prepare-efi.sh centos
 
 FROM $BASE_IMAGE
 
-ENV PKGS_LIST=main-packages-list.txt
+LABEL org.opencontainers.image.url="https://github.com/metal3-io/ironic-image"
+LABEL org.opencontainers.image.title="Metal3 Ironic Container"
+LABEL org.opencontainers.image.description="Container image to run OpenStack Ironic as part of Metal³"
+LABEL org.opencontainers.image.documentation="https://github.com/metal3-io/ironic-image/blob/main/README.md"
```
tuminoid, line 33: Documentation would maybe better point to our user-guide?

Dockerfile
elfosardo: this is also probably not needed

Dockerfile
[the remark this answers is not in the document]
elfosardo: this is ok for released branches, as Dmitry said probably need to add this when we cut a new branch
in main we should just have latest

Dockerfile
elfosardo: no need to set MAKEFLAGS at all

Dockerfile
elfosardo: I agree,  let's point to the  user guide

PR #600, opened 2024-12-05
seedling: Build ironicclient using debian based python image

elfosardo:
This should reduce the size of the container from 340MB to 230MB

### DIFF
--- a/resources/ironic-client/Dockerfile
+++ b/resources/ironic-client/Dockerfile
@@ -1,13 +1,14 @@
-FROM quay.io/centos/centos:stream9
+ARG BASE_IMAGE=docker.io/library/python:3.9-slim-bookworm
+
+FROM $BASE_IMAGE
 
 # Help people find the actual baremetal command
 COPY scripts/openstack /usr/bin/openstack
 
-RUN dnf install -y python3 python3-pip genisoimage && \
-    pip install python-ironicclient --prefix /usr --no-cache-dir && \
-    chmod +x /usr/bin/openstack && \
-    dnf update -y && \
-    dnf clean all && \
-    rm -rf /var/cache/{yum,dnf}/*
+RUN apt-get update && \
+    apt-get install -y genisoimage && \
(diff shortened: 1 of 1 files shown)


resources/ironic-client/Dockerfile
```diff
@@ -1,13 +1,14 @@
...
 COPY scripts/openstack /usr/bin/openstack
 
-RUN dnf install -y python3 python3-pip genisoimage && \
-    pip install python-ironicclient --prefix /usr --no-cache-dir && \
-    chmod +x /usr/bin/openstack && \
-    dnf update -y && \
-    dnf clean all && \
-    rm -rf /var/cache/{yum,dnf}/*
+RUN apt-get update && \
+    apt-get install -y genisoimage && \
```
tuminoid, line 9: Add `apt-get upgrade -y && \`  here to match centos image.

resources/ironic-client/Dockerfile
```diff
@@ -1,13 +1,14 @@
...
+
+FROM $BASE_IMAGE
 
 # Help people find the actual baremetal command
 COPY scripts/openstack /usr/bin/openstack
 
-RUN dnf install -y python3 python3-pip genisoimage && \
-    pip install python-ironicclient --prefix /usr --no-cache-dir && \
-    chmod +x /usr/bin/openstack && \
-    dnf update -y && \
```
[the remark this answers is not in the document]
elfosardo: that's right, this is to match the vbmc and sushy-tools Dockerfiles that don't have `apt-get upgrade`

resources/ironic-client/Dockerfile
elfosardo: I believe it's not needed

tuminoid: We can live without as well.

tuminoid:

tuminoid (commented):
/lgtm

PR #603, opened 2024-12-12
:sparkles: Make deep image inspection configurable

Rozzii:
This PR:

  - Implements the support for disabling the deep_image_inspection via environment variable.

This new functionality is needed because e.g right now the Ironic&IPA has issues running deep image inspection on e.g. live-iso images.

### DIFF
--- a/ironic-config/ironic.conf.j2
+++ b/ironic-config/ironic.conf.j2
@@ -92,6 +92,9 @@ deploy_kernel = file://{{ env.IRONIC_DEFAULT_KERNEL }}
 {% if env.IRONIC_DEFAULT_RAMDISK is defined %}
 deploy_ramdisk = file://{{ env.IRONIC_DEFAULT_RAMDISK }}
 {% endif %}
+{% if env.DISABLE_DEEP_IMAGE_INSPECTION | lower == "true" %}
+disable_deep_image_inspection = True
+{% endif %}
 
 [database]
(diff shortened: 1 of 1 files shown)


Rozzii:
Related BMO Issue:  https://github.com/metal3-io/baremetal-operator/issues/2110

tuminoid (commented):
/lgtm

We need this so we can remove the pinning of ironic to 26.0 in BMO, and can keep testing Ironic main. We also need this backported to 27.0 where the deep inspection issue is also happening.

/cherry-pick release-27.0

tuminoid:
/cherry-pick release-27.0
Wake up bot.

iurygregory (approved):

dtantsur:
@Rozzii for the future, please don't introduce a variable each time we need to tweak something: you can use stuff like `OS_DEFAULT__DISABLE_DEEP_IMAGE_INSPECTION=False`. That would also avoid the need to backport.

I'd prefer this PR to get reverted once we fix the live ISO problem for good.

tuminoid:
> @Rozzii for the future, please don't introduce a variable each time we need to tweak something: you can use stuff like `OS_DEFAULT__DISABLE_DEEP_IMAGE_INSPECTION=False`. That would also avoid the need to backport.
> 
> I'd prefer this PR to get reverted once we fix the live ISO problem for good.

Do we have timeline for the actual live ISO fix? Asking if the cherry-pick for backport should be merged or not.

Issue #605, opened 2024-12-18
inotify-tools is unmaintained


elfosardo:
inotify-tools last release is from september 2023 and last update in the repo is 7 months old (May 2024), it's not present in EPEL10 and not updated in EPEL9
we need to find a replacement or rewrite inotifywait mechanism in some way
we use inotifywait from inotify-tools in multiple places:
https://github.com/metal3-io/ironic-image/blob/main/scripts/runironic#L16
https://github.com/metal3-io/ironic-image/blob/main/scripts/runhttpd#L76
https://github.com/metal3-io/ironic-image/blob/main/scripts/runhttpd#L84
https://github.com/metal3-io/ironic-image/blob/main/scripts/runlogwatch.sh#L6

also we should not accept changes that use inotifywait anymore

tuminoid:
I quickly checked alternatives and very few options in EPEL in general. fswatch, watchman etc all missing in EPEL9 as well. incron is there but it is completely different usage and bad fit for containers.

I might be easiest to just make a custom Python snippet to simulate inotify-tools. We already use similar for processing jinja2 templates.

dtantsur:
Bonus points for avoiding EPEL entirely (python3-inotify seems to be in the main repo).

PR #607, opened 2024-12-23
:seedling: Use pynotify instead of inotify-tools

elfosardo:
The inotify-tools last release is from september 2023
and last update in the repo is 7 months old (May 2024).
The inotify-tools package is not present in the EPEL10 repo
anymore while python3-inotify is in a native repository.
This change allows us to remove dependency from EPEL entirely.

Closes https://github.com/metal3-io/ironic-image/issues/605

### DIFF
--- a/main-packages-list.txt
+++ b/main-packages-list.txt
@@ -7,6 +7,7 @@ ipmitool
 iproute
 mod_ssl
 procps
+python3-inotify
 python3-jinja2
 qemu-img
 sqlite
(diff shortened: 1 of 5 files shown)


dtantsur (changes requested):
The change looks good. There are a few other places where inotify is used, could you update them as well?

Also note that pyinotify is even more ancient: its last release was in 2015. I guess it still works well if it's included in the repos.

dtantsur (approved):

tuminoid:
Good thing about Python inotify packages is that they all just implement ctypes wrappers for libc's functionality and do not try to reinvent the wheel. Hence most of the fixes, barring language incompatibilities down the road, are coming from OS packages, and it is not as relevant to have the python package updating all the time.

PR #612, opened 2025-01-26
🐛 Drop quiet option of grep to avoid race condition with pipefail

### LINKED ISSUE #611: ironic-common.sh: `<cmd> | grep -q` with `set -o pipefail` causes race condition (SIGPIPE)
The grep -q closes input pipe immediately if any match is found, while the command on the left side will exit with SIGPIPE when it tries to write to a closed pipe, causing the Metal3 Pod to get stuck in CrashLoopBackOff sometimes.

This is particularly likely to occur when the command on the left side of the pipe produces a large amount of output. Here is a simple example to reproduce the problem:
$ (set -o pipefail; lspci -vvv | grep -qi "sata"; echo $?)
141

fenggw-fnst:
The grep -q closes input pipe immediately if any match is found, while the command on the left side will exit with SIGPIPE when it tries to write to a closed pipe, causing the Metal3 Pod to get stuck in CrashLoopBackOff sometimes.

**Which issue(s) this PR fixes**:
Fixes #611

### DIFF
--- a/scripts/ironic-common.sh
+++ b/scripts/ironic-common.sh
@@ -19,13 +19,13 @@ get_provisioning_interface()
     local interface="provisioning"
 
     if [[ -n "${PROVISIONING_IP}" ]]; then
-        if ip -br addr show | grep -qi " ${PROVISIONING_IP}/"; then
+        if ip -br addr show | grep -i " ${PROVISIONING_IP}/" &>/dev/null; then
             interface="$(ip -br addr show | grep -i " ${PROVISIONING_IP}/" | cut -f 1 -d ' ' | cut -f 1 -d '@')"
         fi
     fi
 
     for mac in ${PROVISIONING_MACS//,/ }; do
-        if ip -br link show up | grep -qi "$mac"; then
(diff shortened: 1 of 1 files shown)


tuminoid (approved):

Rozzii (commented):
/cc @elfosardo

tuminoid:
/retest
Network flakes.

tuminoid:
This has now come up here at least two times, I don't see this come up elsewhere. 

```text
[2025-01-27T12:11:02.168Z]   [FAILED] Unexpected error:
[2025-01-27T12:11:02.168Z]       <*errors.Error | 0xc001313e00>: 
[2025-01-27T12:11:02.168Z]       unable to find field ".spec.ipAddresses.0" in replacement target
[2025-01-27T12:11:02.168Z]       {
[2025-01-27T12:11:02.168Z]           Err: <*errors.errorString | 0xc000d32160>{
[2025-01-27T12:11:02.168Z]               s: "unable to find field \".spec.ipAddresses.0\" in replacement target",
[2025-01-27T12:11:02.168Z]           },
```

PR #620, opened 2025-02-05
🌱 Run the IrSO functional job on ironic-images changes

dtantsur:

### DIFF
--- a/.github/workflows/irso-functional.yml
+++ b/.github/workflows/irso-functional.yml
@@ -1,0 +1,50 @@
+name: IrSO Functional Tests
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    env:
+      LOGDIR: /tmp/logs
+    steps:
+    - name: Update repositories
+      run: sudo apt-get update
+    - name: Install podman
+      run: sudo apt-get install -y podman
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      with:
+        path: ironic-image
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      with:
+        repository: metal3-io/ironic-standalone-operator
+        path: ironic-standalone-operator
+    - name: Calculate go version
+      id: vars
+      run: echo "go_version=$(make -sC ironic-standalone-operator go-version)" >> $GITHUB_OUTPUT
+    - name: Set up Go
+      uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
+      with:
+        go-version: ${{ steps.vars.outputs.go_version }}
+    - name: Create a Kind cluster
+      uses: helm/kind-action@a1b0e391336a6ee6713a0583f8c6240d70863de3 # v1.12.0
+      with:
+        cluster_name: kind
+        config: ironic-standalone-operator/test/kind.yaml
+    - name: Prepare tests
+      run: cd ironic-standalone-operator && ./test/prepare.sh
+    - name: Run tests
+      run: ironic-image/hack/ci-e2e.sh
+      env:
+        IRSO_PATH: "${{ github.workspace }}/ironic-standalone-operator"
+    - name: Collect logs
+      run: ironic-standalone-operator/test/collect-logs.sh
+      if: always()
+    - name: Upload logs artifacts
+      uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+      with:
+        name: functional
+        path: /tmp/logs/*
+      if: always()
--- a/hack/ci-e2e.sh
+++ b/hack/ci-e2e.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+set -eux -o pipefail
+
+REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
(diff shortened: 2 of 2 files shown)


.github/workflows/functional.yml
```diff
@@ -0,0 +1,51 @@
+name: Functional
```
tuminoid, line 1: Can we have better name for this?

.github/workflows/functional.yml
```diff
@@ -0,0 +1,51 @@
...
+
+on:
+  push:
```
tuminoid, line 4: We don't want it on push? We want it to run in PR, when it is updated. For that, take the events list from BMO.

.github/workflows/functional.yml
```diff
@@ -0,0 +1,51 @@
...
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+
+jobs:
+  functional:
```
tuminoid, line 10: naming again.

.github/workflows/functional.yml
```diff
@@ -0,0 +1,51 @@
...
+    runs-on: ubuntu-latest
+    env:
+      LOGDIR: /tmp/logs
+    steps:
+    - name: Update repositories
+      run: sudo apt update
+    - name: Install podman
+      run: sudo apt install -y podman
+    - uses: actions/checkout@v4
```
tuminoid, line 19: This needs to be pinned.

.github/workflows/functional.yml
```diff
@@ -0,0 +1,51 @@
...
+      with:
+        path: ironic-image
+    - uses: actions/checkout@v4
```
tuminoid, line 22: Same

tuminoid (commented):
Some nitting.

When finished, most of the changes should be applied in workflow in IRSO as well as this is mostly a copy from there.

.github/workflows/functional.yml
tuminoid: Something referring to IRSO would be enough to help distinguish it, maybe simply "IrSO functional test" ?

dtantsur: Ah, you want to see IrSO mentioned? Great, will update.

dtantsur:

(does not involve integration tests in any way)

dtantsur:
@tuminoid https://github.com/metal3-io/ironic-standalone-operator/pull/123 is the follow-up

.github/workflows/irso-functional.yml
```diff
@@ -0,0 +1,49 @@
+name: IrSO Functional Tests
+
+on:
+  pull_request:
```
tuminoid, line 4: In [BMO we have](https://github.com/metal3-io/baremetal-operator/blob/24cecbedb46e09bccd55629977098ba38797f4fb/.github/workflows/pipeline.yml#L5) also
` types: [opened, edited, synchronize, reopened, ready_for_review]`

By [default](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request), `pull_request` triggers only on `opened, synchronize, or reopened`.

I think we should be consistent in our repos for better develover experience, so either this should have the expanded type triggers or we should simplify BMO/CAPM3/IPAM.  

I see IRSO improvement PR also merged just this simple trigger, so same should be applied there.

This can also be done in follow-up, no need to block this.

tuminoid (commented):
Noting the trigger inconsistency which can be fixed in follow-up, if wanted. I'll add a hold, if you want to address it now, and if not, then just unhold.
/hold

Otherwise this is LGTM and is a good cross-check addition.
/lgtm

.github/workflows/irso-functional.yml
tuminoid, line 4: Also, this is now triggering on all branches, including release. The ci-e2e.sh scripts are not present in branches. So we need to either backport this, or limit it to main branch only.

/hold

dtantsur: "edited" seems strange: it will trigger on e.g. editing the PR title?

"ready_for_review" probably makes sense

dtantsur: Pardon my ignorance, workflows are branchless, i.e. always taken from main? I definitely remember us backporting workflows.

tuminoid: It depends which trigger is used, but in case of `pull_request` trigger, you're actually correct, it is triggered from the target branch and not the main like some others.

tuminoid: I cross-checked also CAPM3, there we have yet another set of types, so indeed we should make it consistent there too. 

`edited` is necessary for some of the checks like PR verifier and DCO which inteact with the PR metadata, but having `edited` with `pull_request` is just re-running stuff for no reason. It was probably added as a workaround in BMO to the Prow issue of lacking approval after repushing to branch for external contributors, and we did not have the PR auto-reapprover workflow yet at the time.

 I suppose we agree the types list would `opened, synchronize, reopened, ready_for_review'  ?

dtantsur: Looks reasonable to me

tuminoid: I'll push some PRs to make it consistent across the repos, if you update this one.

tuminoid: Made following to make triggers match in all of the other repos.
- https://github.com/metal3-io/cluster-api-provider-metal3/pull/2318
- https://github.com/metal3-io/ip-address-manager/pull/885
- https://github.com/metal3-io/baremetal-operator/pull/2242
- https://github.com/metal3-io/ironic-standalone-operator/pull/127

dtantsur:
/unhold

PTAL @tuminoid

dtantsur:

tuminoid (commented):
/lgtm

PR #623, opened 2025-02-11
✨ Add scripts to run database upgrades and online migrations

dtantsur:
The new entrypoints rundatabase-upgrade and runonline-data-migrations are
designed to be run as a Job every time Ironic is upgraded. See also:
https://docs.openstack.org/ironic/latest/admin/upgrade-guide.html

### DIFF
--- a/scripts/rundatabase-upgrade
+++ b/scripts/rundatabase-upgrade
@@ -1,0 +1,10 @@
+#!/usr/bin/bash
+
+set -euxo pipefail
+
+# shellcheck disable=SC1091
+. /bin/configure-ironic.sh
+
+# NOTE(dtantsur): no retries here: this script is supposed to be run as a Job
+# that is retried on failure.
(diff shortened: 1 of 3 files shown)


tuminoid (commented):
I believe this [IRSO PR](https://github.com/metal3-io/ironic-standalone-operator/pull/168) will utilize these scripts then, do we add any support for these without IRSO?

dtantsur:
> I believe this https://github.com/metal3-io/ironic-standalone-operator/pull/168 will utilize these scripts then, do we add any support for these without IRSO?

I think it would be very tricky: how do you detect upgrades with Kustomize? Suggestions welcome, but so far I'm planning on IrSO support only.

dtantsur:
> It is fine by me, IRSO is the future way anyways.

Yeah, and these jobs only matter for the HA+MariaDB scenario, which nobody is planning to add to the Kustomize approach anyway.

tuminoid (commented):
/lgtm

PR #634, opened 2025-03-06
:sparkles: Add support for custom configuration directories

Rozzii:
This commit:
 - Introduces an environment variable that will cause all configuration files to be relocated under the specified location.
 - Introduces a .gitignore file in order to exclude temporary files from version control.
 - Converts the probes to be parameterized via environment variables thus removes the need for jinja templates.
 - A few configuration template files that were copied to /etc during image building have been moved to /tmp to keep the root of /etc more organized.

This commit is needed in order to turn the ironic container compatible with K8s readOnlyRootFilesystem pod security context.


scripts/configure-ironic.sh
```diff
@@ -72,9 +71,9 @@ if [[ -f "${IMAGE_CACHE_PREFIX}.kernel" ]] && [[ -f "${IMAGE_CACHE_PREFIX}.initr
     export IRONIC_DEFAULT_RAMDISK="${IMAGE_CACHE_PREFIX}.initramfs"
 fi
 
-if [[ -f /etc/ironic/ironic.conf ]]; then
+if [[ -f "${IRONIC_CONF_DIR}/ironic.conf" ]]; then
     # Make a copy of the original supposed empty configuration file
-    cp /etc/ironic/ironic.conf /etc/ironic/ironic.conf_orig
+    cp "${IRONIC_CONF_DIR}/ironic.conf" "${IRONIC_CONF_DIR}/ironic.conf_orig"
```
tuminoid, line 76:
```suggestion
    cp "${IRONIC_CONF_DIR}/ironic.conf" "${IRONIC_CONF_DIR}/ironic.conf.orig"
```
or `.bak` or something, not `_` in suffix please.

scripts/ironic-common.sh
```diff
@@ -7,6 +7,27 @@ PROVISIONING_INTERFACE="${PROVISIONING_INTERFACE:-}"
 PROVISIONING_IP="${PROVISIONING_IP:-}"
 PROVISIONING_MACS="${PROVISIONING_MACS:-}"
 IPXE_CUSTOM_FIRMWARE_DIR="${IPXE_CUSTOM_FIRMWARE_DIR:-/shared/custom_ipxe_firmware}"
+CUSTOM_CONFIG_DIR="${CUSTOM_CONFIG_DIR:-/conf}"
+CUSTOM_DATA_DIR="${CUSTOM_DATA_DIR:-/data}"
+export IRONIC_CONF_DIR="${CUSTOM_CONFIG_DIR}/ironic"
```
tuminoid, line 12: Group these per component. Alphabetical sort should do the trick.

scripts/rundnsmasq
```diff
@@ -32,12 +32,12 @@ fi
...
-python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ))' </etc/dnsmasq.conf.j2 >/tmp/dnsmasq.conf
+python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ))' <"/tmp/dnsmasq.conf.j2" >"${DNSMASQ_TEMP_DIR}/dnsmasq_temp.conf"
 
 for iface in $(echo "$DNSMASQ_EXCEPT_INTERFACE" | tr ',' ' '); do
-    sed -i -e "/^interface=.*/ a\except-interface=${iface}" /tmp/dnsmasq.conf
+    sed -i -e "/^interface=.*/ a\except-interface=${iface}" "${DNSMASQ_TEMP_DIR}/dnsmasq_temp.conf"
 done
-cat /tmp/dnsmasq.conf > /etc/dnsmasq.conf
-rm /tmp/dnsmasq.conf
+cat "${DNSMASQ_TEMP_DIR}/dnsmasq_temp.conf" > "${DNSMASQ_CONF_DIR}/dnsmasq.conf"
```
tuminoid, line 40: This temp hack was needed in `/etc` due file ownerships for non-root users. I don't think we have same issue here in CONF_DIR?

tuminoid (commented):
Some comments and questions.

Dockerfile
```diff
@@ -44,21 +44,26 @@ RUN prepare-image.sh && \
 
 COPY scripts/ /bin/
 
+RUN mv /bin/ironic-probe.sh /bin/ironic-readiness && \
+    cp /bin/ironic-readiness /bin/ironic-liveness && \
+    mkdir -p /conf /data/db /conf/ipxe
```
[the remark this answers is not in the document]
Rozzii: it is intentional

ironic-config/httpd.conf.j2
```diff
@@ -1,10 +1,10 @@
-ServerRoot "/etc/httpd"
+ServerRoot {{ env.HTTPD_DIR }}
```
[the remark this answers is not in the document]
Rozzii: module configurations and this conf file, It is hard coded in modules.conf because I determined that it more secure to leave the apache libraries on the root file system so that when RO FS is enabled then the libraries are not moved to the external mount.

Rozzii:
@tuminoid  thanks for the review, I will address them soon.

ironic-config/httpd.conf.j2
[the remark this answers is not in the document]
tuminoid: Ah I see. I agree modules are fine where they are, no point moving them.

Dockerfile
```diff
@@ -44,21 +44,26 @@ RUN prepare-image.sh && \
...
-COPY ironic-config/dnsmasq.conf.j2 /etc/
+COPY ironic-config/dnsmasq.conf.j2 /tmp/
 
 # Custom httpd config, removes all but the bare minimum needed modules
 COPY ironic-config/httpd.conf.j2 /etc/httpd/conf/
 COPY ironic-config/httpd-modules.conf /etc/httpd/conf.modules.d/
-COPY ironic-config/apache2-vmedia.conf.j2 /etc/httpd-vmedia.conf.j2
-COPY ironic-config/apache2-ipxe.conf.j2 /etc/httpd-ipxe.conf.j2
+COPY ironic-config/apache2-vmedia.conf.j2 /tmp/httpd-vmedia.conf.j2
+COPY ironic-config/apache2-ipxe.conf.j2 /tmp/httpd-ipxe.conf.j2
```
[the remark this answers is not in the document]
tuminoid: They could also just be in /etc/httpd since they're apache templates, so they don't pollute root of /etc. No biggie, we can debate this in a follow-up.

Dockerfile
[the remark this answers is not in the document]
tuminoid: ...and why is it intentional, what does it achieve?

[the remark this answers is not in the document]
Rozzii: I have miss read your comment I didn't notice the proposed diff, sure I can remove the /conf

Dockerfile
```diff
@@ -44,21 +44,26 @@ RUN prepare-image.sh && \
 
 COPY scripts/ /bin/
 
+RUN mv /bin/ironic-probe.sh /bin/ironic-readiness && \
```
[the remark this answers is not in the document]
elfosardo: having proper naming will avoid also adding another RUN step and we can just reuse the one on L42, or even include any change in the prepare-image.sh script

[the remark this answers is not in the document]
Rozzii: I will do it in prepare-image.sh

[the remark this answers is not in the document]
Rozzii: Done

scripts/ironic-common.sh
Rozzii: I have done it everywhere where the dependency chain allowed it and also separated the file paths from the directory paths.

scripts/configure-ironic.sh
Rozzii: Done, I have went with .orig

scripts/rundnsmasq
Rozzii: I would like to leave this as is because the user can still set the dnsmasq directory to /etc or any other location on the root FS even though by default we are not doing it.

Dockerfile
```diff
@@ -35,36 +35,33 @@ ARG IRONIC_SOURCE
...
+RUN prepare-image.sh && rm -f /bin/prepare-image.sh
+
 # IRONIC #
 COPY --from=ironic-builder /tmp/ipxe/src/bin/undionly.kpxe /tmp/ipxe/src/bin-x86_64-efi/snponly.efi /tftpboot/
 COPY --from=ironic-builder /tmp/esp.img /tmp/uefi_esp.img
 
 COPY ironic-config/ironic.conf.j2 /etc/ironic/
-COPY ironic-config/inspector.ipxe.j2 ironic-config/httpd-ironic-api.conf.j2 ironic-config/ipxe_config.template /tmp/
+COPY ironic-config/inspector.ipxe.j2 ironic-config/httpd-ironic-api.conf.j2  /tmp/
+COPY ironic-config/ipxe_config.template /tmp/ipxe_config.template
```
tuminoid, line 50: Why split into separate COPY ? Same source and target and file is not renamed.

Dockerfile
```diff
@@ -35,36 +35,33 @@ ARG IRONIC_SOURCE
...
 
 # DNSMASQ
-COPY ironic-config/dnsmasq.conf.j2 /etc/
+COPY ironic-config/dnsmasq.conf.j2 /tmp/
```
tuminoid, line 53: These all configs that go to /tmp should be in same COPY.

scripts/ironic-common.sh
```diff
@@ -44,9 +66,12 @@ export LISTEN_ALL_INTERFACES="${LISTEN_ALL_INTERFACES:-true}"
 # Wait for the interface or IP to be up, sets $IRONIC_IP
 wait_for_interface_or_ip()
 {
-    # If $PROVISIONING_IP is specified, then we wait for that to become available on an interface, otherwise we look at $PROVISIONING_INTERFACE for an IP
+    # If $PROVISIONING_IP is specified, then we wait for that to become
+    # available on an interface, otherwise we look at $PROVISIONING_INTERFACE
+    # for an IP
     if [[ -n "$PROVISIONING_IP" ]]; then
```
tuminoid, line 72:
```suggestion
    if [[ -n "${PROVISIONING_IP}" ]]; then
```
While touching basically all lines around it, might fix it while at it.

scripts/ironic-common.sh
```diff
@@ -83,16 +108,20 @@ run_ironic_dbsync()
     if [[ "${IRONIC_USE_MARIADB}" == "true" ]]; then
         # It's possible for the dbsync to fail if mariadb is not up yet, so
         # retry until success
-        until ironic-dbsync --config-file /etc/ironic/ironic.conf upgrade; do
+        until ironic-dbsync --config-file "${IRONIC_CONF_DIR}/ironic.conf" \
```
tuminoid, line 111: This formatting looks awful, can we skip this.

scripts/ironic-common.sh
```diff
@@ -83,16 +108,20 @@ run_ironic_dbsync()
...
-        # the schema in one go if not already created, instead of going through an upgrade
-        DB_VERSION="$(ironic-dbsync --config-file /etc/ironic/ironic.conf version)"
+        # SQLite does not support some statements. Fortunately, we can just
+        # create the schema in one go if not already created, instead of going
+        # through an upgrade
+        cp "/var/lib/ironic/ironic.sqlite" "${IRONIC_DB_DIR}/ironic.sqlite"
+        DB_VERSION="$(ironic-dbsync --config-file "${IRONIC_CONF_DIR}/ironic.conf" version)"
         if [[ "${DB_VERSION}" == "None" ]]; then
-            ironic-dbsync --config-file /etc/ironic/ironic.conf create_schema
+            ironic-dbsync --config-file "${IRONIC_CONF_DIR}/ironic.conf" \
```
tuminoid, line 123: Same

tuminoid (commented):
Some nits.

Dockerfile
Rozzii: done

Dockerfile
Rozzii: done

scripts/ironic-common.sh
Rozzii: done

scripts/ironic-common.sh
Rozzii: done

scripts/ironic-common.sh
Rozzii: done

scripts/ironic-common.sh
```diff
@@ -7,6 +7,28 @@ PROVISIONING_INTERFACE="${PROVISIONING_INTERFACE:-}"
...
+export HTTPD_CONF_DIR="${HTTPD_DIR}/conf"
+export HTTPD_CONF_DIR_D="${HTTPD_DIR}/conf.d"
+export IPXE_CONF_DIR="${CUSTOM_CONFIG_DIR}/ipxe"
+export IRONIC_CONF_DIR="${CUSTOM_CONFIG_DIR}/ironic"
+export IRONIC_DB_DIR="${CUSTOM_DATA_DIR}/db"
+export IRONIC_GEN_CERT_DIR="${CUSTOM_DATA_DIR}/auto_gen_certs"
+export IRONIC_TMP_DATA_DIR="${CUSTOM_DATA_DIR}/tmp"
+export PROBE_CONF_DIR="${CUSTOM_CONFIG_DIR}/probes"
+
+mkdir -p "${IRONIC_CONF_DIR}" "${PROBE_CONF_DIR}" "${HTTPD_CONF_DIR}" \
```
tuminoid, line 25: This is executed runtime, and fails:

```text
  97   │ +++ mkdir -p /conf/ironic /conf/probes /conf/httpd/conf /conf/httpd/conf.d /conf/dnsmasq /conf/dnsmasq /data/db /data/auto_gen_certs /conf/ipxe /data/dnsmasq 
       │ /data/tmp
  98   │ mkdir: cannot create directory '/conf/ironic': Permission denied
  99   │ mkdir: cannot create directory '/conf/probes': Permission denied
 100   │ mkdir: cannot create directory '/conf/httpd': Permission denied
 101   │ mkdir: cannot create directory '/conf/httpd': Permission denied
 102   │ mkdir: cannot create directory '/conf/dnsmasq': Permission denied
 103   │ mkdir: cannot create directory '/conf/dnsmasq': Permission denied
 104   │ mkdir: cannot create directory '/data/db': Permission denied
 105   │ mkdir: cannot create directory '/data/auto_gen_certs': Permission denied
 106   │ mkdir: cannot create directory '/conf/ipxe': Permission denied
 107   │ mkdir: cannot create directory '/data/dnsmasq': Permission denied
 108   │ mkdir: cannot create directory '/data/tmp': Permission denied
```

During image build you need to precreate all these or allow non-root user to create them runtime.

Rozzii: I think the issue here is that I have to make /data and /conf owned by non root so I have to put these two in the scope of the nonroot script.

Rozzii:
The irso issue is related to the location of the htpasswd-rpc file and because it is effected by the change in the location of the ironic config directory. I will fix it soon.

tuminoid (commented):
/lgtm

/cc @dtantsur

Dockerfile
```diff
@@ -35,36 +35,32 @@ ARG IRONIC_SOURCE
 ARG SUSHY_SOURCE
 
 COPY sources /sources/
-
-COPY ${UPPER_CONSTRAINTS_FILE} ironic-packages-list ${PKGS_LIST} ${EXTRA_PKGS_LIST:-$PKGS_LIST} ${PATCH_LIST:-$PKGS_LIST} /tmp/
+COPY ${UPPER_CONSTRAINTS_FILE} ironic-packages-list ${PKGS_LIST} \
+     ${EXTRA_PKGS_LIST:-$PKGS_LIST} ${PATCH_LIST:-$PKGS_LIST} \
+     ironic-config/inspector.ipxe.j2 ironic-config/httpd-ironic-api.conf.j2 \
+     ironic-config/ipxe_config.template ironic-config/dnsmasq.conf.j2 \
+    /tmp/
```
dtantsur, line 42: nit: Maybe we start using a more specific directory for our files like `/src`

ironic-config/ironic.conf.j2
```diff
@@ -100,7 +102,7 @@ disable_deep_image_inspection = True
 {% if env.IRONIC_USE_MARIADB | lower == "true" %}
 connection = {{ env.MARIADB_CONNECTION }}
 {% else %}
-connection = sqlite:////var/lib/ironic/ironic.sqlite
+connection = {{ env.LOCAL_DB_URI }}
```
dtantsur, line 105: nit: we should finally decide whether we prefix variables with IRONIC_ or not :)

scripts/ironic-probe.sh
```diff
@@ -0,0 +1,23 @@
...
+set -eu -o pipefail
+
+# shellcheck disable=SC1091
+. /bin/ironic-common.sh
+# shellcheck disable=SC1091
+. /bin/auth-common.sh
+
+PROBE_CURL_ARGS=
+if [[ "${IRONIC_REVERSE_PROXY_SETUP}" == "true" ]]; then
+    if [[ "${IRONIC_PRIVATE_PORT}" == "unix" ]]; then
```
dtantsur, line 12: I guess this assumes that probes are running with the same variables as the container itself?

scripts/ironic-common.sh
```diff
@@ -7,6 +7,28 @@ PROVISIONING_INTERFACE="${PROVISIONING_INTERFACE:-}"
...
 IPXE_CUSTOM_FIRMWARE_DIR="${IPXE_CUSTOM_FIRMWARE_DIR:-/shared/custom_ipxe_firmware}"
+CUSTOM_CONFIG_DIR="${CUSTOM_CONFIG_DIR:-/conf}"
+CUSTOM_DATA_DIR="${CUSTOM_DATA_DIR:-/data}"
+export DNSMASQ_CONF_DIR="${CUSTOM_CONFIG_DIR}/dnsmasq"
+export DNSMASQ_DATA_DIR="${CUSTOM_DATA_DIR}/dnsmasq"
+export DNSMASQ_TEMP_DIR="${CUSTOM_CONFIG_DIR}/dnsmasq"
+export HTTPD_DIR="${CUSTOM_CONFIG_DIR}/httpd"
+export HTTPD_CONF_DIR="${HTTPD_DIR}/conf"
+export HTTPD_CONF_DIR_D="${HTTPD_DIR}/conf.d"
+export IPXE_CONF_DIR="${CUSTOM_CONFIG_DIR}/ipxe"
```
dtantsur, line 18: What is this used for? Is it even used at all?

scripts/runhttpd
```diff
@@ -76,4 +82,5 @@ configure_restart_on_certificate_update "${IRONIC_TLS_SETUP}" httpd "${IRONIC_CE
 # Set up inotify to kill the container (restart) whenever cert of httpd for /shared/html/<redifsh;ilo> path change
 configure_restart_on_certificate_update "${IRONIC_VMEDIA_TLS_SETUP}" httpd "${IRONIC_VMEDIA_CERT_FILE}"
 
-exec /usr/sbin/httpd -DFOREGROUND
+exec /usr/sbin/httpd -DFOREGROUND \
```
dtantsur, line 85: nit: I don't think the line break is justified here  :)

dtantsur:
@Rozzii could you check my comments and especially update https://github.com/metal3-io/ironic-image/pull/623 in a follow-up before @elfosardo does a release? Otherwise, this change will break my work.

/approve

scripts/ironic-common.sh
Rozzii: The IPXE? Yes at least we use it nowadays and I am working on extra features for it, I have also updated dev-env to be compatible with the latest iPXE builder.

Rozzii: Nice catch, I think it is not used anymore, if I remember correctly at some point I was putting the ipxe template here but then I have just grouped it with the other templates.

scripts/ironic-probe.sh
Rozzii: Yes, this is the same logic we used before I just lifted it from the config script and placed here with small modification.

scripts/runhttpd
Rozzii: Agree

Issue #635, opened 2025-03-10
Ironic's api_workers is configured wrong when processor count is more than 4


senthilkumarkj:
The bug / regression was introduced in this commit - https://github.com/metal3-io/ironic-image/commit/02a56a99570b59005b5319f1f5fd83dfe6934c8e#diff-ab2b20df06f98d48f232d90f0b7aa464704257224862780635ec45b0ce8a26d4L23-L50

Ironic's `api_workers` default is 4 if number of processors is greater than or equal to 4. Otherwise the number of processors is used.

https://docs.openstack.org/ironic/latest/configuration/config.html#api.api_workers

In ironic-image configuration also, it was following the same logic earlier but the above mentioned commit, the number is set to > 4. It doesn't look like it's intentional.

tuminoid:
You are correct. The original code sets the maximum limit of 4, while the new version sets minimum limit of 4. This is indeed change in behavior. Good catch!

/triage accepted
/kind bug

tuminoid:
/assign

I did a mess, so I'll clean it up. PRs incoming soon.

tuminoid:
#636 created, which removes the whole logic. As the comment says, this has been expected to land on Ironic API side, and it did, in 2020(!), so it is long past overdue to get rid of the whole thing.

I've also set it to be cherry-picked to release branches for 26.0, 27.0 and 28.0.

Thanks for noticing this @senthilkumarkj !

PR #636, opened 2025-03-10
:bug: remove faulty NUMPROC logic

tuminoid:
This removes the faulty NUMPROC processor count detection logic from the configure-ironic.sh. It used to limit CPU count to 4, but the "fixed" version made it a MINIMUM of 4.

This detection has now been merged on the ironic api side, so we can remote the note + the faulty logic, and just leave the minimal env var there. 0 means ironic will run its autodetection logic and set maximum of 4 cores.

This fix has been merged on the Ironic side already in 2020, so we can backport this to all branches we have.

Fixes: #635


tuminoid:
Unrelated issue, IPAM problem.  There is work on going to get that resolved.

Rozzii (commented):
/lgtm

PR #637, opened 2025-03-11
⚠️ Deprecate passing MARIADB_PASSWORD in favour of mounting a secret

dtantsur:
Passing a password in clear text is an obviously bad idea. Add support
for mounting a secret and deprecate the clear text approach.

This change also makes it a fatal error to not pass the password.
Our mariadb-image does not support empty passwords anyway, and we should
not rely on anyone actually using "change_me" without realizing it.

### DIFF
--- a/scripts/configure-ironic.sh
+++ b/scripts/configure-ironic.sh
@@ -19,13 +19,19 @@ export IRONIC_ENABLE_VLAN_INTERFACES=${IRONIC_ENABLE_VLAN_INTERFACES:-${IRONIC_I
 
 export HTTP_PORT=${HTTP_PORT:-80}
 
-MARIADB_PASSWORD=${MARIADB_PASSWORD:-change_me}
-MARIADB_DATABASE=${MARIADB_DATABASE:-ironic}
-MARIADB_USER=${MARIADB_USER:-ironic}
-MARIADB_HOST=${MARIADB_HOST:-127.0.0.1}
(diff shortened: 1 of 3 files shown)


scripts/configure-ironic.sh
```diff
@@ -19,13 +19,19 @@ export IRONIC_ENABLE_VLAN_INTERFACES=${IRONIC_ENABLE_VLAN_INTERFACES:-${IRONIC_I
...
-MARIADB_DATABASE=${MARIADB_DATABASE:-ironic}
-MARIADB_USER=${MARIADB_USER:-ironic}
-MARIADB_HOST=${MARIADB_HOST:-127.0.0.1}
-export MARIADB_CONNECTION="mysql+pymysql://${MARIADB_USER}:${MARIADB_PASSWORD}@${MARIADB_HOST}/${MARIADB_DATABASE}?charset=utf8"
-if [[ "$MARIADB_TLS_ENABLED" == "true" ]]; then
-    export MARIADB_CONNECTION="${MARIADB_CONNECTION}&ssl=on&ssl_ca=${MARIADB_CACERT_FILE}"
+export IRONIC_USE_MARIADB=${IRONIC_USE_MARIADB:-true}
+if [[ "${IRONIC_USE_MARIADB}" == true ]]; then
+    if [[ -z "${MARIADB_PASSWORD:-}" ]]; then
+        echo "FATAL: IRONIC_USE_MARIADB requires password, mount a secret to /auth/mariadb"
```
tuminoid, line 25:
```suggestion
        echo "FATAL: IRONIC_USE_MARIADB requires password, mount a secret under /auth/mariadb"
```

tuminoid (commented):
/lgtm

This is close to what we reviewed on mariadb-image side already.

PR #638, opened 2025-03-11
✨ Allow changing MariaDB port

dtantsur:


dtantsur:
/close

IrSO already passes port as part of MARIADB_HOST

PR #643, opened 2025-03-14
🌱 Refactoring: add a helper for restarting a service on certificate update

dtantsur:

### DIFF
--- a/scripts/tls-common.sh
+++ b/scripts/tls-common.sh
@@ -95,3 +95,21 @@ if [[ -f "$MARIADB_CACERT_FILE" ]]; then
 else
     export MARIADB_TLS_ENABLED="false"
 fi
+
+configure_restart_on_certificate_update()
+{
+    local enabled="$1"
+    local service="$2"
+    local cert_file="$3"
+    local signal="TERM"
+
+    if [[ "${enabled}" == "true" ]] && [[ "${RESTART_CONTAINER_CERTIFICATE_UPDATED}" == "true" ]]; then
+        if [[ "${service}" == httpd ]]; then
+            signal="WINCH"
+        fi
+        python3 -m pyinotify --raw-format -e IN_DELETE_SELF -v "${cert_file}" |
+            while read -r; do
+                pkill "-${signal}" "${service}"
+            done &
+    fi
+}
--- a/scripts/runhttpd
+++ b/scripts/runhttpd
@@ -71,19 +71,9 @@ if [[ "$IPXE_TLS_SETUP" == "true" ]]; then
 fi
 
 # Set up inotify to kill the container (restart) whenever cert files for ironic api change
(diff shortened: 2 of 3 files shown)


scripts/tls-common.sh
```diff
@@ -95,3 +95,16 @@ if [[ -f "$MARIADB_CACERT_FILE" ]]; then
...
     export MARIADB_TLS_ENABLED="false"
 fi
+
+configure_restart_on_certificate_update() {
+    local enabled="$1"
+    local service="$2"
+    local cert_file="$3"
+
+    if [[ "${enabled}" == "true" ]] && [[ "${RESTART_CONTAINER_CERTIFICATE_UPDATED}" == "true" ]]; then
+        python3 -m pyinotify -e IN_DELETE_SELF -v "${cert_file}" |
```
dtantsur, line 105: Side note: this invocation does not make any sense to me. I don't understand why we're waiting for a file deletion, it does not look like this will be triggered the 2nd time, and it does look like inotify outputs a lot of junk that will trigger the event.

scripts/tls-common.sh
```diff
@@ -95,3 +95,16 @@ if [[ -f "$MARIADB_CACERT_FILE" ]]; then
...
+            while read -r; do
+                kill $(pgrep "${service}")
```
tuminoid, line 107:
```suggestion
                pkill "${service}"
```

scripts/tls-common.sh
elfosardo: I had in TODO list to change this as it does not work as it is, do we want to do it here or in a followup?

tuminoid (commented):
/lgtm

This is good improvement, great reduction in duplicated code.

I'll add
/hold
for this [comment](https://github.com/metal3-io/ironic-image/pull/643#discussion_r1995601857) and merge control, feel free to remove hold whichever way you decide to address the comment.

scripts/tls-common.sh
elfosardo: let's merge this then, I'll add a follow up right after

scripts/tls-common.sh
```diff
@@ -95,3 +95,21 @@ if [[ -f "$MARIADB_CACERT_FILE" ]]; then
...
+    local enabled="$1"
+    local service="$2"
+    local cert_file="$3"
+    local signal="TERM"
+
+    if [[ "${enabled}" == "true" ]] && [[ "${RESTART_CONTAINER_CERTIFICATE_UPDATED}" == "true" ]]; then
+        if [[ "${service}" == httpd ]]; then
+            signal="WINCH"
+        fi
+        python3 -m pyinotify --raw-format -e IN_DELETE_SELF -v "${cert_file}" |
```
elfosardo, line 110: the problem is the debugging output, removing -v should fix the execution
anyway, let's merge this and I will also change the while loop

dtantsur: Yeah, thank you!

PR #644, opened 2025-03-17
✨ Add configurable max_pool_size and max_overflow

shibaPuppy:

Added support for configuring `max_pool_size` and `max_overflow` in the database section of the Jinja template. These values are now derived from environment variables `IRONIC_MAX_POOL_SIZE` and `IRONIC_MAX_OVERFLOW`, with default values set to 5 and 50, respectively.

This change allows for better control over database connection pooling, improving scalability and performance.

Considering our environment with approximately 3,028 Ironic nodes, adjusting these values may be necessary to handle concurrent operations efficiently.

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,9 @@ functionality:
 - `IRONIC_ENABLE_VLAN_INTERFACES` - Which VLAN interfaces to enable on the
   agent start-up. Can be a list of interfaces or a special value `all`.
   Defaults to `all`.
+- `OS_DATABASE__MAX_POOL_SIZE` - Maximum number of SQL connections to
+  keep open in a pool.
+- `OS_DATABASE__MAX_OVERFLOW` - If set, use this value for max_overflow with SQLAlchemy.
 
 MariaDB configuration:
 
--- a/ironic-config/ironic.conf.j2
+++ b/ironic-config/ironic.conf.j2
@@ -106,6 +106,8 @@ connection = sqlite:////var/lib/ironic/ironic.sqlite
 # IO by not doing syncs all the time.
 sqlite_synchronous = False
 {% endif %}
+max_pool_size = {{ env.OS_DATABASE__MAX_POOL_SIZE | default(5) }}
+max_overflow = {{ env.OS_DATABASE__MAX_OVERFLOW | default(50) }}
 
 [deploy]
 default_boot_option = local


shibaPuppy:
@dtantsur @Rozzii

README.md
```diff
@@ -24,21 +24,21 @@ specific entry points.
 The following entry points are provided:
 
 - `runironic` - Starts the ironic-conductor and ironic-api processes to manage
-   the provisioning of baremetal nodes.  Details on Ironic can be found at
-   <https://docs.openstack.org/ironic/latest/>.  This is the default entry point
-   used by the Dockerfile.
+  the provisioning of baremetal nodes. Details on Ironic can be found at
```
tuminoid, line 27: Please do not touch the whitespaces.

tuminoid (commented):
Thanks for the contribution.

Please leave the leading whitespaces alone. Yes, having 3 spaces is weird but it is (was) legacy tooling issue.

/cc @dtantsur

Rozzii:
I can't find these options in https://opendev.org/openstack/ironic/src/branch/master/ironic/conf/database.py or anywhere else in ironic as an externally configurable option.

Are you sure these can be configured via the config file @shibaPuppy  ?
/hold
I am holding it until we clarify this.

iurygregory:
This is interesting, I will double check upstream with the ironic community 😅

iurygregory:
@Rozzii configs are valid https://github.com/openstack/oslo.db/blob/master/oslo_db/options.py

Rozzii:
> @Rozzii configs are valid https://github.com/openstack/oslo.db/blob/master/oslo_db/options.py

Ooo I see yet again the beautiful "multy tenant " config files of oslo.... so it is not Ironic but the db library that picks these...
/hold cancel

shibaPuppy:
> 스크립트에서 직접 사용되지 않는 새로운 변수를 도입하지 마세요. `OS_GROUP__VARIABLE`대신 feature를 사용할 수 있습니다. README에 언급되어 있습니다.

I have changed it to the OS_<section>__<name>=<value> format.

Rozzii:
> > 스크립트에서 직접 사용되지 않는 새로운 변수를 도입하지 마세요. `OS_GROUP__VARIABLE`대신 feature를 사용할 수 있습니다. README에 언급되어 있습니다.
> 
> I have changed it to the `OS_<section>__<name>=<value>` format.

I think what @dtantsur meant is that you don't have to add these vars to the config file, if you just supply this OS_<section>_<name> variable during deployment then it will be picked up by all the components that are using oslo.config no need to extend the config file.

shibaPuppy:
```
env |grep ^OS
OS_DATABASE__MAX_POOL_SIZE=999

python3 -c 'import os; import sys; import jinja2; sys.stdout.write(jinja2.Template(sys.stdin.read()).render(env=os.environ))' < "/etc/ironic/ironic.conf.j2" > "/etc/ironic/ironic.conf"

cat /etc/ironic/ironic.conf |grep -i pool
```

Even after rendering the Jinja2 template, the expected values do not appear in the final conf file.
Is there something I might have missed?

elfosardo:
@shibaPuppy  can we close this now?

PR #646, opened 2025-03-19
:sparkles: Add build ironic-image with CentOS Stream 10

elfosardo:


.github/workflows/build-images-action.yml
```diff
@@ -23,6 +23,19 @@ jobs:
       QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
       QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
       SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+  build_ironic_cs10:
+    name: Build Ironic container image based on CentOS Stream 10
+    env:
+      IMAGE_BUILD_ARGS: ['BASE_IMAGE=quay.io/centos/centos:stream10']
```
tuminoid, line 29:
```suggestion
      IMAGE_BUILD_ARGS:
      - 'BASE_IMAGE=quay.io/centos/centos:stream10'
```
This list format is more readable, and especially if we add anything else there.

.github/workflows/build-images-action.yml
```diff
@@ -23,6 +23,19 @@ jobs:
...
+    if: github.repository == 'metal3-io/ironic-image'
+    uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main
+    with:
+      image-name: 'ironic_cs10'
```
tuminoid, line 33: This needs to be created in Quay before this merges, otherwise the push will fail.

tuminoid (commented):
/hold
Repository need to be created in Quay before merging this.

.github/workflows/build-images-action.yml
elfosardo: created https://quay.io/repository/metal3-io/ironic_cs10

.github/workflows/build-images-action.yml
```diff
@@ -23,6 +23,20 @@ jobs:
       QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
       QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
       SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+  build_ironic_cs10:
+    name: Build Ironic container image based on CentOS Stream 10
+    env:
```
tuminoid, line 28: Tested this with `gh act` locally.

`Line: 27 Column 5: Failed to match workflow-job: Line: 28 Column 5: Unknown Property env`

elfosardo: I see also other errors running act and from the github workflow VS extension, but I'm not sure they are correct, maybe I'm overlooking something
I based my change on https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-environment-variables-for-a-single-workflow

elfosardo: ooooook so apparently the env does not work with uses in that way, it seems you need to add env inside  with
it does pass both act and VS code check now

elfosardo:

.github/workflows/build-images-action.yml
```diff
@@ -23,6 +23,18 @@ jobs:
...
       QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
       SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+  build_ironic_cs10:
+    name: Build Ironic container image based on CentOS Stream 10
+    if: github.repository == 'metal3-io/ironic-image'
+    uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main
+    with:
+      image-name: 'ironic_cs10'
+      pushImage: true
+      env.IMAGE_BUILD_ARGS: 'BASE_IMAGE=quay.io/centos/centos:stream10'
```
tuminoid, line 33:
```suggestion
      env:
        IMAGE_BUILD_ARGS: 'BASE_IMAGE=quay.io/centos/centos:stream10'
```

tuminoid (commented):
Let's write the formatting more consistent way. Otherwise LGTM.

.github/workflows/build-images-action.yml
elfosardo: @tuminoid this formatting does not work, I tried it already, it results in an error `Line: 27 Column 5: Failed to match workflow-job: Line: 34 Column 9: Failed to match string: Line: 34 Column 9: Expected a scalar got mapping`

tuminoid: OK... maybe that is a `with:` specialty. Lets merge as is then.

tuminoid (commented):
/lgtm

PR #658, opened 2025-04-01
:seedling: group all dependabot github action updates into one

tuminoid:
Group all dependabot github action updates into one PR.

I forgot to do this change for ironic-image, when applied it to all repos we have.


tuminoid:

Rozzii (commented):
/lgtm

PR #661, opened 2025-04-02
🐛 Stop doing mkdir on /certs subdirectories

dtantsur:
With a read-only root filesystem and no TLS, this action fails.
Add a conditional mkdir in the only place where we actually try
to write to one location.


tuminoid (commented):
/lgtm

dtantsur:
> Do we need to cherry-pick further ?

No, the read-only feature was added in 29.0

PR #664, opened 2025-04-03
🐛 Do not use /tmp as a source directory for the image

dtantsur:
Unfortunately, a lot of code in Ironic ignores the tempdir
configuration. Even worse, the default value for this option relies on
having a writable /tmp, which makes the current approach not functional
without mounting an emptyDir to /tmp. But if an emptyDir is mounted to
/tmp, we lose access to all the template files stored there.

This change uses a new directory /templates as a source for files that
needs copying/templating in runtime.

Also fix usage of /var/tmp in httpd.conf.

Document the requirements.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,9 +37,10 @@ ARG SUSHY_SOURCE
 COPY sources /sources/
 COPY ${UPPER_CONSTRAINTS_FILE} ironic-packages-list ${PKGS_LIST} \
      ${EXTRA_PKGS_LIST:-$PKGS_LIST} ${PATCH_LIST:-$PKGS_LIST} \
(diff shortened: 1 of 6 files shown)


Rozzii (changes requested):
As we have discussed prior to this PR, I am okay with this modification, but I think you have to add the "mkdir" of the /template into the "prepare-image.sh"

tuminoid:
> As we have discussed prior to this PR, I am okay with this modification, but I think you have to add the "mkdir" of the /template into the "prepare-image.sh"

/template is already created by the Dockerfile `COPY`, prepare-image.sh is run long after that.

Rozzii (commented):
/lgtm

tuminoid (commented):
/lgtm

PR #666, opened 2025-04-04
:seedling: add labels to dockerfile

tuminoid:
Add labels to Dockerfile. `images.version` label will be set by automation during build.

Fix some indentation issues and convert PKGS_LIST to ARG.

Supersedes abandoned PR https://github.com/metal3-io/ironic-image/pull/591

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,19 +13,29 @@ RUN dnf install -y gcc git make xz-devel
 WORKDIR /tmp
 
 RUN git clone https://github.com/ipxe/ipxe.git && \
-      cd ipxe && \
-      git reset --hard $IPXE_COMMIT_HASH && \
-      cd src && \
-      ARCH=$(uname -m | sed 's/aarch/arm/') && \
-      # NOTE(elfosardo): warning should not be treated as errors by default
-      NO_WERROR=1 make bin/undionly.kpxe "bin-$ARCH-efi/snponly.efi"
+     cd ipxe && \
+     git reset --hard $IPXE_COMMIT_HASH && \
+     cd src && \
+     ARCH=$(uname -m | sed 's/aarch/arm/') && \
+     # NOTE(elfosardo): warning should not be treated as errors by default
+     NO_WERROR=1 make bin/undionly.kpxe "bin-$ARCH-efi/snponly.efi"
 
 COPY prepare-efi.sh /bin/
 RUN prepare-efi.sh centos
 
+# build actual image
 FROM $BASE_IMAGE
 
-ENV PKGS_LIST=main-packages-list.txt
+# image.version will be set by automation during build
+LABEL org.opencontainers.image.authors="metal3-dev@googlegroups.com"
+LABEL org.opencontainers.image.description="Container image to run OpenStack Ironic as part of Metal³"
+LABEL org.opencontainers.image.documentation="https://book.metal3.io/ironic/introduction"
+LABEL org.opencontainers.image.licenses="Apache License 2.0"
+LABEL org.opencontainers.image.title="Metal3 Ironic Container"
+LABEL org.opencontainers.image.url="https://github.com/metal3-io/ironic-image"
+LABEL org.opencontainers.image.vendor="Metal3-io"
+
+ARG PKGS_LIST=main-packages-list.txt
 ARG EXTRA_PKGS_LIST
 ARG PATCH_LIST
 
@@ -60,8 +70,8 @@ COPY ironic-config/apache2-ipxe.conf.j2 /templates/httpd-ipxe.conf.j2
 
 # DATABASE
 RUN mkdir -p /var/lib/ironic && \
-  sqlite3 /var/lib/ironic/ironic.sqlite "pragma journal_mode=wal" && \
-  dnf remove -y sqlite
+     sqlite3 /var/lib/ironic/ironic.sqlite "pragma journal_mode=wal" && \
+     dnf remove -y sqlite
 
 # configure non-root user and set relevant permissions
 RUN configure-nonroot.sh && rm -f /bin/configure-nonroot.sh


Dockerfile
```diff
@@ -13,19 +13,28 @@ RUN dnf install -y gcc git make xz-devel
...
 
 COPY prepare-efi.sh /bin/
 RUN prepare-efi.sh centos
 
+# build actual image
 FROM $BASE_IMAGE
 
-ENV PKGS_LIST=main-packages-list.txt
+LABEL org.opencontainers.image.description="Container image to run OpenStack Ironic as part of Metal³"
+LABEL org.opencontainers.image.documentation="https://book.metal3.io/ironic/introduction"
```
tuminoid, line 30: Any better documentation urls? I'm OK with this or https://book.metal3.io/ironic/ironic-container-images as well.

Dockerfile
```diff
@@ -13,19 +13,28 @@ RUN dnf install -y gcc git make xz-devel
...
+LABEL org.opencontainers.image.licenses="Apache License 2.0"
+LABEL org.opencontainers.image.title="Metal3 Ironic Container"
+LABEL org.opencontainers.image.url="https://github.com/metal3-io/ironic-image"
+LABEL org.opencontainers.image.vendor="Metal3-io"
+LABEL org.opencontainers.image.version="latest"
+
```
tuminoid, line 36: Should we also add authors label? Using what email? 

```text
LABEL org.opencontainers.image.authors="metal3-dev@googlegroups.com"
```
?

Dockerfile
```diff
@@ -13,19 +13,28 @@ RUN dnf install -y gcc git make xz-devel
...
 FROM $BASE_IMAGE
 
-ENV PKGS_LIST=main-packages-list.txt
+LABEL org.opencontainers.image.description="Container image to run OpenStack Ironic as part of Metal³"
+LABEL org.opencontainers.image.documentation="https://book.metal3.io/ironic/introduction"
+LABEL org.opencontainers.image.licenses="Apache License 2.0"
+LABEL org.opencontainers.image.title="Metal3 Ironic Container"
+LABEL org.opencontainers.image.url="https://github.com/metal3-io/ironic-image"
+LABEL org.opencontainers.image.vendor="Metal3-io"
+LABEL org.opencontainers.image.version="latest"
```
tuminoid, line 35: When taking in release automation, we want to update this as part of the build, instead of committing a change to repo. Need to check if the image builder action supports this.

If it doesn't, I rather leave this off, than update this always in a PR before cutting a release.

Dockerfile
```diff
@@ -13,19 +13,31 @@ RUN dnf install -y gcc git make xz-devel
...
+     NO_WERROR=1 make bin/undionly.kpxe "bin-$ARCH-efi/snponly.efi"
 
 COPY prepare-efi.sh /bin/
 RUN prepare-efi.sh centos
 
+# build actual image
 FROM $BASE_IMAGE
 
-ENV PKGS_LIST=main-packages-list.txt
+LABEL io.k8s.description="Container image to run OpenStack Ironic as part of Metal³"
```
tuminoid, line 29: In BMO we also do `io.k8s.*` labels, do we want to add them too, or take them away from BMO and replace them with org.opencontainers ones?

Dockerfile
elfosardo: https://book.metal3.io/ironic/introduction is good for me

Dockerfile
elfosardo: I would leave org.opencontainers only, unles we have a strong reason to keep io.k8s too

Dockerfile
```diff
@@ -13,19 +13,31 @@ RUN dnf install -y gcc git make xz-devel
...
+LABEL org.opencontainers.image.authors="metal3-dev@googlegroups.com"
+LABEL org.opencontainers.image.description="Container image to run OpenStack Ironic as part of Metal³"
+LABEL org.opencontainers.image.documentation="https://book.metal3.io/ironic/introduction"
+LABEL org.opencontainers.image.licenses="Apache License 2.0"
+LABEL org.opencontainers.image.title="Metal3 Ironic Container"
+LABEL org.opencontainers.image.url="https://github.com/metal3-io/ironic-image"
+LABEL org.opencontainers.image.vendor="Metal3-io"
+LABEL org.opencontainers.image.version="latest"
+
+ARG PKGS_LIST=main-packages-list.txt
```
elfosardo, line 40: makes sense, thanks

Dockerfile
tuminoid: We've had the k8s one in BMO since 2020 by @dhellmann and it is actually also copied now to IRSO by @dtantsur . Googling around it seems to be mostly OpenShift thing: https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/images/creating-images#defining-image-metadata

lentzi90: +1 for skipping the io.k8s labels. If they are needed for openshift I am fine with keeping them though

Dockerfile
lentzi90: Both docker and podman support adding this with `docker build --label ...`. Hopefully we can do the same in the automated builds.
Regarding the CS9 and CS10, how do we handle that currently? Do we have different tags for them? I'm thinking that the version label could be the same as the image tag.

lentzi90 (commented):
This looks great!

Dockerfile
tuminoid: We have two different images in the image-build-action.yml workflow: `ironic`  for CS9 and `ironic_cs10` for CS10 images.

Dockerfile
dtantsur: I don't think we actually care in OpenShift too much. If we discover we need them, we can always re-add in our downstream fork.

dtantsur:
/approve

I'm okay with skipping the outdated labels

Dockerfile
tuminoid: https://github.com/marketplace/actions/docker-build-push-action#inputs supports labels.

I think we should thus leave it off the Dockerfile, and set it during build with 
1. primarily tag if build trigger was tagged (release build)
2. secondarily with the branch name; if branch is main, then latest (push triggered build)

WDYT?

lentzi90: +1, sounds good!

elfosardo: sounds good to me

tuminoid:
> > /approve
> > I'm okay with skipping the outdated labels
> 
> Thanks. I've dropped the `io.k8s.*` then.
> 
> /hold I'll check the release automation, holding to avoid premature merging. For everything else it seems we have consensus that it is good to go as proposed now.

/unhold
https://github.com/metal3-io/ironic-image/pull/666#discussion_r2030932120 we have agreement to add the `image.version` label during build time via automation. This automation will be implemented in project-infra's common image building action. We do the tag/branch/latest detection there, so we shouldn't need further changes here. I've replaced the `image.version` label with a comment that it will be added build time.

tuminoid:
Just in case.

lentzi90 (commented):
/lgtm

PR #669, opened 2025-04-11
🌱 Use bootloader through a local file link

dtantsur:
There is no need to go through the HTTP server for it: it is only used
locally by Ironic itself.


lentzi90 (commented):
/lgtm

PR #673, opened 2025-04-25
🐛 Extract files one-by-one in logwatch

dtantsur:
Currently, the file listing is sent to stderr, while files themselves
are sent to stdout. Then the two streams are merged in the resulting
CI report. However, they are not guaranteed to be 100% synchronized,
which results in file names appearing in the middle of log lines.
They are also neither human- nor machine-friendly.

Here I'm switching from extracting all files using -Ov to listing files
first and extracting them one by one.

I'm also adding extra new lines before each file to make the output more
readable.

### DIFF
--- a/scripts/runlogwatch.sh
+++ b/scripts/runlogwatch.sh
@@ -12,6 +12,10 @@ python3 -m pyinotify --raw-format -e IN_CLOSE_WRITE -v "${LOG_DIR}" |
         # <Event dir=False mask=0x8 maskname=IN_CLOSE_WRITE name=mylogs.gzip path=/shared/log/ironic/deploy pathname=/shared/log/ironic/deploy/mylogs.gzip wd=1 >
         FILENAME=$(echo "${filename}" | cut -d'=' -f2-)
         echo "************ Contents of ${LOG_DIR}/${FILENAME} ramdisk log file bundle **************"
-        tar -xOzvvf "${LOG_DIR}/${FILENAME}" | sed -e "s/^/${FILENAME}: /"
+        tar -tzf "${LOG_DIR}/${FILENAME}" | while read -r entry; do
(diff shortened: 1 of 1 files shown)


dtantsur:
/hold

Let's see the CI results before merging

dtantsur:

IPA taking too long to download

dtantsur:
/hold cancel

The CI does not seem to test ironic-images changes :( I've applied the same patch to OpenShift in https://github.com/openshift/ironic-image/pull/654, and the results look correct.

lentzi90:
I think the "e2e-integration" tests are not capable of building the ironic-image. They are CAPM3 tests basically.
The dev-env-integration tests should work better for this.
/test metal3-dev-env-integration-test-centos-main

dtantsur:
> I think the "e2e-integration" tests are not capable of building the ironic-image.

Then why are we running the CAPM3 jobs on this repository @lentzi90? Is it a misconfiguration?

lentzi90:
> Then why are we running the CAPM3 jobs on this repository @lentzi90? Is it a misconfiguration?

Combination of misunderstandings and tradition I would say.

Metal3-dev-env was used for all tests before. It had all the logic for both setting up the virtualized baremetal lab, and running tests against it. The testing functionality was quite limited though, and honestly completely focused on CAPM3. CAPM3 e2e tests were introduced to get more powerful tests for CAPM3, and to be able to reuse more of the CAPI e2e framework. These became an obvious next step for CAPM3.

I think we have switched to them also elsewhere, perhaps without thinking things through enough, or knowing/realizing the difference. Since then we have introduced BMO e2e tests which are far better at testing BMO, but we seem to be lacking a lot in other areas (ironic-image, utility images, IPAM, etc).

PR #674, opened 2025-05-06
:bug: Use local upper-constraints

elfosardo:
This is a temporary workaround to be able to build the container with CS9 until we're ready to switch to CS10 by default, since Openstack upstream removed Python 3.9 support


iurygregory:
waiting for the integration tests before adding lgtm

PR #677, opened 2025-05-08
🌱 Add a helper script to parse ramdisk logs

dtantsur:
The new script can be invoked like this:

```
 ./tools/parse-ramdisk-logs.py ramdisk-logs.txt /path/to/destination
```

and will extract everything in the following directory structure:

```
├── namespace~hostname
│   ├── cleaning-2025-05-06-10-38-12
│   │   ├── dmesg
│   │   ├── ip_addr
│   │   ├── journal
│   │   ├── lsblk
│   │   ├── lsblk-full
│   │   ├── lshw
│   │   ├── mdstat
│   │   ├── mount
│   │   ├── multipath
│   │   ├── parted
│   │   ├── udev
│   │   │   ├── loop0
│   │   │   ├── loop1
│   │   │   ├── sda
│   │   │   └── sr0
│   │   └── var
│   │       └── log
│   │           └── ironic-python-agent.log
│   ├── deploy-2025-05-06-10-39-31
(etc)
```

The ramdisk logs **must** be created by ironic-image with this change: https://github.com/metal3-io/ironic-image/pull/673

### DIFF
--- a/tools/parse-ramdisk-logs.py
+++ b/tools/parse-ramdisk-logs.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import re
+import sys
+
+# extraction datetime is always irrelevant
+EXTRACTION_TS = re.compile(r"^[0-9TZ\.:\-]* ?")
+CONTENTS = re.compile(r"\*{12} Contents of (?P<path>/.*)\.tar\.gz ")
+ENTRY = re.compile(r"\*{4} Entry: (?P<entry>.*) \*{4}")
+
+
+def log(msg):
+    print(msg, file=sys.stderr)
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument("source")
+parser.add_argument("destination")
+args = parser.parse_args()
+
+# Visitor is an object that is currently responsible for handling new lines.
+# These objects are hierarchical: top level is host (Ironic node), second level
+# is specific source (file or command). Initially, visitor is None: such lines
+# are skipped.
+visitor = None
+
+# Skipped is the counter for initially skipped lines.Unfortunately, this may
+# include not just the inotify chatter but also useful lines if the logs have
+# been rotated. There is nothing that can be done here: the file name is only
+# available in the beginning of a dump.
+skipped = 0
+
+
+class VisitorEntry:
+    def __init__(self, dest, entry):
+        assert entry is not None
+        path = os.path.join(dest, entry)
+        os.makedirs(os.path.dirname(path), exist_ok=True)
+        self.file = open(path, "wt")
+
+    def close(self):
+        self.file.close()
+
+    def __call__(self, line):
+        self.file.write(f"{line}\n")
+
+
+class VisitorSingleRun:
+    def __init__(self, filename):
+        assert filename is not None
+        self.filename = os.path.basename(filename)
+        self.delim = f"{self.filename}.tar.gz:"
+
+        log(f"Processing {self.filename}")
+        # Possible formats (path and .tar.gz suffix already stripped):
+        # UUID_NAMESPACE~NAME_inspect_TS
+        # UUID_NAMESPACE~NAME_INSTUUID_cleaning_TS
+        # UUID_NAMESPACE~NAME_INSTUUID_TS
+        uuid, name, *rest, ts = self.filename.split("_")
+        stage = rest[-1]
+        if stage not in ("cleaning", "inspect"):
+            stage = "deploy"
+        log(f".. {stage} on node {name} at {ts}")
+
+        self.dest = os.path.join(args.destination, name, f"{stage}-{ts}")
+
+        # Visitor for the current entry (file from the ramdisk)
+        self.visitor = None
+
+    def close(self):
+        if self.visitor is not None:
+            self.visitor.close()
+
+    def __call__(self, line):
+        try:
+            line = line.split(self.delim, 1)[1]
+        except IndexError:
+            pass
+        else:
+            # Strip exactly one space from the start, if any
+            if line[0:1] == " ":
+                line = line[1:]
+
+        entry = ENTRY.match(line)
+        if entry is not None:
+            if self.visitor is not None:
+                self.visitor.close()
+            self.visitor = VisitorEntry(self.dest, entry.group("entry"))
+        elif self.visitor is None:
+            log(f".. skipping line without a file: {line}")
+        else:
+            self.visitor(line)
+
+
+with open(args.source) as fp:
+    for line in fp:
+        line = line.strip()
+
+        if "pyinotify DEBUG" in line:
+            continue
+
+        line = EXTRACTION_TS.sub("", line, count=1)
+
+        contents = CONTENTS.match(line)
+        if contents is not None:
+            if visitor is None:
+                if skipped > 0:
+                    log(f"Skipped {skipped} lines because they don't have "
+                        "any context")
+                skipped = 0
+            else:
+                visitor.close()
+            visitor = VisitorSingleRun(contents.group("path"))
+            continue
+
+        if visitor is None:
+            skipped += 1
+            continue
+
+        visitor(line)


dtantsur:
/override metal3-centos-e2e-integration-test-main
/override metal3-ubuntu-e2e-integration-test-main

Helper scripts, nothing to do with the CI

tools/parse-ramdisk-logs.py
```diff
@@ -0,0 +1,122 @@
...
+# available in the beginning of a dump.
+skipped = 0
+
+
+class VisitorEntry:
+    def __init__(self, dest, entry):
+        assert entry is not None
+        path = os.path.join(dest, entry)
+        os.makedirs(os.path.dirname(path), exist_ok=True)
+        self.file = open(path, "wt")
```
tuminoid, line 41:
```suggestion
        self.file = open(path, "wt", encoding="utf-8")
```
I'm assuming here the log files are utf-8, adjust as needed.

tools/parse-ramdisk-logs.py
```diff
@@ -0,0 +1,122 @@
...
+
+# Skipped is the counter for initially skipped lines.Unfortunately, this may
+# include not just the inotify chatter but also useful lines if the logs have
+# been rotated. There is nothing that can be done here: the file name is only
+# available in the beginning of a dump.
+skipped = 0
+
+
+class VisitorEntry:
+    def __init__(self, dest, entry):
```
tuminoid, line 37: typehints would be lovely.

```suggestion
    def __init__(self, dest: str, entry: str):
```

tools/parse-ramdisk-logs.py
```diff
@@ -0,0 +1,122 @@
...
+        assert entry is not None
+        path = os.path.join(dest, entry)
+        os.makedirs(os.path.dirname(path), exist_ok=True)
+        self.file = open(path, "wt")
+
+    def close(self):
+        self.file.close()
+
+    def __call__(self, line):
```
tuminoid, line 46: typehints, and this line is redefining the line variable from L98. Scope issue should be solved by adding the `if __name__ == "__main__"`  stuff for the top-level code.

```suggestion
    def __call__(self, line: str):
```

tools/parse-ramdisk-logs.py
```diff
@@ -0,0 +1,122 @@
...
+        self.file.write(f"{line}\n")
+
+
+class VisitorSingleRun:
+    def __init__(self, filename):
```
tuminoid, line 51:
```suggestion
    def __init__(self, filename: str):
```

tools/parse-ramdisk-logs.py
```diff
@@ -0,0 +1,122 @@
...
+        assert filename is not None
+        self.filename = os.path.basename(filename)
+        self.delim = f"{self.filename}.tar.gz:"
+
+        log(f"Processing {self.filename}")
+        # Possible formats (path and .tar.gz suffix already stripped):
+        # UUID_NAMESPACE~NAME_inspect_TS
+        # UUID_NAMESPACE~NAME_INSTUUID_cleaning_TS
+        # UUID_NAMESPACE~NAME_INSTUUID_TS
+        uuid, name, *rest, ts = self.filename.split("_")
```
tuminoid, line 61:
```suggestion
        _, name, *rest, ts = self.filename.split("_")
```
uuid is not used.

tools/parse-ramdisk-logs.py
```diff
@@ -0,0 +1,122 @@
...
+            if self.visitor is not None:
+                self.visitor.close()
+            self.visitor = VisitorEntry(self.dest, entry.group("entry"))
+        elif self.visitor is None:
+            log(f".. skipping line without a file: {line}")
+        else:
+            self.visitor(line)
+
+
+with open(args.source) as fp:
```
tuminoid, line 97:
```suggestion
with open(args.source, "r", encoding="utf-8") as fp:
```

tools/parse-ramdisk-logs.py
```diff
@@ -0,0 +1,122 @@
...
+EXTRACTION_TS = re.compile(r"^[0-9TZ\.:\-]* ?")
+CONTENTS = re.compile(r"\*{12} Contents of (?P<path>/.*)\.tar\.gz ")
+ENTRY = re.compile(r"\*{4} Entry: (?P<entry>.*) \*{4}")
+
+
+def log(msg):
+    print(msg, file=sys.stderr)
+
+
+parser = argparse.ArgumentParser()
```
tuminoid, line 18: I'd put this in a function that is called from main, but if you don't wish to pass args around, its fine.

tools/parse-ramdisk-logs.py
```diff
@@ -0,0 +1,122 @@
...
+parser.add_argument("source")
+parser.add_argument("destination")
+args = parser.parse_args()
+
+# Visitor is an object that is currently responsible for handling new lines.
+# These objects are hierarchical: top level is host (Ironic node), second level
+# is specific source (file or command). Initially, visitor is None: such lines
+# are skipped.
+visitor = None
```
tuminoid, line 27: These global vars should be all uppercase like the ones on top.

tools/parse-ramdisk-logs.py
```diff
@@ -0,0 +1,122 @@
...
+        self.file = open(path, "wt")
+
+    def close(self):
+        self.file.close()
+
+    def __call__(self, line):
+        self.file.write(f"{line}\n")
+
+
+class VisitorSingleRun:
```
tuminoid, line 50: docstring missing

tools/parse-ramdisk-logs.py
```diff
@@ -0,0 +1,122 @@
...
+visitor = None
+
+# Skipped is the counter for initially skipped lines.Unfortunately, this may
+# include not just the inotify chatter but also useful lines if the logs have
+# been rotated. There is nothing that can be done here: the file name is only
+# available in the beginning of a dump.
+skipped = 0
+
+
+class VisitorEntry:
```
tuminoid, line 36: docstring missing

tuminoid (commented):
Feel free to ignore if not agreering, these are all nits. This is good improvement regardless.

tools/parse-ramdisk-logs.py
dtantsur: I can fix that, but note that an `if` condition does not create a new scope (unlike in Go and most compiled languages), so your solution won't improve anything. It has to be a new function.

MahnoorAsghar:
lgtm

tools/parse-ramdisk-logs.py
tuminoid: That is true, I did not want to say "also create def main()" but that is of course required part of this. I'm bad at half nitting :)

elfosardo:
/approve
we can do all the improvements in a follow up, it works as it is

dtantsur:
Yeah, I'm happy to follow-up if this change gets merged as it is

tuminoid:
> /approve we can do all the improvements in a follow up, it works as it is

Btw, I personally dislike this. Most of the time there is no time for a follow-up and the nits stay, because like you say "it works", why touch it? But its my 2 cents only, rant over :)

tuminoid:
> Yeah, I'm happy to follow-up if this change gets merged as it is

That is fine then.
/lgtm

dtantsur:
/override metal3-centos-e2e-integration-test-main
/override metal3-ubuntu-e2e-integration-test-main

Helper scripts, nothing to do with the CI

tools/parse-ramdisk-logs.py
dtantsur: I agree about reading files, but IMO we should use the default encoding when writing the result (which in most practical cases is still utf-8).

tools/parse-ramdisk-logs.py
dtantsur: "r" is the default but can add it for clarity

PR #681, opened 2025-05-12
🐛 OSSA-2025-001: set file_url_allowed_paths to what Metal3 uses

dtantsur:
The fix for OSSA-2025-001 introduces a new option that limits which
paths are allowed for file:// URLs. The default value already contains
the paths that Metal3 uses, but it will change in the future.

### DIFF
--- a/ironic-config/ironic.conf.j2
+++ b/ironic-config/ironic.conf.j2
@@ -96,6 +96,9 @@ deploy_ramdisk = file://{{ env.IRONIC_DEFAULT_RAMDISK }}
 {% if env.DISABLE_DEEP_IMAGE_INSPECTION | lower == "true" %}
 disable_deep_image_inspection = True
 {% endif %}
+# Allowed path for file:// links: ipa-downloader uses /shared/html/images,
+# while the bootloader configuration above refers to /templates.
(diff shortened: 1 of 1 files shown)


tuminoid:
https://security.openstack.org/ossa/OSSA-2025-001.html

Release-29.0 has similar new file paths, we want cherry-pick there right?
/cherry-pick release-29.0

It also seems to be valid for older ironic-images, but since they lack read-only filesystem changes, we need manual backports?

tuminoid:
> Release-29.0 has similar new file paths, we want cherry-pick there right?
> 
> Ideally. It's not critical since the Ironic defaults are good enough.

Let's just take the 29.0 cherry-pick as it should apply cleanly and there is no harm. 

> > It also seems to be valid for older ironic-images, but since they lack read-only filesystem changes, we need manual backports?
> 
> Correct. I'm not sure the effort is worth it.

```python
cfg.ListOpt('file_url_allowed_paths',
                default=['/var/lib/ironic', '/shared/html', '/templates',
                         '/opt/cache/files', '/vagrant'],
```

This default seems reasonable enough to me.

tuminoid (commented):
/lgtm

PR #686, opened 2025-05-19
:warning: Do not install separate sushy-oem-idrac

elfosardo:
The sushy-oem-idrac code is now included in sushy


lentzi90 (commented):
/lgtm

PR #691, opened 2025-06-02
:seedling: move functional test to larger github runner

tuminoid:
Functional test needs more disk space than is available in normal runner. Move it to larger runners.


tuminoid:
/hold
We'll track the configuration of larger runners in IRSO PR: https://github.com/metal3-io/ironic-standalone-operator/pull/278#issuecomment-2929991698

PR #701, opened 2025-06-25
📖 Update DHCP_RANGE description

lentzi90:
This has bitten me a couple of times recently so now I want to fix it.

There is no default, and the value should contain comma separated IPs (not <start>-<end> format).

Fixes #


tuminoid (commented):
/lgtm

PR #708, opened 2025-07-02
:bug: Fix dnsmasq config ipv6 issues

### LINKED ISSUE #614: dnsmasq.conf.j2 incomplete for IPv6 only setup
Running in an IPv6 only environment I run into a couple of problems and had a closer look at the dnsmasq.conf.j2 file:

1. "option6:router" is not a valid option [1] - but it is generated in line 19
2. Lines 65ff of the dnsmasq.conf.j2 which are mostly a (bad) copy of lines 14ff. The later one tries to be for IPv4 and IPv6 (which fails as described earlier) weather the first one is specific for IPv6, but does not even put the "option6" in place and also uses the wrong (IPv4) attributes.
3. There is IMHO no hint in the documentation how IPv6 addresses should be specified in the `dnsmasq.env` file. With or without brackets?

Depending on the input I get errors like

`dnsmasq: bad dhcp-option at line 14 of /etc/dnsmasq.conf`

where the line number differs depending on the set of input variables.

[1] see https://thekelleys.org.uk/gitweb/?p=dnsmasq.git;a=blob_plain;f=src/dhcp-common.c;hb=refs/heads/master lines 638ff

nuhakala:
The current dnsmasq configuration caused dnsmasq container to stop due to wrong configuration in IPv6 only environment. This PR fixes that issue and also removes duplicate block from the configuration.

Fixes #614

### DIFF
--- a/ironic-config/dnsmasq.conf.j2
+++ b/ironic-config/dnsmasq.conf.j2
@@ -12,20 +12,29 @@ port={{ env.DNS_PORT }}
 log-dhcp
 dhcp-range={{ env.DHCP_RANGE }}
 
-# It can be used when setting DNS or GW variables.
 {%- if env["GATEWAY_IP"] is undefined %}
 # Disable default router(s)
 dhcp-option=3
 {% else %}
-dhcp-option=option{% if ":" in env["GATEWAY_IP"] %}6{% endif %}:router,{{ env["GATEWAY_IP"] }}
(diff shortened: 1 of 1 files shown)


Rozzii (commented):
/lgtm
/cc @elfosardo  @iurygregory  @dtantsur

nuhakala:
ping @derekhiggins for approval

PR #710, opened 2025-07-04
🌱 skip detection step with an explicit var when running IPE script

pixelsoccupied:
Instead of dummy variable to satisfy existing scripts (heavily used by Ironic)...added an env that is used to optionally skip checks that are not needed to run IPE as a service.

Fixes # https://github.com/metal3-io/ironic-image/pull/700#discussion_r2173749861  

/cc @tuminoid 
/cc @iurygregory 
/cc @elfosardo 
/cc @dtantsur


scripts/configure-ironic.sh
```diff
@@ -52,7 +52,11 @@ export IRONIC_IPA_COLLECTORS=${IRONIC_IPA_COLLECTORS:-default,logs}
 wait_for_interface_or_ip
 
 # Hostname to use for the current conductor instance.
-export IRONIC_CONDUCTOR_HOST=${IRONIC_CONDUCTOR_HOST:-${IRONIC_URL_HOST}}
+if [[ "${SKIP_PROVISIONING_INTERFACE:-false}" == "true" ]]; then
+    export IRONIC_CONDUCTOR_HOST=${IRONIC_CONDUCTOR_HOST:-${IRONIC_URL_HOST:-localhost}}
+else
+    export IRONIC_CONDUCTOR_HOST=${IRONIC_CONDUCTOR_HOST:-${IRONIC_URL_HOST}}
+fi
```
tuminoid, line 59:
```suggestion
export SKIP_PROVISIONING_INTERFACE=${SKIP_PROVISIONING_INTERFACE:-false}
export IRONIC_CONDUCTOR_HOST=${IRONIC_CONDUCTOR_HOST:-${IRONIC_URL_HOST}}
if [[ "${SKIP_PROVISIONING_INTERFACE}" == "true" ]]; then
    export IRONIC_CONDUCTOR_HOST=${IRONIC_CONDUCTOR_HOST:-localhost}}
fi
```
Something like this would be easier to read and since ironic-common.sh is sourced into other scripts, avoids the `:-false` part on each if.

scripts/ironic-common.sh
```diff
@@ -35,6 +35,12 @@ export IRONIC_USE_MARIADB=${IRONIC_USE_MARIADB:-false}
 
 get_provisioning_interface()
 {
+    # Skip provisioning interface detection if requested (e.g., for exporter)
+    if [[ "${SKIP_PROVISIONING_INTERFACE:-false}" == "true" ]]; then
```
tuminoid, line 39:
```suggestion
    if [[ "${SKIP_PROVISIONING_INTERFACE}" == "true" ]]; then
```
If implemented as suggested in common.sh

tuminoid (commented):
Just some comments on the readability, not going to comment on the logic right now as I'm fresh back from vacations :)

/Cc @elfosardo @dtantsur

scripts/configure-ironic.sh
pixelsoccupied: had to clean it up a bit more make it work but PTAL thanks!

scripts/runironic-exporter
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/bash
 
-# Set dummy provisioning IP to avoid interface detection issues (not needed to run IPE to service `/metrics`)
-export PROVISIONING_IP="127.0.0.1"
+# Skip provisioning interface detection (not needed for exporter)
+export SKIP_PROVISIONING_INTERFACE=true
```
dtantsur, line 4: I think we need to go down a different path instead of adding yet another undocumented variable. I suggest splitting the provisioning interface logic into a new bash helper and not including it automatically from configure-ironic.

Note that regardless of the implementation, configure-ironic now needs to handle non-existing IRONIC_IP. This is the hardest part, but it's also the case here.

PR #712, opened 2025-07-09
🐛 Remove extra slash from deploy kernel/ramdisk paths

dtantsur:
Ironic's security checks don't recognize //shared as a valid prefix.

Follow-up to 315acdab32f296f850cd3961a5e0fde095567659.


dtantsur:

DNS

dtantsur:
> If the centos test passes I will override the ubuntu test as it failed for some unrelated DNS issue .

The CentOS job is badly affected by CentOS mirror issues because of the infrastructure migration, so I'm not sure it's going to pass any soon.

Issue #645, opened 2025-03-17
Support for ARM Architecture in ironic-images & Multi-Arch Compatibility


shibaPuppy:
Is there a plan to support ARM architecture in ironic-images?  

Currently, ironic-images primarily target x86_64 architecture. Are there any plans to introduce support for ARM-based architectures?  

Additionally, is there any plan to enable multi-architecture support within a single Ironic deployment?

tuminoid:
We have issue from last year about the same: https://github.com/metal3-io/ironic-image/issues/502 and some dabblings into it: https://github.com/metal3-io/ironic-image/pull/616

/cc @lentzi90 @Rozzii @dtantsur 
about the second part, multi-arch within single Ironic question.

dtantsur:
Just to clarify: we support single-arch x86_64 or aarch64, what we don't support is mixing them (e.g. installing an aarch64 host from an x86_64 cluster).

I hope that our team can look into multi-arch for at least virtual media (as the first step) within the coming months. Any help is welcome.

Rozzii:
/triage accepted
/kind feature
/help
We are also looking for folks who are willing to join an ARM working group and maintain ARM CI and babysit this whole domain.
Tentatively @dtantsur  is the contact person for the ARM working group but feel free to ping me also.

shibaPuppy:
Then, is there a plan to support multi-architecture in the long term?

lentzi90:
No concrete plan as far as I know, but the feature is accepted and contributions are welcome 🤗

Issue #615, opened 2025-02-03
Enable arm64 builds for sushy-tools container


e-minguez:
https://quay.io/repository/metal3-io/sushy-tools provides the sushy-tool containerized for easy usage. It would be nice if arm64 builds of that container image can be enabled.

e-minguez:
After reading https://github.com/metal3-io/ironic-image/blob/main/.github/workflows/build-images-action.yml#L29C11-L29C83 & https://github.com/metal3-io/project-infra/blob/main/.github/workflows/container-image-build.yml#L85C13-L85C52 it seems the builds are done by the https://github.com/mr-smithers-excellent/docker-build-push action which seems to support multi-platform builds per https://github.com/mr-smithers-excellent/docker-build-push?tab=readme-ov-file#multi-platform-builds

```
  - uses: mr-smithers-excellent/docker-build-push@v6
    name: Build & push Docker image
    with:
      image: repo/image
      registry: docker.io
      multiPlatform: true
      platform: linux/amd64,linux/arm64,linux/arm/v7
      username: ${{ secrets.DOCKER_USERNAME }}
      password: ${{ secrets.DOCKER_PASSWORD }}
```

Probably I'm missing something and it is not that easy 😅

Rozzii:
/triage accepted
Lets use this Issue and your PRs as a POC/experiment. We as a community has been long thinking on incorporating ARM support.
Please add documentation to the README in the root that ARM images are only experimental and not tested.

tuminoid:
/reopen

#616 was reverted.

Rozzii:
/remove-lifecycle stale
There was on failed attempt and  revert so this is WIP.

Issue #502, opened 2024-05-03
Support building for multiple architectures


lentzi90:
User Story

As a developer/user/operator I would like to use pre-built ironic images on multiple architectures because it is convenient (compared to building everything yourself).

Detailed Description
We currently only publish amd64 container images, but there are other popular architectures.
I suggest we look at how CAPI/CAPM3 has structured the Makefile with a docker-build-all target for building all supported architectures. It will probably be a bit different for Ironic though, since it is not written in Go.

Once we can build these images, it should be fairly trivial to also make CI build and publish them together with a multi-arch manifest for easy consumption.

/kind feature

dtantsur:
This is going to be a huge can of worms :) Start with https://github.com/metal3-io/ironic-image/pull/437 if you feel courageous.

Rozzii:
/tirage accepted
After we have sorted out our CI "restructuring" we could ask for some Oracle arm compute power from CNCF to have some ARM CI.
/help

diconico07:
To act as some kind of todo list for that, here are what I believe to be the needed elements to have a fully working multi-architecture support:
- [ ] IPA downloader grabs architecture specific images (handled in metal3-io/ironic-ipa-downloader#69)
- [ ] Ironic image entrypoint detects architecture specific images and fills the `_by_arch` Ironic config fields (handled in #713)
- [ ] Ironic image has EFI image for multiple architecture (also in #713)
- [ ] For correct (i)PXE support, per arch iPXE binaries and config are needed (no PR for now)

For now I focused entirely on refish virtual media support, so I didn't yet took a look at iPXE situation in Ironic configuration files

Rozzii:
> To act as some kind of todo list for that, here are what I believe to be the needed elements to have a fully working multi-architecture support:
> 
[...]

PR #713, opened 2025-07-17
:sparkles: Multi architecture support for IPA kernel/ramdisk and EFI bootloader

diconico07:
- Discovers architecture suffixed kernel and ramdisk (like produced by metal3-io/ironic-ipa-downloader#69 ), and adds them to ironic configuration as `_by_arch` configuration.
- Generates UEFI image for both x86_64 and aarch64.
- Discovers UEFI images for each architecture and populate the `bootloader_by_arch` configuration field with it.

I made the logic a bit more complicated in the `configure-ironic.sh` script so that we don't need to change the logic every time we would like to add more architectures to the mix.


Rozzii:
Could you @diconico07 please create a tracking issue where the community would be able to see all the changes related to multy arch, it can be a project level discussion also.

Anyhow I have found this too https://github.com/metal3-io/ironic-image/issues/615  that might be relevant for you eventually.

prepare-efi.sh
```diff
@@ -3,45 +3,64 @@
...
+ARCH="${HOST_ARCH}"
+build_efi
+
+# Then other architectures
+for ARCH in "${SUPPORTED_ARCH[@]}"; do
+    # Skip native one as we already did it
+    if [[ "$ARCH" != "$HOST_ARCH" ]]; then
+        build_efi
+    fi
+done
```
Rozzii, line 66: I don't think we need to separate the efi build into this host and non host builds, by definition it should be able to build efi for all the arches that are supported so if it fails with any of them then it is a legit ERROR IMO.

scripts/configure-ironic.sh
```diff
@@ -70,11 +70,48 @@ if [[ -z "${DEPLOY_KERNEL_URL:-}" ]] && [[ -z "${DEPLOY_RAMDISK_URL:-}" ]] && \
...
 fi
 
+declare -A detected_arch
+for var_arch in "${!DEPLOY_KERNEL_URL_@}"; do
+    IPA_ARCH="${var_arch#DEPLOY_KERNEL_URL}"
+    detected_arch["${IPA_ARCH,,}"]=1
+done
+for file_arch in "${IMAGE_CACHE_PREFIX}"_*.kernel; do
+    if [[ -f "${file_arch}" ]]; then
+        IPA_ARCH="$(basename "${file_arch#"${IMAGE_CACHE_PREFIX}"_}" .kernel)"
```
Rozzii, line 80: Are the two inner sets of double quotes intentional?

Rozzii (commented):
Some nits but generally I think this PR is OK.

diconico07:
@Rozzii would this fall into #645 or #502, or do you prefer I create a new issue for that ?
If not will add a comment on one of those to highlight all those changes.

Rozzii:
@diconico07  Let's use https://github.com/metal3-io/ironic-image/issues/502 , I have closed the other.

metal3-centos-e2e-integration-test-main and metal3-ubuntu-e2e-integration-test-main have to be triggered manually always.

In BMO we have integration and e2e tests that run automatically but in this repo we use a different CI system.

scripts/configure-ironic.sh
```diff
@@ -70,11 +70,48 @@ if [[ -z "${DEPLOY_KERNEL_URL:-}" ]] && [[ -z "${DEPLOY_RAMDISK_URL:-}" ]] && \
     export DEPLOY_RAMDISK_URL="file://${IMAGE_CACHE_PREFIX}.initramfs"
 fi
 
+declare -A detected_arch
```
dtantsur, line 73: nit: the entire logic is large enough to get split into a helper file

dtantsur: also, as an aside, this is exactly the level of bash complexity where I wish the entire thing was written in Python :)

diconico07:
@elfosardo thanks for your PR, it fixed the issue there and integration test is now green again.
@dtantsur (or anyone else), we may want to run the IrSO job again, but not sure if I can do it myself.

PR #714, opened 2025-07-25
:bug: Install pyasyncore compatibility package to fix pyinotify

elfosardo:


diconico07:
Looks good to me as an immediate fix, still thinks we should find an alternative to pyinotify in the long run as pyasyncore is really meant as a workaround while migrating to something else(will file an issue for that).

I don't think I'm authorized to do it, but will try anyway:
/lgtm

elfosardo:
@diconico07 feel free to propose an alternative

PR #716, opened 2025-07-30
:warning: WIP: Fully support dual-stack/dual socket setups

mchiappero:
The objective of this PR is to enable a proper dual-stack configuration by means of specific environment variables. This implies some preliminary refactoring and slight changes to the control flow in the startup scripts prior to the addition of some logic to determine and set, whenever possible, both an IPv4 and an IPv6 address. Also, introduce the possibility to use a hostname, a key element to achieve dual-stack URLs.

This PR is still WIP as 1) code must reviewed and consensus must be reached (including on breaking this PR into multiple ones) 2) little testing has been run 3) dnsmasq changes have not been produced yet 4) likely more documentation is needed

I would take the chance to highlight the need for proper documentation of the binding logic as well as proper input validation and logging: variables like PROVISIONING_IP, PROVISIONING_INTERFACE, PROVISIONING_MACS (and the new IRONIC_URL_HOSTNAME) can override each other or, in some cases, be optional, but no such logic is clearly stated. Last but not least, at this point bash is showing its limits and it could make sense to start migrating some of this potentially new code to a different programming language.

Fixes #

### DIFF
--- a/ironic-config/apache2-ipxe.conf.j2
+++ b/ironic-config/apache2-ipxe.conf.j2
@@ -1,4 +1,5 @@
-Listen {{ env.IPXE_TLS_PORT }}
+Listen 0.0.0.0:{{ env.IPXE_TLS_PORT }}
+Listen [::]:{{ env.IPXE_TLS_PORT }}
 
 <VirtualHost *:{{ env.IPXE_TLS_PORT }}>
     ErrorLog /dev/stderr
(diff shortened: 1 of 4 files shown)


tuminoid:
/cc @nuhakala 
This IPv6 change could interest you too, PTAL Nuutti.

/cc @dtantsur @elfosardo

lentzi90:
@mchiappero could you check the failing shellcheck?

mchiappero:
> I am not the most familiar with this project but at this point I am not sure if there even is any certain logic behind the env variables. The variables and logic are most likely added to cover some edge cases when needed and hence they can overwrite each other.

In my opinion, it wouldn't be bad to provide clear error messages or warning regarding conflicting or overlapping input, but also maybe a table in the official documentation better explaining the logic and priorities of these control variables. Of course it's not that handy in bash (and shifting to something else, e.g. python, is not effort free but definitely possible), but I might find some time to contribute to this if there is some consensus.

PR #723, opened 2025-08-08
:bug: Revert "Fix dnsmasq config ipv6 issues"

elfosardo:
Reverts metal3-io/ironic-image#708

After testing this patch on an actual ipv6 deployment, we've verified that this change breaks dnsmasq and ip assignment
The router option does work as expected instead
```
dnsmasq: started, version 2.85 DNS disabled
dnsmasq: compile time options: IPv6 GNU-getopt DBus no-UBus no-i18n IDN2 DHCP DHCPv6 no-Lua TFTP no-conntrack ipset auth cryptohash DNSSEC loop-detect inotify dumpfile
dnsmasq-dhcp: DHCPv6, IP range fd00:1101::a -- fd00:1101::ffff:ffff:ffff:fffe, lease time 1d
dnsmasq-dhcp: router advertisement on fd00:1101::
dnsmasq-dhcp: RTR-ADVERT(enp1s0) fd00:1101::
dnsmasq-dhcp: IPv6 router advertisement enabled
dnsmasq-dhcp: DHCP, sockets bound exclusively to interface enp1s0
```

### DIFF
--- a/ironic-config/dnsmasq.conf.j2
+++ b/ironic-config/dnsmasq.conf.j2
@@ -12,29 +12,20 @@ port={{ env.DNS_PORT }}
 log-dhcp
 dhcp-range={{ env.DHCP_RANGE }}
 
+# It can be used when setting DNS or GW variables.
 {%- if env["GATEWAY_IP"] is undefined %}
 # Disable default router(s)
 dhcp-option=3
 {% else %}
-# Setting default router is not supported in v6, there are router advertisements
-# for that. In v4 we set the router option.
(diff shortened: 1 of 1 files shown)


Rozzii:
@nuhakala

Rozzii:
Let's have a discussion about this because in case https://github.com/metal3-io/ironic-image/pull/708 is faulty we have a faulty way to deploy ipv6 only environment in dev-env too and need to fix how we evaluate ipv6 fixes.

elfosardo:
/hold
until we have discussed this

PR #724, opened 2025-08-08
fix(dnsmasq): add missing set

t-h2o:
# DHCP configuration

- **fix(dnsmasq): add missing set**
- **feat(dnsmasq): tag bios machine**


elfosardo:
@t-h2o  is this related to a bug/issue?
if yes, please open an issue with an explanation why this is needed, thanks

Rozzii:
Could you please explain your use case @t-h2o  ? We never had these settings and ipv4 bios deployments worked just fine .

PR #727, opened 2025-08-11
:bug: Fix invalid and duplicate code in dnsmasq config

nuhakala:
The current dnsmasq config contains wrong configuration option `option6:router` and it also has duplicated code block.

### DIFF
--- a/ironic-config/dnsmasq.conf.j2
+++ b/ironic-config/dnsmasq.conf.j2
@@ -12,13 +12,6 @@ port={{ env.DNS_PORT }}
 log-dhcp
 dhcp-range={{ env.DHCP_RANGE }}
 
-# It can be used when setting DNS or GW variables.
-{%- if env["GATEWAY_IP"] is undefined %}
-# Disable default router(s)
-dhcp-option=3
-{% else %}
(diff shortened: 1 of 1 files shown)


ironic-config/dnsmasq.conf.j2
```diff
@@ -12,20 +12,23 @@ port={{ env.DNS_PORT }}
...
-# It can be used when setting DNS or GW variables.
 {%- if env["GATEWAY_IP"] is undefined %}
 # Disable default router(s)
 dhcp-option=3
 {% else %}
-dhcp-option=option{% if ":" in env["GATEWAY_IP"] %}6{% endif %}:router,{{ env["GATEWAY_IP"] }}
+# Setting default router is not supported in v6, there are router advertisements
+# for that. In v4 we set the router option.
+dhcp-option=option:router,{{ env["GATEWAY_IP"] }}
 {% endif %}
```
nuhakala, line 22: TBH, I don't know if we should move this under the IPv4 section, because I think this is IPv4 specific. Moving it would make the config more clear.

Rozzii: feel free to move it

nuhakala: Done

Rozzii (commented):
/lgtm

nuhakala:
@elfosardo please unhold when finished with testing

PR #728, opened 2025-08-11
:bug: Centralize assignment of HTTP_PORT

elfosardo:
This is actually a bug fix, as ironic-common.sh, when invoked without setting HTTP_PORT, will fail, as no default value is defined.
We set HTTP_PORT default value in ironic-common.sh to 80 as correctly written in the docs, and then export it.
Removing assignments from all other scripts to avoid mistakes and confusion.


tuminoid (commented):
/lgtm

All of the script source ironic-common.sh so good simplification.

Rozzii (commented):
/lgtm

PR #730, opened 2025-08-18
:warning: Use watchdog to monitor logs and certificate changes

### LINKED ISSUE #715: Migrate from pyinotify to another file watching module/cli
As noticed recently, pyinotify is unmaintained and incompatible with recent python versions (without some workaround). We should try and migrate to another implementation.

https://pypi.org/project/watchdog/ may be a viable alternative to pyinotify

Some discussion already happened in #714 that implemented the said workaround

elfosardo:
Watchdog comes with a shell utility script, watchmedo, that can perfectly replace pyinotify, and it works natively with Python 3.12
We drop pyinotify in favor of watchdog and convert the scripts to use watchmedo.

Fixes #715

### DIFF
--- a/prepare-image.sh
+++ b/prepare-image.sh
@@ -40,8 +40,7 @@ if [[ ! -s "${UPPER_CONSTRAINTS_PATH}" ]]; then
 fi
 
 # NOTE(elfosardo): install dependencies constrained
-# also install pyasyncore to allow asyncore in Python 3.12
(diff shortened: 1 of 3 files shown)


elfosardo:
the failure is really weird and I can't get much from the logs, seems something related to TLS

dtantsur:
/approve

I hope the TLS reloading thing works, I have no means to verify it

tuminoid (commented):
This is so ugly now :)  Let's clean it up in follow-up before next release.

/lgtm

PR #734, opened 2025-08-20
:bug: Fix CentOS Stream 10 build

elfosardo:
During efi build, remove the shim package but not its
dependencies to avoid conflict between architectures.

This has been tested locally with both CS9 and CS10
base images.


diconico07:
lgtm as well, I only tested with CS9 initially.

PR #747, opened 2025-08-25
:seedling: build ironic-image on both centos:stream versions

tuminoid:
Build ironic-image on both centos:stream versions. We build both in our post merge step, but we don't build cs10 version anywhere else. This leads to situations where we merge stuff that breaks the build.

This will be accompanied by project-infra PR that will run "make build" to run the test.

Also, rename the Makefile variable CONTAINER_ENGINE to CONTAINER_RUNTIME as we use CONTAINER_RUNTIME everywhere else as the engine selector.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,14 +2,15 @@ SHELL=/usr/bin/env bash -o errexit
 
 .PHONY: help build
 
-export CONTAINER_ENGINE ?= podman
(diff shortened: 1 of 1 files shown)


tuminoid:
This is implemented via GH workflow here: https://github.com/metal3-io/ironic-image/pull/763
/close

PR #756, opened 2025-09-18
:sparkles: Use the shared directory to store master iso images on ironic-conductor node

hroyrh:
When read-only root filesystem is enabled for ironic-image based containers, the path /var/lib/ironic/master_iso_images becomes read-only. We need to change the path to `/shared/html/master_iso_images` directory so that changes can be made to this path for updating image cache and hard links can be used.


dtantsur:
Re ✨: I think this change qualifies as a 🐛 fix.

hroyrh:
Yeah, that makes sense.

ironic-config/ironic.conf.j2
```diff
@@ -220,6 +220,7 @@ backend = collector
 boot_retry_timeout = 1200
 images_path = /shared/html/tmp
 instance_master_path = /shared/html/master_images
+iso_master_path = /shared/html/master_iso_images
```
tuminoid, line 223: Q: Can it not be the same `master_images` dir?

tuminoid (commented):
/lgtm

PR #775, opened 2025-10-29
:seedling: Do not install dnf config-manager

elfosardo:
We do not use it anymore.


tuminoid (commented):
/lgtm

PR #788, opened 2025-11-14
:warning: Improve priority and validation of input variables consumed in wait_for_interface_or_ip

mchiappero:
**What this PR does / why we need it**:
The current logic in wait_for_interface_or_ip deals with the input IP addresses or interfaces, according to the following order: PROVISIONING_IP, IRONIC_IP, PROVISIONING_INTERFACE. The improvements this PR introduces are:
- make the test case for IRONIC_IP explicit rather than implicit, improving code readability before the following change
- "group" PROVISIONING_IP and PROVISIONING_INTERFACE in terms of priority by prioritizing IRONIC_IP instead, resulting in the following order: IRONIC_IP, PROVISIONING_IP, PROVISIONING_INTERFACE
- include input validation for the externally provided IRONIC_IP

What the PR does not address is logging a warning in case of multiple input variables set and one taking inevitably precedence over the another(s).

NOTE: based on PR #787

**Checklist:**

- [ ] Documentation has been updated, if necessary.
- [ ] Integration tests have been added, if necessary.

PR #789, opened 2025-11-14
:sparkles: Support both IPv4 and IPv6 addresses in startup scripts and configuration files

mchiappero:
Currently the configuration and startup scripts rely on the assumption that all the services should either be exposed on an IPv4 or an IPv6 address. Dual-stack support is somewhat provided by leveraging IPv4-mapped IPv6 addresses, that is, only an IPv6 socket is allocated (which is a discouraged practice). Enable true dual-stack support, both in ironic.conf and all the Apache instances, by introducing IRONIC_IPV6 alongside the existing IRONIC_IP, so that either or both IPs can be leverage when defined or detected.
To avoid breaking changes and keep the PR simple, externally provided IRONIC_IP and PROVISIONING_IP values will result in either an IPv4 or an IPv6 socket allocated and not both. However, when PROVISIONING_INTERFACE must be used, the new code will attempt to detect both IP versions, potentially resulting in a proper dual-stack configuration on a dual-stack system. In this case, IPv6 will however take precedence for URL generation.

This PR paves the way to a future PR, where a hostname, resolving to both IPv4 and IPv6 addresses, can be leveraged for "dual-stack URLs" and, potentially, in a dual socket configuration. Besides this possibility, an initial conversation on other approaches or improvements can also get started here.

NOTE: based on PR #788

**Checklist:**

### DIFF
--- a/ironic-config/apache2-ipxe.conf.j2
+++ b/ironic-config/apache2-ipxe.conf.j2
@@ -1,4 +1,5 @@
-Listen {{ env.IPXE_TLS_PORT }}
+Listen 0.0.0.0:{{ env.IPXE_TLS_PORT }}
+Listen [::]:{{ env.IPXE_TLS_PORT }}
 
 <VirtualHost *:{{ env.IPXE_TLS_PORT }}>
     ErrorLog /dev/stderr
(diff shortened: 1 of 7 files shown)


tuminoid:
Before I even take a look, please squash the commits, there is no reason to have 12.

PR #791, opened 2025-11-14
:sparkles: Allow IRONIC_IPV4, IRONIC_IPV6 and URLs to be derived from a DNS record

mchiappero:
PR #789 introduced the ability to create IPv4 and IPv6 sockets for Apache, but is limited in the ability to provide true dual-stack behavior with respect to the Ironic API endpoints. Introduce the a new environment variable, IRONIC_URL_HOST, which can contain an hostname which resolves to at least an IPv4 or an IPv6 address. When both are present, the client has the option to connect with whichever version is actually in use or preferred.

This design allows to also bind the servers to the IP addresses the hostname resolves to, when LISTEN_ALL_INTERFACES is not true, which means that such IPs will be searched for on the system when starting. Setting LISTEN_ALL_INTERFACES to true results in Apache binding on IPv4 and/or IPv6 according to the resolved IPs, but not to look for those IPs on the systems. This is to mimic the behavior of IRONIC_IP.

NOTE: depends on PR #1060.

**Checklist:**

### DIFF
--- a/ironic-config/httpd-ironic-api.conf.j2
+++ b/ironic-config/httpd-ironic-api.conf.j2
@@ -15,8 +15,14 @@
 Listen {{ env.IRONIC_LISTEN_PORT }}
  <VirtualHost *:{{ env.IRONIC_LISTEN_PORT }}>
 {% else %}
-Listen {{ env.IRONIC_URL_HOST }}:{{ env.IRONIC_LISTEN_PORT }}
- <VirtualHost {{ env.IRONIC_URL_HOST }}:{{ env.IRONIC_LISTEN_PORT }}>
+{% if env.ENABLE_IPV4 %}
+Listen {{ env.IRONIC_IP }}:{{ env.IRONIC_LISTEN_PORT }}
+{% endif %}
+{% if env.ENABLE_IPV6 %}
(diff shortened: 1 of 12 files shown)


mchiappero:
I'm setting it to WIP as I haven't thoroughly tested yet (and the documentation hasn't been updated), but should be mostly bug free. Feedback on the design is very welcome in the meantime.

ironic-config/httpd-ironic-api.conf.j2
```diff
@@ -22,6 +22,9 @@ Listen {{ env.IRONIC_IP }}:{{ env.IRONIC_LISTEN_PORT }}
 {% if env.ENABLE_IPV6 %}
 Listen [{{ env.IRONIC_IPV6 }}]:{{ env.IRONIC_LISTEN_PORT }}
 {% endif %}
+{% if env.IRONIC_URL_HOSTNAME is defined and env.IRONIC_URL_HOSTNAME|length %}
+<VirtualHost {{ env.IRONIC_URL_HOSTNAME }}:{{ env.IRONIC_LISTEN_PORT }}>
+{% else %}
```
mchiappero: I guess we might also include the IP based VirtualHosts rather than have mutual exclusion. What do you think? Need to check if multiple VirtualHost directives would work...

PR #807, opened 2025-11-25
:bug: Set the default IRONIC_CACERT_FILE path under /conf for readOnlyFilesystem

hroyrh:
Right now, we have `IRONIC_CACERT_FILE=/certs/ca/ironic/tls.crt` which creates an issue, due to that path being readOnly, when [copying cert to ca_cert](https://github.com/metal3-io/ironic-image/blob/main/scripts/tls-common.sh#L74) in case of self-signed certs.

### DIFF
--- a/scripts/tls-common.sh
+++ b/scripts/tls-common.sh
@@ -70,6 +70,8 @@ if [[ -f "$IRONIC_CERT_FILE" ]] || [[ -f "$IRONIC_CACERT_FILE" ]]; then
     export IRONIC_TLS_SETUP="true"
     export IRONIC_SCHEME="https"
     if [[ ! -f "$IRONIC_CACERT_FILE" ]]; then
+        # For missing cacert file, change the var to writable path
+        export IRONIC_CACERT_FILE=/conf/certs/ca/ironic/tls.crt
         mkdir -p "$(dirname "${IRONIC_CACERT_FILE}")"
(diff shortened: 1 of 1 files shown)


scripts/tls-common.sh
```diff
@@ -22,7 +22,11 @@ export RESTART_CONTAINER_CERTIFICATE_UPDATED=${RESTART_CONTAINER_CERTIFICATE_UPD
 export MARIADB_CACERT_FILE=/certs/ca/mariadb/tls.crt
 export BMC_CACERTS_PATH=/certs/ca/bmc
 export BMC_CACERT_FILE=/conf/bmc-tls.pem
-export IRONIC_CACERT_FILE=/certs/ca/ironic/tls.crt
+# Set the default IRONIC_CACERT_FILE directory under CUSTOM_CONFIG_DIR=/conf,
+# which is also used in ironic-common.sh
+# This is for readOnlyRootFilesystem=true, since IRONIC_CACERT_FILE is
+# copied over during runtime
+export IRONIC_CACERT_FILE=${IRONIC_CACERT_FILE:-/conf/certs/ca/ironic/tls.crt}
```
tuminoid: I don't think we need this comment

hroyrh: Done

tuminoid (commented):
/lgtm

Rozzii (commented):
/lgtm

PR #812, opened 2025-11-28
🌱Add yamllint workflow

smoshiur1237:
Here it will add yamllint workflow by reusing the workflow in Project-infra.


Rozzii (commented):
/lgtm

PR #821, opened 2025-12-03
:bug: Run EFI build loop based on actual arch

elfosardo:
This is to avoid conflicts between installed packages.


Rozzii (commented):
/lgtm

PR #822, opened 2025-12-04
:seedling: Pin constraints, prepare release-33.0

elfosardo:
upper-constraints from https://opendev.org/openstack/requirements/src/commit/44dbd103640a38272df3ad29048b4c5dece2c1c7/upper-constraints.txt


elfosardo:
yep we should wait for https://github.com/metal3-io/project-infra/pull/1189

PR #824, opened 2025-12-05
:seedling: Remove unneeded mkdirs from dnsmasq

### LINKED ISSUE #819: We should not execute the command such as "mkdir -p /shared/html/images" in `dnsmasq`
I think we should remove some commend in ironic-image/scripts/dnsmasq. those commends has no value for dnsmasq.
```shell
mkdir -p /shared/html/images
```

Image

matthewei:
Cleaning up unnecessary commands to make our engineering code more concise and reduce the learning cost for users.

Fixes #819


tuminoid:
Squash commits please.
/hold

tuminoid (commented):
/lgtm

PR #842, opened 2025-12-11
:seedling: add re-approve workflow for github workflows

tuminoid:
Since we have IRSO functional and other workflows here now, add the re-approver workflow so the workflows re-run automatically if ok-to-test is present. This affects non-member contributors.


tuminoid:
/cc @lentzi90 @Rozzii @elfosardo @dtantsur

PR #851, opened 2025-12-15
:sparkles: Inject Ironic CA Cert in IPA

diconico07:

This PR adds a way to inject the Ironic CA as well as additional CA certificates into the IPA image, it allows removing the `ipa-insecure` flag and actually validate the Ironic certificate from the IPA.

This solution should work for both virtualmedia and PXE based deployments.

**Checklist:**

### DIFF
--- a/ironic-config/inspector.ipxe.j2
+++ b/ironic-config/inspector.ipxe.j2
@@ -3,7 +3,7 @@
 echo In inspector.ipxe
 
 {%- macro kernel_cmdline(kernel_path, ramdisk_name) -%}
-kernel --timeout 60000 {{ kernel_path }} ipa-insecure={{ env.IRONIC_IPA_INSECURE | default('1') }} ipa-inspection-collectors={{ env.IRONIC_IPA_COLLECTORS }} systemd.journald.forward_to_console={{ env.IPA_FORWARD_CONSOLE | default('yes') }} BOOTIF=${mac} ipa-debug={{ env.IRONIC_IPA_DEBUG | default('1') }} ipa-enable-vlan-interfaces={{ env.IRONIC_ENABLE_VLAN_INTERFACES }} ipa-inspection-dhcp-all-interfaces={{ env.IRONIC_IPA_INSPECTION_DHCP_ALL_INTERFACES | default('1') }} ipa-collect-lldp={{ env.IRONIC_IPA_COLLECT_LLDP | default('1') }} {{ env.INSPECTOR_EXTRA_ARGS }} initrd={{ ramdisk_name }} {% if env.IRONIC_RAMDISK_SSH_KEY %}sshkey="{{ env.IRONIC_RAMDISK_SSH_KEY|trim }}"{% endif %} {{ env.IRONIC_KERNEL_PARAMS|trim }} || goto retry_boot
+kernel --timeout 60000 {{ kernel_path }} ipa-insecure={{ env.IRONIC_IPA_INSECURE | default('1') }} ipa-inspection-collectors={{ env.IRONIC_IPA_COLLECTORS }} systemd.journald.forward_to_console={{ env.IPA_FORWARD_CONSOLE | default('yes') }} BOOTIF=${mac} ipa-debug={{ env.IRONIC_IPA_DEBUG | default('1') }} ipa-enable-vlan-interfaces={{ env.IRONIC_ENABLE_VLAN_INTERFACES }} ipa-inspection-dhcp-all-interfaces={{ env.IRONIC_IPA_INSPECTION_DHCP_ALL_INTERFACES | default('1') }} ipa-collect-lldp={{ env.IRONIC_IPA_COLLECT_LLDP | default('1') }} {{ env.INSPECTOR_EXTRA_ARGS }} initrd={{ ramdisk_name }} {% if env.IRONIC_INJECT_IPA == 'true' %}initrd=ipa-cacert-bundle{% endif %} {% if env.IRONIC_RAMDISK_SSH_KEY %}sshkey="{{ env.IRONIC_RAMDISK_SSH_KEY|trim }}"{% endif %} {{ env.IRONIC_KERNEL_PARAMS|trim }} || goto retry_boot
 {%- endmacro -%}
 
 {% if env.DEPLOY_KERNEL_BY_ARCH is defined %}
@@ -55,6 +55,9 @@ imgfree
 # ironic-inspector-image and configuration in configure-ironic.sh
 {{ kernel_cmdline('${ipa_kernel}', '${ipa_ramdisk_name}') }}
 initrd --timeout 60000 ${ipa_ramdisk} || goto retry_boot
+{%- if env.IRONIC_INJECT_IPA == 'true' %}
+initrd --timeout 60000 {{ env.IRONIC_IPA_BASE_URL }}/ipa-cacert-bundle || goto retry_boot
+{%- endif %}
 boot
 {% else %}
 :retry_boot
@@ -63,5 +66,8 @@ imgfree
 # ironic-inspector-image and configuration in configure-ironic.sh
 {{ kernel_cmdline(env.IRONIC_IPA_BASE_URL + '/images/ironic-python-agent.kernel', 'ironic-python-agent.initramfs') }}
 initrd --timeout 60000 {{ env.IRONIC_IPA_BASE_URL }}/images/ironic-python-agent.initramfs || goto retry_boot
+{%- if env.IRONIC_INJECT_IPA == 'true' %}
+initrd --timeout 60000 {{ env.IRONIC_IPA_BASE_URL }}/ipa-cacert-bundle || goto retry_boot
+{%- endif %}
 boot
 {% endif %}
--- a/ironic-config/ipxe_config.template
+++ b/ironic-config/ipxe_config.template
@@ -7,21 +7,29 @@ goto deploy
 
 :deploy
 imgfree
+{%- if ipxe_tls_setup %}
 {%- if pxe_options.deployment_aki_path %}
 {%- set aki_path_https_elements = pxe_options.deployment_aki_path.split(':') %}
 {%- set aki_port_and_path = aki_path_https_elements[2].split('/') %}
 {%- set aki_afterport = aki_port_and_path[1:]|join('/') %}
-{%- set aki_path_https = ['https:', aki_path_https_elements[1], ':8084/', aki_afterport]|join %}
+{%- set aki_path = ['https:', aki_path_https_elements[1], ':8084/', aki_afterport]|join %}
 {%- endif %}
 {%- if pxe_options.deployment_ari_path %}
 {%- set ari_path_https_elements = pxe_options.deployment_ari_path.split(':') %}
 {%- set ari_port_and_path = ari_path_https_elements[2].split('/') %}
 {%- set ari_afterport = ari_port_and_path[1:]|join('/') %}
-{%- set ari_path_https = ['https:', ari_path_https_elements[1], ':8084/', ari_afterport]|join %}
+{%- set ari_path = ['https:', ari_path_https_elements[1], ':8084/', ari_afterport]|join %}
 {%- endif %}
-kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path_https }} selinux=0 troubleshoot=0 text {{ pxe_options.pxe_append_params|default("", true) }} BOOTIF=${mac} initrd={{ pxe_options.initrd_filename|default("deploy_ramdisk", true) }} || goto retry
+{%- else %}
+{%- set ari_path = pxe_options.deployment_ari_path %}
+{%- set aki_path = pxe_options.deployment_aki_path %}
+{%- endif %}
+kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path }} selinux=0 troubleshoot=0 text {{ pxe_options.pxe_append_params|default("", true) }} BOOTIF=${mac} initrd={{ pxe_options.initrd_filename|default("deploy_ramdisk", true) }} {% if inject_cacert_bundle %}initrd=ipa-cacert-bundle{% endif %} || goto retry
 
-initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path_https }} || goto retry
+initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path }} || goto retry
+{%- if inject_cacert_bundle %}
+initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %} ../ipa-cacert-bundle || goto retry
+{%- endif %}
 boot
 
 :retry
@@ -39,8 +47,8 @@ poweroff
 
 :boot_anaconda
 imgfree
-kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path_https }} text {{ pxe_options.pxe_append_params|default("", true) }} inst.ks={{ pxe_options.ks_cfg_url }} {% if pxe_options.repo_url %}inst.repo={{ pxe_options.repo_url }}{% else %}inst.stage2={{ pxe_options.stage2_url }}{% endif %} initrd=ramdisk || goto boot_anaconda
-initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path_https }} || goto boot_anaconda
+kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path }} text {{ pxe_options.pxe_append_params|default("", true) }} inst.ks={{ pxe_options.ks_cfg_url }} {% if pxe_options.repo_url %}inst.repo={{ pxe_options.repo_url }}{% else %}inst.stage2={{ pxe_options.stage2_url }}{% endif %} initrd=ramdisk || goto boot_anaconda
+initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path }} || goto boot_anaconda
 boot
 
 :boot_ramdisk
@@ -48,8 +56,8 @@ imgfree
 {%- if pxe_options.boot_iso_url %}
 sanboot {{ pxe_options.boot_iso_url }}
 {%- else %}
-kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path_https }} root=/dev/ram0 text {{ pxe_options.pxe_append_params|default("", true) }} {{ pxe_options.ramdisk_opts|default('', true) }} initrd=ramdisk || goto boot_ramdisk
-initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path_https }} || goto boot_ramdisk
+kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path }} root=/dev/ram0 text {{ pxe_options.pxe_append_params|default("", true) }} {{ pxe_options.ramdisk_opts|default('', true) }} initrd=ramdisk || goto boot_ramdisk
+initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path }} || goto boot_ramdisk
 boot
 {%- endif %}
 
--- a/ironic-config/ironic.conf.j2
+++ b/ironic-config/ironic.conf.j2
@@ -62,6 +62,9 @@ deploy_logs_local_path = /shared/log/ironic/deploy
 # See https://bugzilla.redhat.com/show_bug.cgi?id=1822763
 max_command_attempts = 30
 certificates_path = {{ env.IRONIC_GEN_CERT_DIR }}
+{% if env.IRONIC_INJECT_IPA == 'true' %}
+api_ca_file = {{ env.IPA_CACERT_FILE }}
+{% endif %}
 
 [api]
 {% if env.IRONIC_REVERSE_PROXY_SETUP == "true" %}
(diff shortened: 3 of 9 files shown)


diconico07:
Putting as draft for now, as I didn't get time to test it works as expected in all situations (in particular I'll try to setup a PXE environment to test it works with PXE). 
This is not the solution we currently use at SUSE to get rid of the `ipa-insecure` flag (we do the injection in the ipa downloader), but this one should be more future-proof as we can get IPA images that are not coming from the ipa downloader.

scripts/tls-common.sh
```diff
@@ -129,3 +133,32 @@ if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
...
+    cd "${temp_dir}" || return
+
+    mkdir -p etc/ironic-python-agent.d etc/ironic-python-agent
+    cat > etc/ironic-python-agent.d/ironic-tls.conf <<EOF
+[DEFAULT]
+cafile = /etc/ironic-python-agent/ironic.crt
+EOF
+    cp "${IPA_CACERT_FILE}" etc/ironic-python-agent/ironic.crt
+
+    find . | cpio -o -H newc --reproducible >> "${output_path}"
```
Rozzii: Just checking.
If I remember correctly you have mentioned this only works with an initrd that is compressed with "zstd" or some other specific compression method right?

diconico07: Not this one, the way we currently do it at SUSE in the IPA downloader only works with "zstd" or other compression methods where you can simply append to the file.
Here it is a different initrd file that is loaded by PXE in addition to the IPA one (I do not compress it actually since I don't think it would be relevant, but we could also compress it here if needed)

ironic-config/ironic.conf.j2
```diff
@@ -240,14 +241,14 @@ ipxe_config_template = /templates/ipxe_config.template
 
 [redfish]
 use_swift = false
-kernel_append_params = nofb nomodeset vga=normal ipa-insecure=1 {% if env.ENABLE_FIPS_IPA %}fips={{ env.ENABLE_FIPS_IPA|trim }}{% endif %} {% if env.IRONIC_RAMDISK_SSH_KEY %}sshkey="{{ env.IRONIC_RAMDISK_SSH_KEY|trim }}"{% endif %} {{ env.IRONIC_KERNEL_PARAMS|trim }} systemd.journald.forward_to_console=yes
+kernel_append_params = nofb nomodeset vga=normal {% if env.IRONIC_INSECURE %}ipa-insecure=1{% endif %} {% if env.ENABLE_FIPS_IPA %}fips={{ env.ENABLE_FIPS_IPA|trim }}{% endif %} {% if env.IRONIC_RAMDISK_SSH_KEY %}sshkey="{{ env.IRONIC_RAMDISK_SSH_KEY|trim }}"{% endif %} {{ env.IRONIC_KERNEL_PARAMS|trim }} systemd.journald.forward_to_console=yes
```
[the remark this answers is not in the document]
diconico07: This is on purpose as the additional initrd is not required for virtualmedia, the api_ca_file field handles the injection work here

scripts/tls-common.sh
```diff
@@ -129,3 +133,37 @@ if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
...
+    cd "${temp_dir}" || return
+
+    mkdir -p etc/ironic-python-agent.d etc/ironic-python-agent
+    cp "${IPA_CACERT_FILE}" etc/ironic-python-agent/ironic.crt
+    cat > etc/ironic-python-agent.d/ironic-tls.conf <<EOF
+[DEFAULT]
+cafile = /etc/ironic-python-agent/ironic.crt
+EOF
+
+    find . | cpio -o -H newc --reproducible >> "${output_path}"
```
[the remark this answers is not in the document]
diconico07: not relevant to us

scripts/tls-common.sh
```diff
@@ -129,3 +133,37 @@ if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
...
+
+    # Remove temp directory
+    cd && rm -rf "${temp_dir}"
```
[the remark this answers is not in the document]
diconico07: the `&&` prevents that

ironic-config/ipxe_config.template
```diff
@@ -7,21 +7,11 @@ goto deploy
...
-{%- set aki_port_and_path = aki_path_https_elements[2].split('/') %}
-{%- set aki_afterport = aki_port_and_path[1:]|join('/') %}
-{%- set aki_path_https = ['https:', aki_path_https_elements[1], ':8084/', aki_afterport]|join %}
-{%- endif %}
-{%- if pxe_options.deployment_ari_path %}
-{%- set ari_path_https_elements = pxe_options.deployment_ari_path.split(':') %}
-{%- set ari_port_and_path = ari_path_https_elements[2].split('/') %}
-{%- set ari_afterport = ari_port_and_path[1:]|join('/') %}
-{%- set ari_path_https = ['https:', ari_path_https_elements[1], ':8084/', ari_afterport]|join %}
-{%- endif %}
```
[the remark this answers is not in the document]
diconico07: To properly inject the CA when using iPXE it needs to always use the custom templated version (see the change in `ironic.conf.j2`).
I then remembered #898 that stated this is no longer needed, if this is not true (wasn't able to check as I don't have TLS enabled on my local PXE setup), then we'll likely need to use a different template whether we are in TLS enabled setup or not.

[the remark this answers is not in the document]
diconico07: Looked again, more closely and found it safer to go back to having this, put a conditional to only use it when we're in HTTPS iPXE (look at the change in `runironic` for how it's handled)

scripts/tls-common.sh
```diff
@@ -129,3 +132,42 @@ if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
...
+    cp "${IPA_CACERT_FILE}" etc/ironic-python-agent/ironic.crt
+    cat > etc/ironic-python-agent.d/ironic-tls.conf <<EOF
+[DEFAULT]
+cafile = /etc/ironic-python-agent/ironic.crt
+EOF
+
+    find . | cpio -o -H newc -R +0:+0 --reproducible >> "${output_path}"
+
+    # Remove temp directory
+    cd && rm -rf "${temp_dir}"
```
elfosardo: maybe let's use pushd/popd instead of cd ?
although I think you don't need `cd` here as you're running in a subshell, you could set a trap in the function to just remove the temp_dir, which is more robust because it will run even if the script fails

tuminoid (commented):
Some nits, but some concerns as well. Hope you can follow the chain of though through the review comments (I use ... to signal the next step is next comment)

scripts/tls-common.sh
```diff
@@ -129,3 +132,42 @@ if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
...
+generate_cacert_bundle_initrd()
+(
+    local output_path="$1"
+    local temp_dir
+
+    temp_dir="$(mktemp -d)"
+
+    chmod 0755 "${temp_dir}" || return
+
+    cd "${temp_dir}" || return
```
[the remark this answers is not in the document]
diconico07: Removed the returns and made the whole subshell `set -euo pipefail`

ironic-config/ipxe_config.template
```diff
@@ -39,17 +46,19 @@ poweroff
 
 :boot_anaconda
 imgfree
-kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path_https }} text {{ pxe_options.pxe_append_params|default("", true) }} inst.ks={{ pxe_options.ks_cfg_url }} {% if pxe_options.repo_url %}inst.repo={{ pxe_options.repo_url }}{% else %}inst.stage2={{ pxe_options.stage2_url }}{% endif %} initrd=ramdisk || goto boot_anaconda
-initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path_https }} || goto boot_anaconda
+kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path }} text {{ pxe_options.pxe_append_params|default("", true) }} inst.ks={{ pxe_options.ks_cfg_url }} {% if pxe_options.repo_url %}inst.repo={{ pxe_options.repo_url }}{% else %}inst.stage2={{ pxe_options.stage2_url }}{% endif %} initrd=ramdisk || goto boot_anaconda
+initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path }} || goto boot_anaconda
+initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %} ../ipa-cacert-bundle || goto boot_anaconda
```
Rozzii: I think you need to put the "initrd" string inside the conditional block, otherwise you will have 2 initrd lines rendered.

scripts/tls-common.sh
```diff
@@ -129,3 +132,42 @@ if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
 else
     export BMC_TLS_ENABLED="false"
 fi
+
+if [[ -f "${WEBSERVER_CACERT_FILE:-}" ]]; then
+    copy_atomic "${WEBSERVER_CACERT_FILE}" "${IPA_CACERT_FILE}"
+elif [[ -f "${DEFAULT_CACERT_BUNDLE}" ]]; then
```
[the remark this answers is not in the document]
diconico07: Made it opt-in then, one need to set `WEBSERVER_CACERT_FILE` to enable injection into IPA as well, and set `IRONIC_IPA_INSECURE` to `0` to disable insecure mode for IPA.
This way we don't break any existing behavior and it can still be enable easily.

scripts/tls-common.sh
```diff
@@ -129,3 +132,42 @@ if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
...
+    copy_atomic "${DEFAULT_CACERT_BUNDLE}" "${IPA_CACERT_FILE}"
+fi
+
+if [[ -f "${IRONIC_CACERT_FILE}" ]]; then
+    cat "${IRONIC_CACERT_FILE}" >> "${IPA_CACERT_FILE}"
+fi
+
+if ! openssl verify -CAfile "${IPA_CACERT_FILE}" "${IRONIC_CERT_FILE}" > /dev/null 2>&1; then
+    # if we are unable to verify the Ironic cert file set IPA_INSECURE to true
```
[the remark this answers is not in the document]
diconico07: No longer required as it makes more sense to leave IPA_INSECURE flag to be managed by user as one may end up in a situation where Ironic is served behind some ingress controller or other element that may prevent this to work.

scripts/tls-common.sh
```diff
@@ -129,3 +132,42 @@ if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
 else
     export BMC_TLS_ENABLED="false"
 fi
+
+if [ -f "${WEBSERVER_CACERT_FILE:-}" ]; then
+    copy_atomic "${WEBSERVER_CACERT_FILE}" "${IPA_CACERT_FILE}"
+elif [ -f "${DEFAULT_CACERT_BUNDLE}" ]; then
+    copy_atomic "${DEFAULT_CACERT_BUNDLE}" "${IPA_CACERT_FILE}"
+fi
```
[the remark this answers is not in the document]
diconico07: Made it fully opt-in, this way we don't break any existing behavior and it can still be enabled easily, also changed the iPXE templates a bit to only generate and use the initrd when `WEBSERVER_CACERT_FILE` is set. This should cover this comment and the other ones below

diconico07:
seems to be some issue reaching opendev, trying again
/retest

ironic-config/ipxe_config.template
```diff
@@ -39,17 +49,23 @@ poweroff
 
 :boot_anaconda
```
dtantsur, line 48: nit: maybe we shouldn't pretend we support the anaconda driver...

ironic-config/ipxe_config.template
```diff
@@ -39,17 +49,23 @@ poweroff
...
 :boot_ramdisk
 imgfree
 {%- if pxe_options.boot_iso_url %}
 sanboot {{ pxe_options.boot_iso_url }}
 {%- else %}
-kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path_https }} root=/dev/ram0 text {{ pxe_options.pxe_append_params|default("", true) }} {{ pxe_options.ramdisk_opts|default('', true) }} initrd=ramdisk || goto boot_ramdisk
-initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path_https }} || goto boot_ramdisk
+kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path }} root=/dev/ram0 text {{ pxe_options.pxe_append_params|default("", true) }} {{ pxe_options.ramdisk_opts|default('', true) }} initrd=ramdisk || goto boot_ramdisk
+initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path }} || goto boot_ramdisk
+{%- if inject_cacert_bundle %}
```
dtantsur: This is not for IPA, it's for the live ISO feature. Are you sure we want to inject the bundle here?

ironic-config/ironic.conf.j2
```diff
@@ -62,6 +62,9 @@ deploy_logs_local_path = /shared/log/ironic/deploy
 # See https://bugzilla.redhat.com/show_bug.cgi?id=1822763
 max_command_attempts = 30
 certificates_path = {{ env.IRONIC_GEN_CERT_DIR }}
+{% if env.WEBSERVER_CACERT_FILE %}
+api_ca_file = {{ env.IPA_CACERT_FILE }}
```
dtantsur, line 66: The way this conditional is written exposes an implementation detail that IPA_CACERT_FILE is valid when WEBSERVER_CACERT_FILE is set. I'd rather have an explicit boolean.

Also, in other places you're checking for `WEBSERVER_CACERT_FILE is defined`

ironic-config/ironic.conf.j2
```diff
@@ -242,20 +245,18 @@ images_path = /shared/html/tmp
 instance_master_path = /shared/html/master_images
 tftp_master_path = /shared/tftpboot/master_images
 tftp_root = /shared/tftpboot
-kernel_append_params = nofb nomodeset vga=normal ipa-insecure=1 {% if env.ENABLE_FIPS_IPA %}fips={{ env.ENABLE_FIPS_IPA|trim }}{% endif %} {% if env.IRONIC_RAMDISK_SSH_KEY %}sshkey="{{ env.IRONIC_RAMDISK_SSH_KEY|trim }}"{% endif %} {{ env.IRONIC_KERNEL_PARAMS|trim }} systemd.journald.forward_to_console={{ env.IPA_FORWARD_CONSOLE | default('yes') }}
+kernel_append_params = nofb nomodeset vga=normal initrd=ipa-cacert-bundle ipa-insecure={{ env.IRONIC_IPA_INSECURE | default('1') }} {% if env.ENABLE_FIPS_IPA %}fips={{ env.ENABLE_FIPS_IPA|trim }}{% endif %} {% if env.IRONIC_RAMDISK_SSH_KEY %}sshkey="{{ env.IRONIC_RAMDISK_SSH_KEY|trim }}"{% endif %} {{ env.IRONIC_KERNEL_PARAMS|trim }} systemd.journald.forward_to_console={{ env.IPA_FORWARD_CONSOLE | default('yes') }}
```
dtantsur: Shouldn't the `initrd=ipa-cacert-bundle` part be conditional now too?

scripts/configure-ironic.sh
```diff
@@ -94,6 +94,16 @@ if [[ -f /proc/sys/crypto/fips_enabled ]]; then
     export ENABLE_FIPS_IPA
 fi
 
+if [[ "${IPXE_TLS_SETUP,,}" == "true" ]]; then
+    sed 's/set ipxe_tls_setup = false/set ipxe_tls_setup = true/' /templates/ipxe_config.template > "${IRONIC_CONF_DIR}/ipxe_config.template"
```
dtantsur: Ugh. Maybe sanitize IPXE_TLS_SETUP to always be true/false and inject it directly into the template?

scripts/configure-ironic.sh
```diff
@@ -94,6 +94,16 @@ if [[ -f /proc/sys/crypto/fips_enabled ]]; then
...
+else
+    cp /templates/ipxe_config.template "${IRONIC_CONF_DIR}/ipxe_config.template"
+fi
+
+if [[ -n "${WEBSERVER_CACERT_FILE:-}" ]]; then
+    sed -i 's/set inject_cacert_bundle = false/set inject_cacert_bundle = true/' "/${IRONIC_CONF_DIR}/ipxe_config.template"
```
dtantsur: One more argument for a separate boolean flag (see comment on ironic.conf.j2)

ironic-config/inspector.ipxe.j2
```diff
@@ -55,6 +55,9 @@ imgfree
 # ironic-inspector-image and configuration in configure-ironic.sh
 {{ kernel_cmdline('${ipa_kernel}', '${ipa_ramdisk_name}') }}
 initrd --timeout 60000 ${ipa_ramdisk} || goto retry_boot
+{%- if env.WEBSERVER_CACERT_FILE is defined %}
+initrd --timeout 60000 {{ env.IRONIC_IPA_BASE_URL }}/ipa-cacert-bundle || goto retry_boot
+{%- endif %}
```
Rozzii, line 60:
```suggestion
{%- if env.WEBSERVER_CACERT_FILE is defined %}
initrd --timeout 60000 {{ env.IRONIC_IPA_BASE_URL }}/ipa-cacert-bundle || goto retry_boot
{% else %}
initrd --timeout 60000 ${ipa_ramdisk} || goto retry_boot
{%- endif %}
```

ironic-config/inspector.ipxe.j2
```diff
@@ -63,5 +66,8 @@ imgfree
 # ironic-inspector-image and configuration in configure-ironic.sh
 {{ kernel_cmdline(env.IRONIC_IPA_BASE_URL + '/images/ironic-python-agent.kernel', 'ironic-python-agent.initramfs') }}
 initrd --timeout 60000 {{ env.IRONIC_IPA_BASE_URL }}/images/ironic-python-agent.initramfs || goto retry_boot
+{%- if env.WEBSERVER_CACERT_FILE is defined %}
+initrd --timeout 60000 {{ env.IRONIC_IPA_BASE_URL }}/ipa-cacert-bundle || goto retry_boot
+{%- endif %}
```
Rozzii, line 71:
```suggestion
{%- if env.WEBSERVER_CACERT_FILE is defined %}
initrd --timeout 60000 {{ env.IRONIC_IPA_BASE_URL }}/ipa-cacert-bundle || goto retry_boot
{% else %}
initrd --timeout 60000 {{ env.IRONIC_IPA_BASE_URL }}/images/ironic-python-agent.initramfs || goto retry_boot
{%- endif %}
```

Rozzii (commented):
Shaping up nicely but there are a few rendering issues IMO

Rozzii:
AFIK @iurygregory @hroyrh are planning to do the release please keep in mind that this PR is planned for 0.13 so let's try  to fet this in before the release

ironic-config/ipxe_config.template
diconico07: Will make this in a separate PR as it's unrelated

ironic-config/ipxe_config.template
diconico07: True we don't want that here, also moved the initrd kernel argument from ironic config to this file to ensure we don't ass it in this case as well

ironic-config/ironic.conf.j2
diconico07: I added an `IRONIC_INJECT_IPA` boolean.

ironic-config/ironic.conf.j2
diconico07: True, missed that one, moved it to the ipxe template though to ensure we only have it in cases where we actually add the bundle as initrd

scripts/configure-ironic.sh
diconico07: Noticed it's already sanitized in tls-common, so moved to direct injection

ironic-config/ipxe_config.template
diconico07: It's the intent of this, we add an additional initrd with the CA bundle and the IPA agent configuration that will get layered on top of the ironic-python-agent.initramfs one

ironic-config/inspector.ipxe.j2
dtantsur: Two initrd lines are intentional. IPA is always present, the 2nd one with the CA is added conditionally.

ironic-config/ipxe_config.template
```diff
@@ -7,21 +7,29 @@ goto deploy
...
-kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path_https }} selinux=0 troubleshoot=0 text {{ pxe_options.pxe_append_params|default("", true) }} BOOTIF=${mac} initrd={{ pxe_options.initrd_filename|default("deploy_ramdisk", true) }} || goto retry
+{%- else %}
+{%- set ari_path = pxe_options.deployment_ari_path %}
+{%- set aki_path = pxe_options.deployment_aki_path %}
+{%- endif %}
+kernel {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ aki_path }} selinux=0 troubleshoot=0 text {{ pxe_options.pxe_append_params|default("", true) }} BOOTIF=${mac} initrd={{ pxe_options.initrd_filename|default("deploy_ramdisk", true) }} {% if inject_cacert_bundle %}initrd=ipa-cacert-bundle{% endif %} || goto retry
 
-initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path_https }} || goto retry
+initrd {% if pxe_options.ipxe_timeout > 0 %}--timeout {{ pxe_options.ipxe_timeout }} {% endif %}{{ ari_path }} || goto retry
+{%- if inject_cacert_bundle %}
```
dtantsur, line 30: You no longer create these variables, just use env.IRONIC_INJECT_IPA

scripts/configure-ironic.sh
```diff
@@ -94,6 +94,11 @@ if [[ -f /proc/sys/crypto/fips_enabled ]]; then
     export ENABLE_FIPS_IPA
 fi
 
+# inject TLS and cacert bundle variables in ipxe_template
+sed "1 a\
+\{%- set ipxe_tls_setup = $IRONIC_TLS_SETUP -%}\\
+\{%- set inject_cacert_bundle = $IRONIC_INJECT_IPA -%}" /templates/ipxe_config.template > "${IRONIC_CONF_DIR}/ipxe_config.template"
```
dtantsur, line 100: You've dropped the lines you're replacing here, use the env variables directly in the template.

ironic-config/ipxe_config.template
```diff
@@ -7,21 +7,29 @@ goto deploy
 
 :deploy
 imgfree
+{%- if ipxe_tls_setup %}
```
dtantsur, line 10: `env.IRONIC_TLS_SETUP == "true"`

diconico07: I don't think `env` is injected by Ironic here

scripts/configure-ironic.sh
diconico07: I'm not replacing them here but inserting them in the file (at line 2)

ironic-config/ipxe_config.template
dtantsur: Ah, right, we have a template of templates here :)

scripts/configure-ironic.sh
dtantsur: Hmm, right. Please add an explicit placeholder (can be a follow-up)

dtantsur:
/lgtm

Thank you!

Rozzii (commented):
/approve

PR #871, opened 2026-01-14
:warning: Use CentOS minimal as base image

elfosardo:
This helps reducing the size of the final image.


elfosardo:
> Q: Does it make sense to switch the builder images to minimal as well? We'll end up installing dnf and downloading/installing more deps back and we only take out the resulting binaries anyways?

@tuminoid in the long run we may be able to remove the dnf installation bit, and I'd rather have the same base image for all the stages. The installation of dnf does not take a lot of time/space anyway.

PR #880, opened 2026-01-21
:seedling: add pip/setuptools versions to Renovate config

tuminoid:
Make renovate handle updating pip/setuptools. Require 3 days of stability before updating.


tuminoid:

PR #881, opened 2026-01-22
✨ Add multi-architecture support for IPA image selection

s3rj1k:
Add multi-architecture support for IPA image selection and improve configuration flexibility

Implements partially: https://github.com/metal3-io/ironic-standalone-operator/issues/355
Related: https://github.com/metal3-io/ironic-standalone-operator/pull/517

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,87 @@ functionality:
 - `DEPLOY_KERNEL_URL` and `DEPLOY_RAMDISK_URL` provide the default IPA kernel
   and initramfs images. If they're not set, the images from IPA downloader are
   used (if present).
+
+### Multi-Architecture IPA Support
+
+The ironic-image container supports automatic detection and boot of
+architecture-specific IPA (Ironic Python Agent) images. This allows a single
+deployment to serve different CPU architectures (x86_64, aarch64) with
+appropriate kernel and ramdisk images.
+
+- `DEPLOY_KERNEL_BY_ARCH` - Comma-separated list of architecture-specific
+  kernel URLs in the format `arch:url[,arch:url...]`. Example:
+
+  ```bash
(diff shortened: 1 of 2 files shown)


Rozzii:
> > Pleas add example configuration for this feature in the README, it is not immediately clear IMO that how this feature should be enabled, partly because the cleanup is in shell while the code that utilizes the architecture specific env variables for boot configuration is purely jinja.
> > Other question: I see you are adding the multy arch support to inspector.ipxe, but that is a generic fallback configuration for ipxe, Ironic renders node specific ipxe configuration during deployment based on an "internal template", and I don't see how your change would affect that.
> > You can se an example here if you are interested how the internal ipxe template can be customized, this is the template https://github.com/metal3-io/ironic-image/blob/main/ironic-config/ipxe_config.template and for example if custom runtime ipxe firmware building is enabled (e.g. when we run ipxe with TLS enabled) then this template is used to generate the node specific configs. If custom ipxe building is not enabled then ironic uses the templatefile contained within the ironc source code. AFAIK inspector.ipxe is used only when the node specific config fails.
> > If your experience is that you always boot based on the inspector.ipxe config that could be a sign of incorrect ironic configuration as your ironic is constantly using the fallback (I have experienced situations where users thought using inspector.ipxe is the default way to boot), ofcourse if you are using inspector.ipxe intentionally then all is good but you still need to add the multy arch support to the node specific config template.
> 
> Wait, all this change is about is to allow [metal3-io/ironic-standalone-operator#517](https://github.com/metal3-io/ironic-standalone-operator/pull/517) to work

Okay, thanks for clarifying, I see so that PR handles the part I was worried about. Okay, then please just add some documentation here about the purpose of your newly introduced env vars and I guess it is all from my side, also support the comment coming from @diconico07 .

PR #894, opened 2026-01-27
:bug: Fix microdnf segfault by removing unsupported GPG keys

elfosardo:
Remove problematic GPG keys (CentOS-SIG-Extras, centosofficial-PQC)
and explicitly import valid keys before package installation in
both Dockerfile wheel-builder stages and prepare-image.sh.
This preventsa microdnf segfault caused by attempting to use
unsupported keys.


Rozzii:
@elfosardo should we close this?

PR #898, opened 2026-02-03
:sparkles: Add multi-architecture PXE boot support

honza:
Enable aarch64 workers to be provisioned from x86_64 control plane using architecture-aware IPA image selection.

Changes:
- inspector.ipxe.j2: Use ${buildarch} to select arch-specific kernel, initramfs, and rootfs for fallback inspection path
- ipxe_config.template: Add arch-specific rootfs URL, remove port 8084 rewrite hack (BMO now provides correct URLs directly)
- ironic.conf.j2: Make ipxe_config_template unconditional (needed for multi-arch rootfs), add inspect_wait_timeout for slow ARM emulation
- runhttpd: Add inspection parameters to IRONIC_KERNEL_PARAMS so UUID-specific PXE configs have the same params as inspector.ipxe

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/ironic-config/inspector.ipxe.j2
+++ b/ironic-config/inspector.ipxe.j2
@@ -1,10 +1,16 @@
 #!ipxe
 
+# Detect CPU architecture for multi-arch support.
+# iPXE's ${buildarch} is "arm64" on aarch64, "x86_64" on x86.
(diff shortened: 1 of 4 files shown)


honza:
Oh, nice!

Issue #758, opened 2025-09-25
Test with BMO e2e


lentzi90:
The BMO e2e test suite is relatively light-weight and more comprehensive, in regards to Ironic features tested, than the CAPM3 integration tests. For example, we have recently seen that a change in the ironic-image caused the BMO live-ISO test to fail, while the CAPM3 tests didn't catch the issue. Therefore, we should implement a way to test the ironic-image using the BMO e2e suite.

I think it would be reasonable to test against the latest BMO release (not `main`), so we only have to build the ironic-image. The most straight forward way to do this is probably to just clone BMO, provide an e2e config file and kustomization specific for ironic-image, and then run the tests using the usual [ci-e2e.sh](https://github.com/metal3-io/baremetal-operator/blob/main/hack/ci-e2e.sh) script. For this we would need to make the script a bit more flexible so that we don't have to build the BMO image.

We do have more options though. For example, we could make a re-usable workflow to make it easier to run BMO e2e. This is not much different from the above but it may be nicer to use, especially if we want to use BMO e2e in more places later.

A third option would be to separate the tests from the specs (see https://github.com/metal3-io/baremetal-operator/issues/1626). Then we could implement our own version of them here, just like CAPM3 can re-use the clusterctl-upgrade tests from CAPI. For ironic-image specifically I don't think this makes much sense though.

I would suggest the first or second option to start with.

PR #901, opened 2026-02-04
✨ Add BMO e2e test integration

abhibongale:
Integrates Bare Metal Operator's e2e test suite to test ironic-image with live-ISO provisioning scenarios. Uses Method 1 from issue #758: clones BMO at release v0.12.1, builds custom ironic image, and runs tests via BMO's ci-e2e.sh script.

Fixes: #758

### DIFF
--- a/hack/prepare-bmo-tests.sh
+++ b/hack/prepare-bmo-tests.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+set -eux -o pipefail
+
+REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
+cd "${REPO_ROOT}" || exit 1
+
+CLUSTER_TYPE="${CLUSTER_TYPE:-kind}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
+IRONIC_CUSTOM_IMAGE="${IRONIC_CUSTOM_IMAGE:-localhost/ironic:bmo-e2e}"
+BMO_ROOT="${REPO_ROOT}/../baremetal-operator"
+
+# Build the ironic image
+echo "Building ironic image: ${IRONIC_CUSTOM_IMAGE}"
+"${CONTAINER_RUNTIME}" build -t "${IRONIC_CUSTOM_IMAGE}" .
+
+# Create a custom kustomize overlay in BMO that uses our image.
+# This avoids duplicating BMO's entire e2e config - we only override the ironic image.
+# TODO: Propose to BMO to accept IRONIC_IMAGE_OVERRIDE env var to simplify this further.
+# See: https://github.com/metal3-io/baremetal-operator/commit/b35cccb8 for their
+# recent refactoring that moved image overrides to kustomize overlays.
+CUSTOM_OVERLAY="${BMO_ROOT}/test/e2e/data/ironic-standalone-operator/ironic/overlays/ironic-image-custom"
+mkdir -p "${CUSTOM_OVERLAY}"
+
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
/ok-to-test

hack/prepare-bmo-tests.sh
```diff
@@ -0,0 +1,71 @@
...
+# Save and load the image into the cluster
+IMAGE_ARCHIVE="$(mktemp --suffix=.tar)"
+"${CONTAINER_RUNTIME}" save "${IRONIC_CUSTOM_IMAGE}" > "${IMAGE_ARCHIVE}"
+
+if [[ "${CLUSTER_TYPE}" == "kind" ]]; then
+    kind load image-archive -v 2 "${IMAGE_ARCHIVE}"
+else
+    minikube image load --logtostderr "${IMAGE_ARCHIVE}"
+fi
+rm -f "${IMAGE_ARCHIVE}"
```
lentzi90: We should not create the cluster here. That is handled as part of `ci-e2e.sh` in BMO. In order to load the image, it is enough to build the image and then put it in the e2e-config, like here: https://github.com/metal3-io/baremetal-operator/blob/7c6a70bd08495e7d7eba3d711aefe4c28d321f3f/test/e2e/config/ironic.yaml#L5.

hack/prepare-bmo-tests.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# Update the BMO e2e configuration to use our custom overlay
+E2E_CONFIG="${BMO_ROOT}/test/e2e/config/ironic.yaml"
+if [[ -f "${E2E_CONFIG}" ]]; then
+    # Backup original config
+    cp "${E2E_CONFIG}" "${E2E_CONFIG}.bak"
+
+    # Replace the kustomization path to use our custom overlay
+    sed -i 's|data/ironic-standalone-operator/ironic/overlays/e2e|data/ironic-standalone-operator/ironic/overlays/ironic-image-e2e|g' "${E2E_CONFIG}"
+fi
```
lentzi90: It might be better to just save a copy here in this repo with the changes we want. I am ok with this for now, but if we have to change multiple things this will be hard to understand.
We could also have the overlay saved as files here. Then copy to the right place before running the test.

test/e2e/config/ironic.yaml
```diff
@@ -0,0 +1,154 @@
...
+  listenAddress: "0.0.0.0"
+  protocol: TCP
+- containerPort: 6385
+  hostPort: 6385
+  listenAddress: "0.0.0.0"
+  protocol: TCP
+
+
+bmoIronicUpgradeSpecs:
+# Upgrade BMO 0.9 -> main
```
dtantsur: Why do we need BMO upgrade tests on ironic-image?

lentzi90 (commented):
Sorry for the delay. I am happy with this now.

One challenge now is to get green tests.
I can see that the latest run failed due to a [sushy bug](https://bugs.launchpad.net/sushy/+bug/2146416). You can see this if you download the artifacts and check `external-inspection/crd/baremetalhosts`:

```yaml
  status:
    errorCount: 1
    errorMessage: 'Failed to get power state for node 830438c6-44c4-4f3c-b399-88ed0573396e.
      Error: Invalid timeout (30, (30, 60)). Pass a (connect, read) timeout tuple,
      or a single float to set both timeouts to the same value.'
    errorType: registration error
```

There is already a fix merged, but I guess it did not propagate all the way here. Perhaps if you rebase your branch on main it would help? We can also try re-running a few times. It is not a deterministic failure so we might get lucky.

TL;DR: Good job! The failure is from an external issue and not because of any problem with this PR.

/lgtm
/retest

PR #935, opened 2026-03-06
:bug: Fix interface detection when MAC/IP matches multiple interfaces

elfosardo:
When a physical interface is enslaved to a bridge (e.g. OVN-Kubernetes), both the physical interface and the bridge share the same MAC address. The bash text-parsing pipelines in get_provisioning_interface() and get_interface_of_ip() would return multi-line values like "eno12399\nbr-ex", which is not a valid interface name and causes ironic to fail to start.

Replace both functions with a Python script (detect_interface.py) that uses `ip -json -detail` for structured output and correctly selects a single interface when multiple match. For MAC-based detection, the selection prefers the interface that already carries an IP address (important for dnsmasq binding), then non-bridge interfaces. For IP-based detection, only the first match is returned.

The script is invoked via two subcommands:
  - default (no args): MAC-based detection from PROVISIONING_MACS
  - interface-of-ip <addr> [4|6]: IP-based detection

Unit tests are added in tests/test_detect_interface.py covering all selection branches, and a GitHub Actions workflow runs them on every PR.

Assisted-By: Claude 4.6 Opus High (Commercial License)

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,0 +1,29 @@
+name: Unit Tests
+
+on:
+  pull_request:
+    types: [ opened, edited, reopened, synchronize, ready_for_review ]
+
+permissions: {}
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    strategy:
+      matrix:
+        python-version: ["3.12"]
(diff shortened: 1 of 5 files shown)


tuminoid:
This will supersede https://github.com/metal3-io/ironic-image/pull/788 as well.

PR #986, opened 2026-05-04
:seedling: remove obsoleted release action

tuminoid:
We don't need the softprops/action-gh-release as the same can be done with gh cli that is part of the runners and we have tokens present. This has been flagged by zizmor earlier.

This also causes issues for dependabot, as it cannot bump the release action due zizmor's extra comment on the same line.

This also improves the release by tagging the pre-releases properly in the draft note automatically.


lentzi90 (commented):
/lgtm

Rozzii (commented):
/approve

PR #990, opened 2026-05-07
🐛 Use placeholder in iPXE template for variable injection

### LINKED ISSUE #971: Bug in ipxe_config.template rendering
# Effect of bug
Since v35.0.0, the rendered iPXE script starts with this:

```
#!ipxeset attempts:int32 10
set i:int32 0
```

It should be

```
#!ipxe
set attempts:int32 10
set i:int32 0
```

# Visible result of bug

This iPXE script is invalid and can't be used by the iPXE runtime and is therefore discarded.  

Provisioning therefore only works with fast-track enabled, because then it doesn't power-cycle and rely on this script to boot into the deployment image. Instead, it downloads and boots the deployment image directly from IPA.

# Analysis

In [configure-ironic.sh](https://github.com/metal3-io/ironic-image/blob/0586e746bdc46f461b9e80f7ecf661168b19af8e/scripts/configure-ironic.sh#L98) there's a sed that adjusts the ipxe_config template. This adjustment results in this bug.

The [input file](https://github.com/metal3-io/ironic-image/blob/main/ironic-config/ipxe_config.template) starts like this:

```
#!ipxe

set attempts:int32 10
set i:int32 0
```

The following sed commands are run on that file:
```
sed "1 a\
\{%- set ipxe_tls_setup = $IRONIC_TLS_SETUP -%}\\
\{%- set inject_cacert_bundle = $IRONIC_INJECT_IPA -%}" /templates/ipxe_config.template > "${IRONIC_CONF_DIR}/ipxe_config.template"
```
The result looks like this:
```
#!ipxe
{%- set ipxe_tls_setup = false -%}
{%- set inject_cacert_bundle = false -%}

set attempts:int32 10$
set i:int32 0
```

Ironic then uses this template when rendering the iPXE configuration of a node.  
The `{%- ... -%}` - introduced by the sed - result in the newlines before and after the template being removed, leading to an iPXE configuration that is invalid and starts with this:

```
#!ipxeset attempts:int32 10
set i:int32 0
```

# Fix:
In [configure-ironic.sh](https://github.com/metal3-io/ironic-image/blob/0586e746bdc46f461b9e80f7ecf661168b19af8e/scripts/configure-ironic.sh#L98), in the sed, use `{% ... %}` instead of `{%- ... -%}`

diconico07:
Replace sed append operation with a placeholder-based substitution for injecting Jinja2 variables into the iPXE template. This makes the template modification more explicit and maintainable.
Also remove whitespace control to avoid issues with it being too greedy.

Fixes #971
Implements change requested in #851 (comment)

**Checklist:**

### DIFF
--- a/ironic-config/ipxe_config.template
+++ b/ironic-config/ipxe_config.template
@@ -1,4 +1,5 @@
 #!ipxe
+{# INJECT_VARIABLES #}
 
 set attempts:int32 10
 set i:int32 0
--- a/scripts/configure-ironic.sh
+++ b/scripts/configure-ironic.sh
@@ -95,9 +95,7 @@ if [[ -f /proc/sys/crypto/fips_enabled ]]; then
 fi
 
 # inject TLS and cacert bundle variables in ipxe_template
-sed "1 a\
-\{%- set ipxe_tls_setup = $IRONIC_TLS_SETUP -%}\\
-\{%- set inject_cacert_bundle = $IRONIC_INJECT_IPA -%}" /templates/ipxe_config.template > "${IRONIC_CONF_DIR}/ipxe_config.template"
+sed "s/{# INJECT_VARIABLES #}/{% set ipxe_tls_setup = $IRONIC_TLS_SETUP %}\n{% set inject_cacert_bundle = $IRONIC_INJECT_IPA %}/" /templates/ipxe_config.template > "${IRONIC_CONF_DIR}/ipxe_config.template"
 
 # The original ironic.conf is empty, and can be found in ironic.conf.orig
 render_j2_config "/etc/ironic/ironic.conf.j2" \


Rozzii:
I am not a fan of using sed to edit the jinja template, jinja should be able to pick up environment vars directly. IMO this sed based approach is hard to maintain in the long term. Let's have a discussion about some alternative.

/hold

dtantsur:
> jinja should be able to pick up environment vars directly

I thought about that too, and then realized that the template is rendered by Ironic in runtime, not by ironic-image.

Rozzii (commented):
I don't think this PR is needed anymore, also IPXE tls setup does not equal to Ironic TLS setup neither in practice nor conceptually IMO.

PR #992, opened 2026-05-08
:bug: Support using partial set of custom ipxe firmware

Rozzii:
This PR:
 - Makes the rundnsmasq and runhttpd scripts more flexible when it comes to loading custom ipxe firmware. In case of custom firmware it could be that the firmware is not compiled for all the architectures.

I feel this is a bug, I don't think ironic-image should fail to bootstrap if the arm64 ipxe is missing for a user who was/is using only custom x86_64 ipxe firmware with prior and current ironic-image versions.

### DIFF
--- a/scripts/rundnsmasq
+++ b/scripts/rundnsmasq
@@ -8,6 +8,7 @@ set -eux
 . /bin/tls-common.sh
 
 DNSMASQ_EXCEPT_INTERFACE=${DNSMASQ_EXCEPT_INTERFACE:-lo}
+TFTP_BOOT_DIR="/shared/tftpboot"
 export DNS_PORT=${DNS_PORT:-0}
 
 wait_for_interface_or_ip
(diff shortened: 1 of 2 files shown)


scripts/rundnsmasq
```diff
@@ -19,17 +20,27 @@ if [[ "${DNS_IP:-}" == "provisioning" ]]; then
...
+   for fw_file in undionly.kpxe snponly-x86_64.efi snponly-arm64.efi; do
+       if [[ -f "${IPXE_CUSTOM_FIRMWARE_DIR}/${fw_file}" ]]; then
+           cp "${IPXE_CUSTOM_FIRMWARE_DIR}/${fw_file}" "${TFTP_BOOT_DIR}"
+       else
+           echo "INFO: ${IPXE_CUSTOM_FIRMWARE_DIR}/${fw_file} is unavailable, skipping."
+       fi
+   done
+
+   # Validate that at least one iPXE binary was successfully copied
+   if ! ls "${TFTP_BOOT_DIR}/"*.{kpxe,efi} &>/dev/null; then
```
tuminoid: This will fail unless both are found, which contradicts the comment.

scripts/rundnsmasq
```diff
@@ -19,17 +20,27 @@ if [[ "${DNS_IP:-}" == "provisioning" ]]; then
...
+       echo "ERROR: No iPXE firmware files found in ${TFTP_BOOT_DIR} after copying from ${IPXE_CUSTOM_FIRMWARE_DIR}!"
+       exit 1
+   fi
```
tuminoid: Indents by 3 ?

scripts/runhttpd
```diff
@@ -65,14 +65,19 @@ if [[ "$IRONIC_VMEDIA_TLS_SETUP" == "true" ]]; then
...
-    cp "${IPXE_CUSTOM_FIRMWARE_DIR}/undionly.kpxe" \
-       "${IPXE_CUSTOM_FIRMWARE_DIR}/snponly.efi" \
-       "/shared/html/custom-ipxe"
+    for fw_file in undionly.kpxe snponly-x86_64.efi snponly-arm64.efi; do
+        if [[ -f "${IPXE_CUSTOM_FIRMWARE_DIR}/${fw_file}" ]]; then
+            cp "${IPXE_CUSTOM_FIRMWARE_DIR}/${fw_file}" "${HTTPS_IPXE_BOOT_DIR}"
+        else
+            echo "INFO: ${IPXE_CUSTOM_FIRMWARE_DIR}/${fw_file} is unavailable, skipping."
+        fi
+    done
```
tuminoid, line 80: No validation that anything is actually copied. If check makes it silently skip all if missing.

scripts/runhttpd
```diff
@@ -65,14 +65,19 @@ if [[ "$IRONIC_VMEDIA_TLS_SETUP" == "true" ]]; then
         "${HTTPD_CONF_DIR_D}/vmedia.conf"
 fi
 
+HTTPS_IPXE_BOOT_DIR="/shared/html/custom-ipxe"
```
tuminoid: Why outside the if block?

scripts/rundnsmasq
```diff
@@ -19,17 +20,27 @@ if [[ "${DNS_IP:-}" == "provisioning" ]]; then
...
+mkdir -p "${TFTP_BOOT_DIR}"
 mkdir -p /shared/html
 
 # Copy files to shared mount
 if [[ -r "${IPXE_CUSTOM_FIRMWARE_DIR}" ]]; then
-    cp "${IPXE_CUSTOM_FIRMWARE_DIR}/undionly.kpxe" \
-        "${IPXE_CUSTOM_FIRMWARE_DIR}/snponly-x86_64.efi" \
-        "${IPXE_CUSTOM_FIRMWARE_DIR}/snponly-arm64.efi" \
-        "/shared/tftpboot"
+   echo "INFO: Loading custom iPXE firmware!"
```
tuminoid: We run this with `set -x` so echo is just noise.

scripts/rundnsmasq
Rozzii: I updated the comment to clarify that 1 binary is needed for the legacy bios and one for UEFI.

scripts/rundnsmasq
Rozzii: weird, fixed

scripts/runhttpd
Rozzii: moved it

scripts/runhttpd
Rozzii: good catch forogt to add the validation here, fixed

scripts/rundnsmasq
Rozzii: removed

lentzi90 (commented):
/lgtm

tuminoid (commented):
/approve

PR #1002, opened 2026-05-15
🌱 Rebuild sushy-tools with the latest version from git

dtantsur:
I suggest we stop relying on PyPI releases. They don't bring any extra
convenience or extra testing for us, and the release process in
OpenStack is somewhat long.

This change includes a fix for BIOS settings, which makes it possible to
test HostFirmwareSettings on virtual machines.

As a bonus, it's now much easier to build local versions of sushy-tools
with an arbitrary changeset in Gerrit.


lentzi90 (commented):
/lgtm

iurygregory (approved):

PR #1003, opened 2026-05-15
[Prototype] Allow disabling iPXE when building the image

### LINKED ISSUE #823: Make efi firmware building and other ipxe related image preparation task optional.
Image

In my opinion, executing this script makes no sense.

dtantsur:
Some users indicated desire to save some build time and size by building
images that only support virtual media. This change adds a new
build-time option DISABLE_IPXE that can be set to skip building iPXE
and disable the corresponding boot interfaces.

Fixes: #823 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ ARG SETUPTOOLS_VERSION=82.0.1
 
 FROM $BASE_IMAGE AS ironic-builder
 
+ARG DISABLE_IPXE=false
 ARG IPXE_COMMIT_HASH=d0ea2b1bb8f78b219f74424d435b92ff8aa0ea8d
 ARG TARGETARCH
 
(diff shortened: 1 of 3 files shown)


Dockerfile
1nv0k32, line 1:
```suggestion
ARG BASE_IMAGE=quay.io/centos/centos:stream9-minimal
ARG BASE_IMAGE_OS=centos
```

Dockerfile
```diff
@@ -19,8 +20,12 @@ COPY prepare-ipxe.sh build-ipxe.sh prepare-efi.sh /bin/
...
 set -euxo pipefail
-prepare-ipxe.sh
-build-ipxe.sh
+if [[ "${DISABLE_IPXE,,}" != true ]]; then
+    prepare-ipxe.sh
+    build-ipxe.sh
+else
+    mkdir -p /tmp/ipxe/out
+fi
 prepare-efi.sh centos
```
1nv0k32, line 29:
```suggestion
prepare-efi.sh $BASE_IMAGE_OS
```

Dockerfile
1nv0k32: Would be good if we can also set the OS type of the base image :D

Dockerfile
1nv0k32: Would be good if we can also set the OS type of the base image :D

Dockerfile
elfosardo: this would require a huge change, it's not just a variable, everything turns around centos (scripts, configuration)

Dockerfile
elfosardo: see my comment above

MahnoorAsghar:
looks alright to me

dtantsur:
FYI: I won't proceed with this prototype for now because for my case splitting the iPXE builder into a separate image achieves much more.

Rozzii:
/close
It was decided that the solution will be to move ipxe building into a separate image.

PR #1007, opened 2026-05-19
:seedling: harden the file extraction in runlogwatch

Rozzii:
This PR:
 - Makes sure that files extracted from the Ironic logs by "runlogwatch" are not interpreted as options.

Although the content of the deployment logs are expected to have deterministic names and contain only logs sent back by a registered IPA, the hardening has very little footprint and backward compatible so there is no reason to not include it.

### DIFF
--- a/scripts/runlogwatch.sh
+++ b/scripts/runlogwatch.sh
@@ -14,7 +14,7 @@ process_log_file() {
     echo "************ Contents of ${LOG_DIR}/${FILENAME} ramdisk log file bundle **************"
     tar -tzf "${FILEPATH}" | while read -r entry; do
         echo "${FILENAME}: **** Entry: ${entry} ****"
-        tar -xOzf "${FILEPATH}" "${entry}" | sed -e "s/^/${FILENAME}: /"
+        tar -xOzf "${FILEPATH}" -- "${entry}" | sed -e "s/^/${FILENAME}: /"
         echo
     done
     rm -f "${FILEPATH}"


scripts/runlogwatch.sh
```diff
@@ -14,7 +14,7 @@ process_log_file() {
     echo "************ Contents of ${LOG_DIR}/${FILENAME} ramdisk log file bundle **************"
     tar -tzf "${FILEPATH}" | while read -r entry; do
         echo "${FILENAME}: **** Entry: ${entry} ****"
-        tar -xOzf "${FILEPATH}" "${entry}" | sed -e "s/^/${FILENAME}: /"
+        tar -xOzf "${FILEPATH}" -- "${entry}" | sed -e "s/^/${FILENAME}: /"
```
tuminoid, line 17:
```suggestion
        tar -xOzf -- "${FILEPATH}" "${entry}" | sed -e "s/^/${FILENAME}: /"
```
The filepath itself can start with -

Rozzii: other than the file name filpath is hardcoded and the file name is generated by ironic according to a ironic specific naming convention and the file names have to be actual files on the disk that are detected via watchmedo, but sure why not

Rozzii: I run a test and with the recommended change and it seems to be failing, 
`tar (child): --: Cannot open: No such file directory\n","stream":"stderr","time":"2026-05-21T06:04:38.756663489Z"}`
I would like to keep the original solution if possible, I don't see a realistic issue with the "FILEPATH" variable anyways.

tuminoid: AH ok in this case tar takes it as directly fed into -f so it doesn't already care it can start with -. Original is fine.

tuminoid (commented):
/approve

tuminoid:
Tide is stuck.
/hold

PR #1016, opened 2026-06-05
✨ Use pre-built iPXE binaries to save build time

dtantsur:


Dockerfile
```diff
@@ -1,29 +1,28 @@
 ARG BASE_IMAGE=quay.io/centos/centos:stream9-minimal
+ARG IPXE_BINARIES_IMAGE=quay.io/metal3-io/ipxe-binaries@sha256:155a410dbafc9537fe75c42f55c93843dfc6bba9e4f0676ac05f872f8a9e674d # iPXE commit d0ea2b1bb8f78b219f74424d435b92ff8aa0ea8d
```
tuminoid, line 2: Do we plan to add this to renovate or manually update as needed?

tuminoid (commented):
/approve

Dockerfile
dtantsur: It's potentially renovate-able but since updating the iPXE commit is currently a manual process, I don't see a big issue in updating this digest manually too.

lentzi90 (commented):
/lgtm

PR #1021, opened 2026-06-09
:seedling: Add strict mode to runlogwatch.sh

elfosardo:
runlogwatch.sh was the only entry-point script without any error handling. Failures in tar, sed, or rm were silently ignored. Add set -euxo pipefail consistent with the other scripts.


Rozzii (commented):
/approve
Thanks!

PR #1030, opened 2026-06-11
:bug: Suppress MARIADB_CONNECTION from trace output

elfosardo:
The connection string contains the MariaDB password and was being printed to logs via set -x. Wrap the export with set +x/set -x, consistent with the existing pattern in auth-common.sh.


tuminoid (commented):
/lgtm

PR #1031, opened 2026-06-11
:seedling: Prepare release-37.0

elfosardo:
upper-constraints from https://raw.githubusercontent.com/openstack/requirements/bc78431533f2e05fc2c68d44c245e3f21d7a4550/upper-constraints.txt


elfosardo:
related to https://github.com/metal3-io/project-infra/pull/1364

Dockerfile
tuminoid, line 24: Why does it still complain about this?

tuminoid (commented):
/lgtm

Dockerfile
elfosardo: is it still an old image before the multi arch change maybe ?

diconico07: This is the "new" ipxe-binaries image, but it already contains binaries suited for the different architecture, this one is architecture independent in the end. Is there a way to tell it to ignore that warning here ?

PR #1049, opened 2026-06-15
🌱 Disable prometheus exporter by default

UncivilizedGhost:

Disable the exporter by default so `IRONIC_EXPORTER_ENABLED` must be explicitly set to true to enable the exporter. Update documentation to explain this and explain how `FLASK_RUN_HOST` can be changed for added hardening

Fixes #

**Checklist:**

- [X] Documentation has been updated, if necessary.

### DIFF
--- a/scripts/runironic-exporter
+++ b/scripts/runironic-exporter
@@ -1,5 +1,15 @@
 #!/usr/bin/bash
 
+set -euxo pipefail
+
+# The operator sets this to "true" when enabling the exporter.
+IRONIC_EXPORTER_ENABLED=${IRONIC_EXPORTER_ENABLED:-false}
+
+if [[ "${IRONIC_EXPORTER_ENABLED}" != "true" ]]; then
+    echo " Exporter is disabled. Set EXPORTER_ENABLED=true to enable."
+    exit 0
+fi
+
 # Set dummy provisioning IP to avoid interface detection issues (not needed to run IPE to service `/metrics`)
 export PROVISIONING_IP="127.0.0.1"
 # Set to true since running this script implies sensor data metrics are needed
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
/hold
I think we agreed to do the same as in IRSO, i.e. make sure it is disabled by default and configurable by the user. Then keep the default 0.0.0.0 since 127.0.0.1 won't be useful.
Do we have a way to disable/enable the exporter?

lentzi90:
For reference, this is the IRSO PR: https://github.com/metal3-io/ironic-standalone-operator/pull/635

dtantsur (changes requested):
I agree, operators running `runironic-exporter` probably expect it to work.

scripts/runironic-exporter
```diff
@@ -1,5 +1,15 @@
 #!/usr/bin/bash
 
+set -euxo pipefail
+
+# The operator sets this to "true" when enabling the exporter.
+IRONIC_EXPORTER_ENABLED=${IRONIC_EXPORTER_ENABLED:-false}
```
dtantsur, line 6: What's the point of this? If someone runs runironic-exporter, they want it to run, right?

UncivilizedGhost: In that case how would it be possible to harden the exporter in the image repo? If it's already disabled by default in IRSO wouldn't that make this redundant?

Rozzii: @UncivilizedGhost 
I checked, this script is not run by any other script in this repo, meaning, whatever deploys Ironic, runs this script as a separate container. In other words exporter is "disabled" by default, user has to deliberately initialize a container e.g. in the ironic pod to run this.

Rozzii (changes requested):
Please close this PR .

PR #1050, opened 2026-06-16
 🐛 Fix: atomic BMC CA cert bundle write in tls-common script

migiFi:
Replaces `cat ... >` with `mktemp` + `mv`. The `mv` operation translates to a single `rename(2)` syscall which POSIX guarantees is atomic, any reader sees either the complete old bundle or the complete new one, never an empty or partial file.

The tempfile is created in the same directory as the destination to ensure source and destination are on the same filesystem, which is required for `rename(2)` atomicity.

### DIFF
--- a/scripts/tls-common.sh
+++ b/scripts/tls-common.sh
@@ -126,7 +126,11 @@ configure_restart_on_certificate_update()
 
 if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
     export BMC_TLS_ENABLED="true"
-    cat "${BMC_CACERTS_PATH}"/* > "${BMC_CACERT_FILE}"
+    BMC_CACERT_TMPFILE="$(mktemp --tmpdir="$(dirname "${BMC_CACERT_FILE}")")"
+    trap 'rm -f "${BMC_CACERT_TMPFILE}"' EXIT
+    cat "${BMC_CACERTS_PATH}"/* > "${BMC_CACERT_TMPFILE}" || { echo "ERROR: Failed to concatenate BMC certificates"; exit 1; }
+    chmod --reference="${BMC_CACERT_FILE}" "${BMC_CACERT_TMPFILE}" 2>/dev/null || { echo "WARNING: Could not preserve permissions"; }
+    mv "${BMC_CACERT_TMPFILE}" "${BMC_CACERT_FILE}" || { echo "ERROR: Failed to move certificate file"; exit 1; }
 else
     export BMC_TLS_ENABLED="false"
 fi


tests/test_tls_common.sh
dtantsur, line 1: So far, only Python based tests run in the CI. You need to convert this test to Python or update the runner.

scripts/tls-common.sh
```diff
@@ -126,7 +126,11 @@ configure_restart_on_certificate_update()
 
 if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
     export BMC_TLS_ENABLED="true"
-    cat "${BMC_CACERTS_PATH}"/* > "${BMC_CACERT_FILE}"
+    BMC_CACERT_TMPFILE="$(mktemp --tmpdir="$(dirname "${BMC_CACERT_FILE}")")"
+    trap 'rm -f "${BMC_CACERT_TMPFILE}"' EXIT
+    cat "${BMC_CACERTS_PATH}"/* > "${BMC_CACERT_TMPFILE}" \
+        && chmod --reference="${BMC_CACERT_FILE}" "${BMC_CACERT_TMPFILE}" 2>/dev/null || true \
```
dtantsur: Why `|| true` here? In what case is it expected to fail?

migiFi: if on startup `BMC_CACERT_FILE` doesn't exist yet,  `chmod --reference` has nothing to copy from and fails. The `|| true` ignores that, `mktemp's default permissions are fine in that case

tests/test_tls_common.sh
Rozzii: We had only python unit test because it was testing a python script, his change is not related to python code, it would be a bit strange/overkill imo to start writing python tests for shell scripts. I know probably we would need to transition to go or python in general as shell is getting a bit messy for this ironic-image but I would say let's jut leave out this test from this particular PR.

migiFi: I have remove the test 🐍

scripts/tls-common.sh
Rozzii: @migiFi your reasoning is correct but in practice this code is run in a container/pod so by the time this code runs, all cert files should be mounted, if they are not mounted by the time this code runs then there is a problem, please remove the || true

Rozzii (dismissed):

Rozzii (commented):
/approve

scripts/tls-common.sh
```diff
@@ -126,7 +126,11 @@ configure_restart_on_certificate_update()
 
 if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
     export BMC_TLS_ENABLED="true"
-    cat "${BMC_CACERTS_PATH}"/* > "${BMC_CACERT_FILE}"
+    BMC_CACERT_TMPFILE="$(mktemp --tmpdir="$(dirname "${BMC_CACERT_FILE}")")"
+    trap 'rm -f "${BMC_CACERT_TMPFILE}"' EXIT
+    cat "${BMC_CACERTS_PATH}"/* > "${BMC_CACERT_TMPFILE}" \
```
elfosardo: I get the point of having a long command here, although if something fails we won't know where exactly

migiFi: Thank you for pointing that out, I have added error messages for each step so failures are visible in logs. Additionally, runironic.sh  on this [PR](https://github.com/metal3-io/ironic-image/pull/1059) includes validation checks and error handling before watchmedo use the paths, providing multiple layers of error visibility.

PR #1052, opened 2026-06-17
📖 Document how the last value in the cipher suite list is choosen as the default value

UncivilizedGhost:
**What this PR does / why we need it**: Ironic checks the values of the list from last to first so the final value is the default. This behavior is  unconventional and should thus be documented.
https://github.com/openstack/ironic/blob/62dc38f151fc7ceee749794ae919757cbdd77fb7/ironic/drivers/modules/ipmitool.py#L591

Fixes #

**Checklist:** 

### DIFF
--- a/ironic-config/ironic.conf.j2
+++ b/ironic-config/ironic.conf.j2
@@ -194,6 +194,8 @@ command_retry_timeout = 60
 # List of possible cipher suites versions that can be
 # supported by the hardware in case the field `cipher_suite`
 # is not set for the node. (list value)
+# Last value in the list is the chosen as the default value
+# Ironic traverses the list last-to-first
 cipher_suite_versions = 3,17
 
 {% if env.IRONIC_EXPOSE_JSON_RPC | lower == "true" %}


dtantsur:
/lgtm
/cc @iurygregory 

Iury, do you remember the reason of this ordering?

tuminoid:
> /lgtm /cc @iurygregory
> 
> Iury, do you remember the reason of this ordering?

https://github.com/openstack/ironic/blob/62dc38f151fc7ceee749794ae919757cbdd77fb7/ironic/drivers/modules/ipmitool.py#L591

Doesn't that logic take the LAST value as preferred and move towards the first? That is some weird code, but if it applies it would mean this does the opposite it intends to do?

UncivilizedGhost:
In that case should I just close the PR since that means the original code was correct? (Or maybe add a comment explaining the last version in the list is chosen for future reference)

dtantsur:
Good catch Tuomo! I'm not sure why it's so bizarre in Ironic, but Claude also agrees:

This is a critical finding. The PR's assumption about list ordering is **inverted**.

## The Problem

The commit message says: *"Set cipher suite 17 as default and cipher suite 3 as fallback"*

But Ironic's `choose_cipher_suite()` in `ironic/drivers/modules/ipmitool.py:573-599` traverses the list **last-to-first**:

```python
if actual_cipher_suite is None:
    return available_cs_versions[-1]  # ← starts with the LAST element
```

On failure, it decrements the index backward through the list.

## What this means

| Config value             | Tried first (default) | Fallback |
| ------------------------ | --------------------- | -------- |
| `3,17` (before, on main) | **17** (last element) | 3        |
| `17,3` (after, this PR)  | **3** (last element)  | 17       |

The **original order `3,17` already does what the PR intends** — it prefers cipher suite 17 and falls back to 3. The PR reverses it, making 3 the default and 17 the fallback.

This is confirmed by the Ironic unit test `test__choose_cipher_suite_returns_last_to_first` which explicitly tests the list `['0', '1', '2', '3', '17']` and asserts it starts from `'17'` (the last element) and works backward.

**The PR should be reverted or the change dropped** — the existing `cipher_suite_versions = 3,17` is already correct for preferring cipher suite 17 with 3 as fallback.

tuminoid:
Yeah, its working as expected then. Around line 140 it also calls out explicitly that 3,17 are the only valid ones, and we have them in right order, but it is not being obvious thanks to Ironic code weirdness, so I would repivot this maybe to documenting the fact, because that is not standard way of parsing configuration and will surprise every user that touches that line for whatever purposes.

tuminoid (commented):
/lgtm

Rozzii (commented):
/lgtm
Huhh... I learned something weird today too.

tuminoid (commented):
/approve

We have 3 lgtm from approvers, so switching to get it merged.

Issue #1015, opened 2026-06-04
Fix IRSO functional for 32.0


lentzi90:
See https://github.com/metal3-io/ironic-image/pull/973#issuecomment-4562702645.

We run IRSO functional tests with latest IRSO. This does not support ironic 32.0, so validation fails.
We should make the tests use a specific version of IRSO that is compatible with the ironic version on the release branch.

dtantsur:
Fixed in https://github.com/metal3-io/ironic-image/pull/1054. Should we keep this open for releasing docs update?

Rozzii:
> Fixed in [#1054](https://github.com/metal3-io/ironic-image/pull/1054). Should we keep this open for releasing docs update?

Yes.
/triage accepted
/kind ci

tuminoid:
/assign @elfosardo 
Can you link the PR updating the docs to close this?

tuminoid:
https://github.com/metal3-io/ironic-image/pull/1069 will close this

lentzi90:
Closing as #1069 is merged.

PR #1054, opened 2026-06-17
:seedling: fix irso-functional by using IRSO 0.8 checkout

tuminoid:
We need to use IRSO version that supports this Ironic version.

Fixes: #1015


lentzi90 (commented):
Thank you!
/lgtm

PR #1055, opened 2026-06-17
:seedling: future-proof irso-functional by using IRSO 0.8 checkout

tuminoid:
This would break in future when IRSO main drops supports, so pin it.


kashifest (commented):
/lgtm

PR #1057, opened 2026-06-17
:seedling: future-proof irso-functional by using IRSO 0.9 checkout

tuminoid:
Functional would break in future if not pinned, when IRSO main drops it.


lentzi90 (commented):
/lgtm

PR #1059, opened 2026-06-17
 🐛 fix: BMC atomic cert injection/update

migiFi:
Switches the cert bundle write to a temp file + atomic move, so the live file is never seen empty or partially written during rotation.

### DIFF
--- a/scripts/runironic
+++ b/scripts/runironic
@@ -14,17 +14,50 @@ fi
 if [[ "${IRONIC_INJECT_IPA}" == "true" ]]; then
     generate_cacert_bundle_initrd /shared/html/ipa-cacert-bundle
 fi
+
 configure_restart_on_certificate_update "${IRONIC_INJECT_IPA}" ironic "${WEBSERVER_CACERT_FILE:-}"
 configure_restart_on_certificate_update "${IRONIC_TLS_SETUP}" ironic "${IRONIC_CERT_FILE}"
 
 configure_ironic_auth
 
 if [[ "${BMC_TLS_ENABLED}" == "true" ]]; then
-    # shellcheck disable=SC2034
+    # Check if BMC_CACERTS_PATH is set
+    if [[ -z "${BMC_CACERTS_PATH}" ]]; then
+        echo "ERROR: BMC_CACERTS_PATH is not set" >&2
+        exit 1
+    fi
+    
+    # Check IF any shell injection attempts in BMC_CACERTS_PATH
+    if [[ "${BMC_CACERTS_PATH}" =~ [';|&$`!'] ]] || [[ "${BMC_CACERTS_PATH}" =~ \$\( ]]; then
+        echo "ERROR: BMC_CACERTS_PATH contains invalid characters (potential shell injection): ${BMC_CACERTS_PATH}" >&2
+        exit 1
+    fi
+    
+    # Check if BMC_CACERT_FILE is set
+    if [[ -z "${BMC_CACERT_FILE}" ]]; then
+        echo "ERROR: BMC_CACERT_FILE is not set" >&2
+        exit 1
+    fi
+    
+    # Check IF any hell injection attempts in BMC_CACERT_FILE
+    if [[ "${BMC_CACERT_FILE}" =~ [';|&$`!'] ]] || [[ "${BMC_CACERT_FILE}" =~ \$\( ]]; then
+        echo "ERROR: BMC_CACERT_FILE contains invalid characters (potential shell injection): ${BMC_CACERT_FILE}" >&2
+        exit 1
+    fi
+    
+    # Verify directory exists
+    if ! test -d "${BMC_CACERTS_PATH}"; then
+        echo "ERROR: BMC_CACERTS_PATH is not a valid directory: ${BMC_CACERTS_PATH}" >&2
+        exit 1
+    fi
+    
+    # shellcheck disable=SC2016
     watchmedo shell-command \
         --patterns="*" \
         --ignore-directories \
-        --command='cat "${BMC_CACERTS_PATH}"/* > "${BMC_CACERT_FILE}"' \
+        --command='tmp=$(mktemp --tmpdir="$(dirname "${BMC_CACERT_FILE}")") && \
+cat "${BMC_CACERTS_PATH}"/* > "${tmp}" && \
+copy_atomic "${tmp}" "${BMC_CACERT_FILE}"' \
         "${BMC_CACERTS_PATH}" &
 fi
 
--- a/scripts/runironic
+++ b/scripts/runironic
@@ -21,43 +21,47 @@ configure_restart_on_certificate_update "${IRONIC_TLS_SETUP}" ironic "${IRONIC_C
 configure_ironic_auth
 
 if [[ "${BMC_TLS_ENABLED}" == "true" ]]; then
+    # Characters we reject in the BMC cert paths
+    # Defined once and matched literally as a regex class.
+    bad_chars='[;|&$`!()]'
+
     # Check if BMC_CACERTS_PATH is set
     if [[ -z "${BMC_CACERTS_PATH}" ]]; then
         echo "ERROR: BMC_CACERTS_PATH is not set" >&2
         exit 1
     fi
-    
-    # Check IF any shell injection attempts in BMC_CACERTS_PATH
-    if [[ "${BMC_CACERTS_PATH}" =~ [';|&$`!'] ]] || [[ "${BMC_CACERTS_PATH}" =~ \$\( ]]; then
+
+    # Check if any shell injection attempts in BMC_CACERTS_PATH
+    if [[ "${BMC_CACERTS_PATH}" =~ $bad_chars ]]; then
         echo "ERROR: BMC_CACERTS_PATH contains invalid characters (potential shell injection): ${BMC_CACERTS_PATH}" >&2
         exit 1
     fi
-    
+
     # Check if BMC_CACERT_FILE is set
     if [[ -z "${BMC_CACERT_FILE}" ]]; then
         echo "ERROR: BMC_CACERT_FILE is not set" >&2
         exit 1
     fi
-    
-    # Check IF any hell injection attempts in BMC_CACERT_FILE
-    if [[ "${BMC_CACERT_FILE}" =~ [';|&$`!'] ]] || [[ "${BMC_CACERT_FILE}" =~ \$\( ]]; then
+
+    # Check if any shell injection attempts in BMC_CACERT_FILE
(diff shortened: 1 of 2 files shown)


scripts/runironic
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/bash
 
 # shellcheck disable=SC1091
-. /bin/configure-ironic.sh
+
+. "${CONFIGURE_IRONIC_SH:-/bin/configure-ironic.sh}"
```
nuhakala: Why is this replaced as variable? Nothing against variables but none of the scripts here use variables, and it is little redundant because the file path for the configuration script does not need to change and hence can remain hard coded.

scripts/runironic
```diff
@@ -14,17 +15,41 @@ fi
...
+    validate_filesystem_path "BMC_CACERT_FILE" "${BMC_CACERT_FILE}"
+    
+    # shellcheck disable=SC2016
     watchmedo shell-command \
         --patterns="*" \
         --ignore-directories \
-        --command='cat "${BMC_CACERTS_PATH}"/* > "${BMC_CACERT_FILE}"' \
+        --command='tmp=$(mktemp --tmpdir="$(dirname "${BMC_CACERT_FILE}")") && \
+cat "${BMC_CACERTS_PATH}"/* > "${tmp}" && \
+mv "${tmp}" "${BMC_CACERT_FILE}"' \
```
nuhakala, line 64: There is a function `copy_atomic` in `scripts/tls-common.sh` that could perhaps make this command little shorter.

scripts/runironic
```diff
@@ -14,17 +15,41 @@ fi
...
+    if [[ -L "${var_value}" ]]; then
+        local target
+        target=$(readlink -f "${var_value}")
+        
+        if [[ -z "${target}" || "${target}" == /etc/* || "${target}" == /root/* || ! "${target}" =~ ^/[a-zA-Z0-9._/-]+$ ]]; then
+            exit 1
+        fi
+    fi
+}
+
```
nuhakala: This seems little overkill to me. The container deployer needs to mount the certificates to the container, so blacklisting certain folders might not be a good idea. Checking the filename characters seems also little redundant for the same reason: user needs to mount the certificates to the container, so why restrict the naming?

Furthermore, using `exit 1` is little hardcore, because no reason for the exit is exposed to the user and the whole execution stops without any errors. If the checks fail, it would be helpful to print something to the logs, for example `echo "Cert file is empty, continuing without TLS"` or `echo "Cert file is empty, aborting"` so that if the deployment fails, users can easily see why when inspecting the logs.

IMO, it would be enough to simply check that the files exist and are not empty for example with `test -s`.

scripts/runironic
migiFi: Something I used during testing but I have changed back now.

scripts/runironic
migiFi: Switched it over to use `copy_atomic()`. Much cleaner than doing the temp file stuff inline. Thank you for  that!

scripts/runironic
migiFi: Fair point on the original being overkill, but we do need some validation here since these vars go straight into a shell command for watchmedo, which can be questionable if someone passes in something weird.

So the update has something simpler, just blocks the shell metacharacters that actually matter ` ; | & $ ! and $( ) `. Lets normal filesystem stuff through, shows clear error messages, and doesn't gate-keep legitimate mount paths.

Basically just catching the actual invalid file and/or path.

scripts/runironic
migiFi: after some research, this wont work (unless big changes are made), `copy_atomic` is a function in `tls-common.sh `, so its not available inside the watchmedo command (runs in a separate `/bin/sh -c`), it would fail with `command not found`. Since the temp file is in the same dir as the target, `mv` is an atomic rename and does the same job, so I kept the inline `mv`.

lentzi90 (commented):
/lgtm

scripts/runironic
```diff
@@ -14,17 +14,54 @@ fi
...
 configure_ironic_auth
 
 if [[ "${BMC_TLS_ENABLED}" == "true" ]]; then
-    # shellcheck disable=SC2034
+    # Characters we reject in the BMC cert paths
+    # Defined once and matched literally as a regex class.
+    bad_chars='[;|&$`!()]'
+
+    # Check if BMC_CACERTS_PATH is set
+    if [[ -z "${BMC_CACERTS_PATH}" ]]; then
```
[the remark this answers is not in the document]
migiFi: Moved validation into `tls-common.sh` before first use, made both paths honor operator overrides (${VAR:-default}), and switched the initial write to temp file + atomic mv.  Removed the redundant block in `runironic`.

lentzi90 (commented):
/lgtm
/cc @Rozzii

Rozzii:
/hold
On first glance it is a bit too much disruption for very little benefit, I will check in more detail but I am placing hold on this to avoid accidental merge

Rozzii:
I have a few observations:

1. The startup path in tls-common.sh lost error handling: the old code checked cat/mv failures and exited with a message, cleaned up the temp file via trap, and preserved permissions with chmod --reference. The new version drops all three. A failed cat now leaves a truncated bundle in place silently, and the new file gets mktemp's 0600 with default ownership instead of the original file's mode. That's a regression, unless set -e and the container's uid make it moot — worth confirming.

2. The # shellcheck disable=SC2016 swap is correct (single-quoted $ is intentional here), but the multi-line --command with trailing \ inside a single-quoted string embeds literal backslash-newlines into the command passed to watchmedo. It works because the receiving shell treats them as line continuations, but it's fragile; a ;-separated one-liner or && without the backslashes would be clearer.

3. Other than the small technical regressions,  what is the justification for introducing user configurable BMC_CACERTS_PATH and BMC_CACERT_FILE paths? Nowadays Ironic-image has a relatively strict directory structure, "/shared, /conf, /data, /certs"  /conf and /certs are paths that should be handled correctly even with RO file systems. Is there really a good reason to allow the users to mess with these paths?  In my cost-benefit analysis so far this feature represent a lot of cost for unclear benefit.

Rozzii (commented):
Had 3 observations, I will keep this PR on hold to sort handle avoid premature merging.

migiFi:
Thanks for checking this out.  Below are some fixes I did. 

> 1. The startup path in tls-common.sh lost error handling: ...

The startup path now handles this explicitly: it writes to a temp file and on `cat` failure it removes the temp, prints an error and exits, so a failed assembly cant leave a truncated bundle, and the live file is only replaced by `mv` on success. I also set an explicit `chmod 0644` on the bundle so it no longer inherits mktemp's 0600 default ownership. `tls-common.sh` also runs under `set -euxo pipefail`, so failures about regardless.  

> 2. The # shellcheck disable=SC2016 swap is correct (single-quoted $ is intentional here), but the multi-line --command ...

Collapsed the `watchmedo --command` to a  single line with `&&`. The `# shellcheck disable=SC2016` stays since the single quoting is intentional.

> 3. Other than the small technical regressions,  what is the justification for introducing user configurable BMC_CACERTS_PATH and BMC_CACERT_FILE paths? Nowadays Ironic-image ...

You are right, cost-benefit doesnt hold up. There is no real reason to let the users move these given the fixed directory layout. I reverted `BMC_CACERTS_PATH` and `BMC_CACERTS_FILE` to hardcoded values and removed the path validation entirely, now with fixed, trusted paths there is nothing to validate. 

This now PR is now just the atomic bundle write.

scripts/tls-common.sh
```diff
@@ -145,19 +145,23 @@ configure_restart_on_certificate_update()
 
 if ls "${BMC_CACERTS_PATH}"/* > /dev/null 2>&1; then
     export BMC_TLS_ENABLED="true"
-    BMC_CACERT_TMPFILE="$(mktemp --tmpdir="$(dirname "${BMC_CACERT_FILE}")")"
-    trap 'rm -f "${BMC_CACERT_TMPFILE}"' EXIT
-    cat "${BMC_CACERTS_PATH}"/* > "${BMC_CACERT_TMPFILE}" || { echo "ERROR: Failed to concatenate BMC certificates"; exit 1; }
-    chmod --reference="${BMC_CACERT_FILE}" "${BMC_CACERT_TMPFILE}" 2>/dev/null || { echo "WARNING: Could not preserve permissions"; }
-    mv "${BMC_CACERT_TMPFILE}" "${BMC_CACERT_FILE}" || { echo "ERROR: Failed to move certificate file"; exit 1; }
+    bmc_tmp=$(mktemp "${BMC_CACERT_FILE}.XXXXXX")
```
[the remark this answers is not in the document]
Rozzii: There is no validation anymore, the operator provided custom path feature has been removed from the PR .

scripts/tls-common.sh
```diff
@@ -145,19 +145,23 @@ configure_restart_on_certificate_update()
...
+    if ! cat "${BMC_CACERTS_PATH}"/* > "${bmc_tmp}"; then
+        rm -f "${bmc_tmp}"
+        echo "Failed to assemble BMC CA bundle from ${BMC_CACERTS_PATH}" >&2
+        exit 1
+    fi
+    chmod 0644 "${bmc_tmp}"
+    mv "${bmc_tmp}" "${BMC_CACERT_FILE}"
```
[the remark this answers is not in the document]
Rozzii: AFAIK if this fails that results in a bootstrap failure of ironic anyhow so it doesn't matter whether we have left bmc_tmp or not  /tmp should be ephemeral anyhow. I don't see a realistic scenario of tempfiles piling up .

Rozzii (commented):
/lgtm

scripts/tls-common.sh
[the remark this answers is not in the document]
migiFi: Intentional, as per @Rozzii review, configurable paths and validation was dropped. These are built-in paths now, so there is nothing to honor or validate.

scripts/tls-common.sh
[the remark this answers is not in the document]
migiFi: Cleaned up now covers all failure paths, `cat`/`chmod`/`mv` are grouped and the temp file is removed on any failure. Used inline cleanup instead of an EXIT trap since the file is sourced.

lentzi90:
@Rozzii 's concerns were addressed, removing hold
/hold cancel

scripts/tls-common.sh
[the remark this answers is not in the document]
Rozzii: yeah inline is enough IMO, as I mentioned it is very unlikely that we would need to actually clean up /tmp as it is ephemeral by default, and it is generally expected on UNIX like systems that /tmp is ephemeral. (IMO if a user mounts /tmp from a persistent volume then cleanup should be their problem, so we are already doing more than what should be reasonably expected) .

PR #1060, opened 2026-06-17
⚠️ Replace IRONIC_IP with IRONIC_IPV4 and IRONIC_IPV6, for correct dual-stack and single stack binding.

mchiappero:
What this PR does / why we need it:

Currently the configuration and startup scripts rely on the assumption that all the services should either be exposed on an IPv4 or an IPv6 address. Dual-stack support is somewhat provided by leveraging IPv4-mapped IPv6 addresses, that is, only an IPv6 socket is allocated (which is a discouraged practice). Enable true dual-stack support, both in ironic.conf and all the Apache instances, by introducing IRONIC_IPV6 alongside the existing IRONIC_IP (now renamed IRONIC_IPV4), so that either or both IPs can be leveraged when defined or detected.

To avoid breaking changes and keep the PR simple, externally provided IRONIC_IP and PROVISIONING_IP values will result in either an IPv4 or an IPv6 socket allocated and not both. However, when PROVISIONING_INTERFACE is used, the new code will attempt to detect both IP versions, potentially resulting in a proper dual-stack configuration on a dual-stack system. In this case, IPv6 will however take precedence for URL generation. Future PRs will introduce the ability to handle hostname based URLs, potentially resolving to both IPv4 and IPv6 address, but the core logic to make it possible is contained in this PR.

NOTE:
- it depends on #788, commits are stacked wrt that PR
- it replaces #789, now stale and closed

### DIFF
--- a/ironic-config/apache2-ipxe.conf.j2
+++ b/ironic-config/apache2-ipxe.conf.j2
@@ -1,4 +1,5 @@
-Listen {{ env.IPXE_TLS_PORT }}
+Listen 0.0.0.0:{{ env.IPXE_TLS_PORT }}
+Listen [::]:{{ env.IPXE_TLS_PORT }}
 
 <VirtualHost *:{{ env.IPXE_TLS_PORT }}>
     ErrorLog /dev/stderr
--- a/ironic-config/dnsmasq.conf.j2
+++ b/ironic-config/dnsmasq.conf.j2
@@ -42,12 +42,12 @@ dhcp-option={{ opt }}
 dhcp-match=set:efi,option:client-arch,7
 dhcp-match=set:efi,option:client-arch,9
 # Client is (i)PXE booting on EFI machine
-dhcp-boot=tag:efi,/{{ env.SNP_BASENAME }}-x86_64.efi,{{ env.IRONIC_IP }}
+dhcp-boot=tag:efi,/{{ env.SNP_BASENAME }}-x86_64.efi,{{ env.IRONIC_IPV4 }}
 # Client is (i)PXE booting on arm64 EFI machine
 dhcp-match=set:efi-arm64,option:client-arch,11
-dhcp-boot=tag:efi-arm64,/{{ env.SNP_BASENAME }}-arm64.efi,{{ env.IRONIC_IP }}
+dhcp-boot=tag:efi-arm64,/{{ env.SNP_BASENAME }}-arm64.efi,{{ env.IRONIC_IPV4 }}
 # Client is running (i)PXE on BIOS machine
-dhcp-boot=tag:!efi,tag:!efi-arm64,/undionly.kpxe,{{ env.IRONIC_IP }}
+dhcp-boot=tag:!efi,tag:!efi-arm64,/undionly.kpxe,{{ env.IRONIC_IPV4 }}
 {%- if env.IPXE_TLS_SETUP != "true" %}
 dhcp-boot=tag:ipxe,http://{{ env.IRONIC_IPV4 }}:{{ env.HTTP_PORT }}/boot.ipxe
 {% endif %}
--- a/ironic-config/dnsmasq.conf.j2
+++ b/ironic-config/dnsmasq.conf.j2
@@ -15,17 +15,13 @@ dhcp-range={{ item }}
 {%- endfor %}
 {% endif %}
 
-{%- if env["DNS_IP"] is undefined %}
-# Disable DNS over provisioning network
-dhcp-option=6
-{% else %}
-dhcp-option=option{% if ":" in env["DNS_IP"] %}6{% endif %}:dns-server,{{ env["DNS_IP"] }}
-{% endif %}
-
 {# Network boot options for IPv4 and IPv6 #}
-{%- if env.IPV == "4" or env.IPV is undefined %}
+{%- if env.ENABLE_IPV4 %}
 # IPv4 Configuration:
 dhcp-match=ipxe,175
+
+# Either set the DNS option with DNS_IPV4 or disable the option
+dhcp-option=option:dns-server{% if env.DNS_IPV4 %},{{ env.DNS_IPV4 }}{% endif %}
 
 {# Set the router or disable it. Setting router is IPv4 specific, in v6 there #}
 {# are router advertisements that do the same thing. #}
@@ -53,14 +49,17 @@ dhcp-boot=tag:efi-arm64,/{{ env.SNP_BASENAME }}-arm64.efi,{{ env.IRONIC_IP }}
 # Client is running (i)PXE on BIOS machine
 dhcp-boot=tag:!efi,tag:!efi-arm64,/undionly.kpxe,{{ env.IRONIC_IP }}
 {%- if env.IPXE_TLS_SETUP != "true" %}
-dhcp-boot=tag:ipxe,{{ env.IRONIC_HTTP_URL }}/boot.ipxe
+dhcp-boot=tag:ipxe,http://{{ env.IRONIC_IPV4 }}:{{ env.HTTP_PORT }}/boot.ipxe
 {% endif %}
 {% endif %}
 
-{% if env.IPV == "6" %}
+{%- if env.ENABLE_IPV6 %}
 # IPv6 Configuration:
 enable-ra
 ra-param={{ env.PROVISIONING_INTERFACE }},0,0
+
+# Either set the DNS option with DNS_IPV6 or disable the option
+dhcp-option=option6:dns-server{% if env.DNS_IPV6 %},[{{ env.DNS_IPV6 }}]{% endif %}
 
 dhcp-vendorclass=set:pxe6,enterprise:343,PXEClient
 dhcp-userclass=set:ipxe6,iPXE
--- a/ironic-config/httpd-ironic-api.conf.j2
+++ b/ironic-config/httpd-ironic-api.conf.j2
@@ -17,12 +17,12 @@ Listen [::]:{{ env.IRONIC_LISTEN_PORT }}
  <VirtualHost *:{{ env.IRONIC_LISTEN_PORT }}>
 {% else %}
 {% if env.ENABLE_IPV4 %}
-Listen {{ env.IRONIC_IP }}:{{ env.IRONIC_LISTEN_PORT }}
+Listen {{ env.IRONIC_IPV4 }}:{{ env.IRONIC_LISTEN_PORT }}
 {% endif %}
 {% if env.ENABLE_IPV6 %}
 Listen [{{ env.IRONIC_IPV6 }}]:{{ env.IRONIC_LISTEN_PORT }}
 {% endif %}
-<VirtualHost {% if env.ENABLE_IPV4 %}{{ env.IRONIC_IP }}:{{ env.IRONIC_LISTEN_PORT }}{% endif %} {% if env.ENABLE_IPV6 %}[{{ env.IRONIC_IPV6 }}]:{{ env.IRONIC_LISTEN_PORT }}{% endif %}>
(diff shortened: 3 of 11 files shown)


mchiappero:
/hold

As I'm not sure if the above changes are sufficient for dnsmasq, let's hold until some feedback is received.

mchiappero:
Sorry for the lack of progress. There was indeed an error in http-ironic-api.conf.j2, which I fixed a few days ago. However the template rendering is still failing:

>  File "<unknown>", line 85, in template
> jinja2.exceptions.TemplateSyntaxError: Unexpected end of template. Jinja was looking for the following tags: 'endif'. The innermost block that needs to be closed is 'if'.
> 

However, I'm not sure what is actually causing it as the template looks now correct and I'm not able to reproduce. I'm sure @diconico07 will also look into this in the coming days.

lentzi90 (commented):
This looks promising!
I would be interested to hear @nuhakala 's thoughs, since he has been working on IPv6 related stuff before.
/cc @nuhakala 
I'll also trigger an AI review, please take the comments with a grain of salt.
/copilot-review
/label tide/merge-method-squash

ironic-config/httpd-ironic-api.conf.j2
```diff
@@ -12,11 +12,17 @@
...
 {% if env.LISTEN_ALL_INTERFACES | lower == "true" %}
-Listen {{ env.IRONIC_LISTEN_PORT }}
+Listen 0.0.0.0:{{ env.IRONIC_LISTEN_PORT }}
+Listen [::]:{{ env.IRONIC_LISTEN_PORT }}
  <VirtualHost *:{{ env.IRONIC_LISTEN_PORT }}>
 {% else %}
-Listen {{ env.IRONIC_URL_HOST }}:{{ env.IRONIC_LISTEN_PORT }}
- <VirtualHost {{ env.IRONIC_URL_HOST }}:{{ env.IRONIC_LISTEN_PORT }}>
+{% if env.ENABLE_IPV4 %}
+Listen {{ env.IRONIC_IPV4 }}:{{ env.IRONIC_LISTEN_PORT }}
```
nuhakala, line 20: Not sure if having domain names here is a real use case, but now this assumes that we are dealing with IP addresses and only IP addresses.

scripts/ironic-common.sh
```diff
@@ -167,7 +196,12 @@ wait_for_interface_or_ip()
             exit 1
         fi
 
-        export IRONIC_IP="${PARSED_IP}"
+        if [[ "${PARSED_IP}" =~ .*:.* ]]; then
+            export IRONIC_IPV6="${PARSED_IP}"
+            unset IRONIC_IP
+        else
+            export IRONIC_IPV4="${PARSED_IP}"
+        fi
```
nuhakala, line 204: Why unset `IRONIC_IP` only in v6 case?

ironic-config/apache2-ipxe.conf.j2
```diff
@@ -1,4 +1,5 @@
-Listen {{ env.IPXE_TLS_PORT }}
+Listen 0.0.0.0:{{ env.IPXE_TLS_PORT }}
+Listen [::]:{{ env.IPXE_TLS_PORT }}
```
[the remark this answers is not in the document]
diconico07: For this one, I'm unsure about the right path forward, it has a point about the enable IPvx variables, but about the `LISTEN_ALL_INTERFACES` vs binding specific addresses I'm not sure what we want to do here.

ironic-config/dnsmasq.conf.j2
```diff
@@ -46,22 +42,25 @@ dhcp-option={{ opt }}
...
 
-{% if env.IPV == "6" %}
+{%- if env.ENABLE_IPV6 %}
 # IPv6 Configuration:
 enable-ra
 ra-param={{ env.PROVISIONING_INTERFACE }},0,0
 
+# Either set the DNS option with DNS_IPV6 or disable the option
+dhcp-option=option6:dns-server{% if env.DNS_IPV6 %},[{{ env.DNS_IPV6 }}]{% endif %}
+
```
nuhakala, line 63: With IPv6 only deployment this rendered `dhcp-option=option6:dns-server` to the final config. This line broke PXE boot. I tested overriding this to `dhcp-option=6` which was the rendered config before these changes and it was working.

nuhakala (changes requested):
Few nits below. Overall I like how v4 and v6 can be enabled, leaving room for v6 only deployments also.

I tested these changes with my IPv6 only dev env setup: https://github.com/metal3-io/metal3-dev-env/pull/1709 and it did not work. PXE boot did not start due to network initialization failing. I did some investigation for the root cause, see my comment in the dnsmasq.conf for details.

ironic-config/httpd-ironic-api.conf.j2
diconico07: That's the goal of the 3rd PR on this subject (this one being the 2nd): https://github.com/metal3-io/ironic-image/pull/791

Also, the `Listen` instruction only accepts IP addresses already, so no domain names.

scripts/ironic-common.sh
diconico07: Good catch, just missed it, will move the unset out of the if/else clause

ironic-config/dnsmasq.conf.j2
diconico07: Well `dhcp-option=6` sets the DHCPv4 dns-server (not v6). 
This syntax is there to disable sending dnsmasq local address as DNS server here and let the SLAAC RA handle the DNS server setting. May it be that your v6 only deployment doesn't publish DNS through the RA (or doesn't have SLAAC) ?

The solution to get back to previous behavior might be to set `DNS_IPV6` to `::` (as IIRC it's the default value advertised by dnsmasq)

nuhakala: I am not most knowledgeable about these and have mostly debugged this with LLM. But judging from dnsmasq logs, the PXE client is receiving the dnsmasq responses and LLM suspects that PXE might consider the network broken when DNS server option is provided but empty. EDIT: PXE client fails to configure networking at all. The logs look like it does not even receive RA including SLAAC prefix, because it is not doing even SLAAC with v6.

Also, not sure how DNS is actually handled in dev env. The provisioning network should be its own network and the only router running there is ironic dnsmasq. AFAIK the PXE client cannot get DNS or DHCP information from the host's other networks and the host most likely is not sending any RA to the network. So the only RA the PXE does get, is from the dnsmasq instance. And PXE client should have ability to do SLAAC, it has configured networking with SLAAC in other tests I have done.

Maybe @Rozzii has more details of the dev env provisioning network.

elfosardo:
/hold
I'm quite concerned about this patch, mainly because we still don't have ipv6 CI jobs
I see there are already blocking comments from @nuhakala that need to be addressed
Before moving forward I think this would be a great opportunity to focus our attention to implementing ipv6 tests 
Probably worth a discussion for the next weekly meeting

ironic-config/dnsmasq.conf.j2
elfosardo: When ENABLE_IPV6 is set and DNS_IPV6 is empty, the template renders:
dhcp-option=option6:dns-server

That is not equivalent to the old dhcp-option=6 (DHCPv4 “no DNS”). Nuutti saw PXE fail to configure networking at all in an IPv6-only setup.
If DNS should be disabled, omit the option (or send a documented sentinel).
Do not emit an empty option6:dns-server.

scripts/ironic-common.sh
```diff
@@ -192,30 +226,47 @@ wait_for_interface_or_ip()
...
-        export IRONIC_IP="${PARSED_IP}"
+        if [[ "${PARSED_IP}" =~ .*:.* ]]; then
+            export IRONIC_IPV6="${PARSED_IP}"
+        else
+            export IRONIC_IPV4="${PARSED_IP}"
+        fi
     elif [[ -n "${PROVISIONING_INTERFACE}" ]]; then
         local deadline=$(( SECONDS + IRONIC_IP_WAIT_TIMEOUT ))
-        until [[ -n "$IRONIC_IP" ]]; do
+        until [[ -n "${IRONIC_IPV6}" ]] || [[ -n "${IRONIC_IPV4}" ]]; do
```
elfosardo, line 236: The stated goal of PROVISIONING_INTERFACE is to detect both addresses. As soon as one family appears, the loop stops, so a dual-stack NIC that gets IPv4 first never records IPv6 (and the reverse). There is also a sleep between the v6 and v4 lookups, so each poll costs 2 × IRONIC_IP_WAIT_INTERVAL.

ironic-config/dnsmasq.conf.j2
```diff
@@ -46,22 +42,25 @@ dhcp-option={{ opt }}
...
 # Client is (i)PXE booting on arm64 EFI machine
 dhcp-match=set:efi-arm64,option:client-arch,11
-dhcp-boot=tag:efi-arm64,/{{ env.SNP_BASENAME }}-arm64.efi,{{ env.IRONIC_IP }}
+dhcp-boot=tag:efi-arm64,/{{ env.SNP_BASENAME }}-arm64.efi,{{ env.IRONIC_IPV4 }}
 # Client is running (i)PXE on BIOS machine
-dhcp-boot=tag:!efi,tag:!efi-arm64,/undionly.kpxe,{{ env.IRONIC_IP }}
+dhcp-boot=tag:!efi,tag:!efi-arm64,/undionly.kpxe,{{ env.IRONIC_IPV4 }}
 {%- if env.IPXE_TLS_SETUP != "true" %}
-dhcp-boot=tag:ipxe,{{ env.IRONIC_HTTP_URL }}/boot.ipxe
+dhcp-boot=tag:ipxe,http://{{ env.IRONIC_IPV4 }}:{{ env.HTTP_PORT }}/boot.ipxe
```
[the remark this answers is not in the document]
elfosardo: this is correct
Previously this used IRONIC_HTTP_URL, which exists specifically when the provisioning address is not reachable from the node. IPv6 still uses IRONIC_HTTP_URL; IPv4 does not.

scripts/configure-ironic.sh
```diff
@@ -50,6 +50,14 @@ export IRONIC_IPA_COLLECTORS=${IRONIC_IPA_COLLECTORS:-default,logs}
 
 wait_for_interface_or_ip
 
+if [[ "$(echo "${LISTEN_ALL_INTERFACES}" | tr '[:upper:]' '[:lower:]')" == "true" ]]; then
+    export IRONIC_LISTEN_ADDRESS="::"
+elif [[ -n "${ENABLE_IPV6}" ]]; then
```
elfosardo, line 55: ENABLE_IPV6 is only exported when an IPv6 address was found. With set -u, an IPv4-only bind (LISTEN_ALL_INTERFACES=false) aborts here. Use "${ENABLE_IPV6:-}".

elfosardo (changes requested):

ironic-config/dnsmasq.conf.j2
diconico07: We still send the old option when it matters in IPv4 (see line 24, `dhcp-option=6` is equivalent to `dhcp-option=option:dns-server`).

And omitting the option doesn't disable the DNS, it gets dnsmasq to send its default value, that is its own IP address (I have to check though if it is actually equivalent to setting it to `::`).
I believe that Nuutti environment has no router on the provisioning network that sends RA for DNS server configuration (or IP assignment). However not disabling the DNS will have dnsmasq one to take precedence over the router provided one, and can cause issue if dnsmasq cannot be reached on DNS port or cannot resolve things that would otherwise be resolved by router given DNS server.

I agree we need a way to trigger the old behavior (IPv6 DNS advertisement is dnsmasq itself), but it's not DNS disabled.

PR #1066, opened 2026-06-19
🌱 Minor dnsmasq.conf cleanup/variable reuse

mchiappero:
Over time different commits to dnsmasq.conf have probably created inconsistencies in the use of variables, created to minimize text and maximize readability. This can also create errors when such variables are injected from the outside.

Refersh the config file template to improve their use.


Rozzii (commented):
/approve

elfosardo:
/lgtm
thanks!

PR #1076, opened 2026-06-24
:bug: Fix dnsmasq data dir path and make it configurable

elfosardo:
The dnsmasq lease file was prepared at /var/lib/dnsmasq/ during build
but referenced at /data/dnsmasq/ at runtime via dnsmasq.conf.j2. Fix
the mismatch by using the DNSMASQ_DATA_DIR variable in the template
and creating the lease file at runtime in rundnsmasq.

Make DNSMASQ_DATA_DIR configurable via a Dockerfile build arg so that
configure-nonroot.sh no longer hardcodes the path. Document the new
variable in README.md.

### DIFF
--- a/configure-nonroot.sh
+++ b/configure-nonroot.sh
@@ -15,6 +15,9 @@ set -eux
 IRONIC_USER="ironic"
 IRONIC_GROUP="ironic"
 
+# Default must match the runtime value in scripts/ironic-common.sh
+DNSMASQ_DATA_DIR="${DNSMASQ_DATA_DIR:-/data/dnsmasq}"
+
 # most containers mount /shared but dnsmasq can live without it
 mkdir -p /shared
 chown "${IRONIC_USER}":"${IRONIC_GROUP}" /shared
@@ -38,10 +41,11 @@ chmod 2775 /var/lib/ironic
 chmod 664 /var/lib/ironic/ironic.sqlite
 
 # dnsmasq, and the capabilities required to run it as non-root user
-chown -R root:"${IRONIC_GROUP}" /etc/dnsmasq.conf /var/lib/dnsmasq
-chmod 2775 /var/lib/dnsmasq
-touch /var/lib/dnsmasq/dnsmasq.leases
-chmod 664 /etc/dnsmasq.conf /var/lib/dnsmasq/dnsmasq.leases
+chown root:"${IRONIC_GROUP}" /etc/dnsmasq.conf
+chmod 664 /etc/dnsmasq.conf
+mkdir -p "${DNSMASQ_DATA_DIR}"
+chown -R root:"${IRONIC_GROUP}" "${DNSMASQ_DATA_DIR}"
+chmod 2775 "${DNSMASQ_DATA_DIR}"
 
 # probes that are created before start
 touch /bin/ironic-{readi,live}ness
--- a/Dockerfile
+++ b/Dockerfile
@@ -119,6 +119,7 @@ ARG PKGS_LIST=main-packages-list.txt
 ARG ARCH_PKGS_LIST=main-packages-list-${TARGETARCH}.txt
 ARG EXTRA_PKGS_LIST
 ARG PATCH_LIST
+ARG DNSMASQ_DATA_DIR=/data/dnsmasq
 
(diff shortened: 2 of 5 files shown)


kashifest (commented):
/lgtm

configure-nonroot.sh
```diff
@@ -38,10 +38,8 @@ chmod 2775 /var/lib/ironic
 chmod 664 /var/lib/ironic/ironic.sqlite
 
 # dnsmasq, and the capabilities required to run it as non-root user
-chown -R root:"${IRONIC_GROUP}" /etc/dnsmasq.conf /var/lib/dnsmasq
-chmod 2775 /var/lib/dnsmasq
```
dtantsur, line 42: This code initializes DNSMASQ_DATA_DIR, I don't think you should remove it

elfosardo: gosh, thanks!
totally overlooked this

kashifest: ow, it seems my previous lgtm was immature, good catch @dtantsur

configure-nonroot.sh
```diff
@@ -38,10 +38,11 @@ chmod 2775 /var/lib/ironic
...
-touch /var/lib/dnsmasq/dnsmasq.leases
-chmod 664 /etc/dnsmasq.conf /var/lib/dnsmasq/dnsmasq.leases
+chown root:"${IRONIC_GROUP}" /etc/dnsmasq.conf
+chmod 664 /etc/dnsmasq.conf
+mkdir -p /data/dnsmasq
```
diconico07: I don't like the path being hardcoded here as we are very likely to forget about this place if we ever change the DNSMASQ_DATA_DIR value (but don't really see any option here).
Also in the case of RO root filesystem, `/data` is expected to be an `emptyDir` volume, so this will have no effect.

elfosardo: in that case we could rely on the mkdir -p in ironic-common.sh line 36-39 plus fsGroup for ownership then, and remove those lines entirely

Rozzii: Please consider that there could be non root setups where we have RW file system and no /data/dnsmasq or whatever we decide to be the (dnsmasq working dir) is not a mounted volume.

I would say we should make this configurable as @diconico07 said but keep in mind this script is executed at build time thus I think the only change we need is to make the path here configurable, maybe introduce a dockerfile build arg.

Rozzii (commented):
Looks good except on aspect:
I think the use of DNSMASQ_DATA_DIR has to be aligned during runtime and build tome, in other words it has to be configurable during runtime and build time too

configure-nonroot.sh
elfosardo: I was considering that, or something similar
I see if we can integrate it here
this change is becoming bigger than expected :D

configure-nonroot.sh
```diff
@@ -15,11 +15,16 @@ set -eux
...
+CUSTOM_CONFIG_DIR="${CUSTOM_CONFIG_DIR:-/conf}"
+DNSMASQ_DATA_DIR="${CUSTOM_DATA_DIR}/dnsmasq"
+
 # most containers mount /shared but dnsmasq can live without it
 mkdir -p /shared
 chown "${IRONIC_USER}":"${IRONIC_GROUP}" /shared
-chown "${IRONIC_USER}":"${IRONIC_GROUP}" /data
-chown "${IRONIC_USER}":"${IRONIC_GROUP}" /conf
+chown "${IRONIC_USER}":"${IRONIC_GROUP}" "${CUSTOM_DATA_DIR}"
+chown "${IRONIC_USER}":"${IRONIC_GROUP}" "${CUSTOM_CONFIG_DIR}"
```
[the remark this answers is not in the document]
elfosardo: done but in the Dockerfile

elfosardo:
/hold
this is too much, I will split this in 2 and just leave this for the dnsmasq changes

Rozzii (commented):
/approve
I am ok with this, as this PR now specifically focuses on only the dnsmasq "data" dir .

elfosardo:
/override metal3-centos-e2e-integration-test-main metal3-centos-e2e-integration-test-main

PR #1079, opened 2026-06-25
:bug: Use IRONIC_ACCESS_PORT instead of hardcoded 6385 in ironic-probe.sh

elfosardo:
The probe URL for Unix socket mode had port 6385 hardcoded, ignoring any override via IRONIC_ACCESS_PORT. Use the variable, which is already sourced from ironic-common.sh and defaults to 6385.


lentzi90 (commented):
/lgtm

PR #1083, opened 2026-06-26
Allow to fallback to http OCI registries in ironic ; added secure_cdn_registries parameter

lukasmrtvy:

Support for additional OCI parameters.

**Checklist:**


dtantsur (changes requested):
Any options can be overridden by environment variables like this: `OS_OCI__PERMIT_FALLBACK_TO_HTTP_TRANSPORT` (note two underscores after the section). You don't need to add explicit variables for them.

PR #1084, opened 2026-06-26
:seedling: Reduce code duplication in entry point scripts

elfosardo:
Replace the inline python3.12 Jinja2 render in rundnsmasq with the render_j2_config function already defined in ironic-common.sh.

Move the IRONIC_FAST_TRACK default to ironic-common.sh instead of duplicating it in both runhttpd and configure-ironic.sh.


dtantsur:
/approve

I don't care about the comment, it's self-explanatory

Rozzii (commented):
/lgtm

PR #1090, opened 2026-07-03
:seedling: Use chmod 2775 instead of 0777 in runhttpd

elfosardo:
The overly permissive 0777 on /shared/html and the iPXE boot directory is unnecessary. Use 2775 (setgid + group-writable) to match the permission model used in configure-nonroot.sh.


Rozzii (commented):
/approve

PR #1091, opened 2026-07-03
:seedling: Add pipefail and missing local declarations

elfosardo:
Add pipefail to set -eux in rundnsmasq and configure-nonroot.sh to catch failures in piped commands.

Declare var_arch and file_arch as local in detect_ipa_by_arch to prevent leaking loop variables into the outer scope.


Rozzii (commented):
/approve
/cc @migiFi  FYI

PR #1095, opened 2026-07-06
🐛 Fix argument injection in SUSHY_TOOLS_CONFIG with a bash array

UncivilizedGhost:
**What this PR does / why we need it**: Replace manually built string with a bash array to prevent argument injections.

Fixes #

**Checklist:**


Rozzii (commented):
/approve

PR #1107, opened 2026-07-15
🌱 Modify image build workflow for PR

UncivilizedGhost:
**What this PR does / why we need it**: Utilize container-build-workflow from project-infra for consistency.

Fixes #

**Checklist:**


elfosardo (changes requested):
pelase don't, we're going to lose the stream10 variant with this
Also the original workflow explicitly built on both linux/amd64 and linux/arm64, does the project-infra workflow cover that?

PR #1108, opened 2026-07-15
:sparkles: Add Redfish connect_timeout to ironic.conf template

elfosardo:
When a BMC is unreachable, sushy waits up to 60 seconds per connection attempt using a single timeout for both TCP connect and HTTP read. This can cause significant delays when managing nodes with invalid or unreachable BMC addresses.

Add a connect_timeout option under [redfish] in the ironic.conf.j2 template, defaulting to 30 seconds. This separates the TCP connection timeout from the read timeout, allowing faster failure for unreachable BMCs while preserving the 60-second read timeout for slow but responsive ones.

The value is configurable via the REDFISH_CONNECT_TIMEOUT environment variable or the OS_REDFISH__CONNECT_TIMEOUT oslo.config override.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,16 @@ MariaDB configuration:
    Deprecated. Instead, mount a secret with `password` (optionally with a
    `username`) under `/auth/mariadb` mount point.
 
+Redfish BMC connection configuration:
+
+- `REDFISH_CONNECT_TIMEOUT` - TCP connection timeout in seconds for Redfish BMC
+  connections (default `30`).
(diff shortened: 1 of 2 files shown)


dtantsur:
Let's not add variables for things that can be overridden via `OS__`.

If you think that 30 seconds is a better default, let's change Ironic.

elfosardo:
> Let's not add variables for things that can be overridden via `OS__`.
> 
> If you think that 30 seconds is a better default, let's change Ironic.

right, will do....
A textbook case of me contradicting my own preaching! :D

PR #1109, opened 2026-07-16
:warning: Extract shared python-build-base for wheel builders

elfosardo:
DRY the duplicated GPG/dnf/Python toolchain setup between deps-wheel-builder and ironic-wheel-builder. Building the toolchain once improves BuildKit cache reuse and keeps git-core only on the Ironic wheel builder.


Rozzii (commented):
/lgtm

PR #1110, opened 2026-07-17
:book: Document OS_<section>__<option> Ironic config overrides

elfosardo:
Make the double-underscore env var override pattern more visible in the README with concrete examples, and note it in ironic.conf.j2 for common sections.

Assisted-By: Claude Opus 4.6 (Commercial License)

### DIFF
--- a/ironic-config/ironic.conf.j2
+++ b/ironic-config/ironic.conf.j2
@@ -1,3 +1,8 @@
+# Any option in this file can be overridden at runtime with an environment
+# variable of the form OS_<section>__<option> (two underscores after the
+# section). See README.md "Overriding Ironic configuration options" for
+# examples; dedicated image env vars are not required for each option.
+
 [DEFAULT]
 auth_strategy = noauth
 debug = true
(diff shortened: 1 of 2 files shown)


ironic-config/ironic.conf.j2
```diff
@@ -101,6 +101,9 @@ verify_step_priority_override = management.clear_job_queue:90
 node_history = False
 # Provide for a timeout longer than 60 seconds for certain vendor's hardware
 power_state_change_timeout = 120
+# Additional conductor options can be configured via OS_ environment variables,
```
diconico07: Those would be needed for every sections, not just these few. 
Also I think it would be better to have them all at either the beginning of the section or the end of the sections, as this one is in the middle of it.

elfosardo: I disagree, those are just examples, I don't want to fille the conf files with comments, the update to the README file is enough
I can maybe add a generic comment on top to point to the README file with the variables

diconico07: A generic one on top pointing to the README feels more useful to me than examples randomly distributed across the file for sure.

To leave those examples or not is more a kind of nitpicking probably, but I'd still say it kind of feels bad to me to have them randomly scattered.

Rozzii (commented):
/approve
Much more organized, thanks!

PR #1115, opened 2026-07-20
🌱 add set -euxo pipefail to entrypoint scripts for strict error handling

migiFi:
Adds `set -euxo pipefail` to the ironic-image entrypoint and sourced shell scripts so they fail fast on non-zero exit codes and pipeline failures, instead of continuing in a broken state and starting the service anyway.

### Scripts Changed
- `scripts/runironic`
- `scripts/runhttpd`
- `scripts/runironic-exporter`
- `scripts/runironic-networking`
- `scripts/rundnsmasq`
- `scripts/tls-common.sh`
- `scripts/ironic-networking-common.sh`

### Testing
- `docker build` of the image succeeds with all the changes
- Verified flags are present in the built image

### Note
`set -euxo pipefail` only catches commands and functions that exit with a non-zero code. Some functions such as `wait_for_interface_or_ip()` and `run_ironic_dbsync` in `ironic-common.sh` hang waiting on a condition instead of returning a non-zero exit code, so these flags wont act on them. 

Refactoring those functions to signal errors instead of blocking will be handled in a separate follow-up PR.


elfosardo:
ok, thanks
little bit of overlap with https://github.com/metal3-io/ironic-image/pull/1091

migiFi:
> ok, thanks little bit of overlap with #1091

rebased on main.

PR #1117, opened 2026-07-21
:seedling: Harden renovate-validator script

nuhakala:
Add check for WORKDIR variable containing a colon. A colon in the variable would break volume mount syntax.

This PR adds a check for WORKDIR variable which is used in volume mount. Theoretically the variable could contain a colon, which would cause the container runtime to consider everything after the colon as options. It can lead to undefined behavior. Forbidding colon from the directory name solves the problem and forces the volume to mount always as read only, because the script adds the read only option to the volume mount.

Fixes #

**Checklist:**

- [x] Integration tests have been added, if necessary.

### DIFF
--- a/hack/renovate-validator.sh
+++ b/hack/renovate-validator.sh
@@ -11,6 +11,14 @@ WORKDIR="${WORKDIR:-/workdir}"
 if [ "${IS_CONTAINER}" != "false" ]; then
     npx --yes -p renovate renovate-config-validator
 else
+	# Check if workdir contains :, it would break the volume mount syntax.
+	case "${WORKDIR}" in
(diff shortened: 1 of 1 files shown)


Rozzii (commented):
/approve

Rozzii:
/override metal3-ubuntu-e2e-integration-test-main metal3-centos-e2e-integration-test-main

diconico07:
I think it would be safer to just move to the mount syntax, this way we could handle `WORKDIR` and or `PWD` having a colon in them, but this may be another completely different discussion.

tuminoid (commented):
/lgtm

PR #1118, opened 2026-07-21
:warning:  Fail entrypoint waits with a timeout instead of hanging forever

migiFi:
The wait loops in ironic-common.sh:

- `wait_for_interface_or_ip`
- `run_ironic_dbsync`

looped indefinitely on their conditions, so set -euo pipefail could never act on them and a stuck dependency would hang the entrypoint silently.

Bound each loop with configurable timeout and exit non-zero when it elapses. Setting the timeout to 0 (or negative) restores the previous wait-forever behaviour.

New env vars documented in README.

**Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,12 @@ functionality:
    `PROVISIONING_MACS` is provided)
 - `PROVISIONING_IP` - the specific IP to use (instead of calculating it based on
   the `PROVISIONING_INTERFACE`)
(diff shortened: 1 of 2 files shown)


Rozzii (commented):
It could be argued that this is not a bug but a change in design, anyhow it is potentially backward incompatible so I changed this to ⚠️ from :bug:  .
Otherwise looks nice.
/approve

PR #1123, opened 2026-07-22
🌱 e2e: escape Ironic version to make it a string

dtantsur:
Otherwise, 37.0 is interpreted as a floating number.


dtantsur:

PR #1126, opened 2026-07-24
:book: Document why runhttpd does not fall back to built-in iPXE firmware

elfosardo:
Unlike rundnsmasq, the TLS iPXE setup in runhttpd intentionally
requires custom firmware via IPXE_CUSTOM_FIRMWARE_DIR. TLS boot
needs certificates and an HTTPS boot script baked into the
firmware; falling back to the generic /tftpboot binaries would
silently serve non-TLS-capable firmware.

### DIFF
--- a/scripts/runhttpd
+++ b/scripts/runhttpd
@@ -68,6 +68,9 @@ if [[ "$IPXE_TLS_SETUP" == "true" ]]; then
     mkdir -p "${HTTPS_IPXE_BOOT_DIR}"
     chmod 2775 "${HTTPS_IPXE_BOOT_DIR}"
     render_j2_config "/templates/httpd-ipxe.conf.j2" "${HTTPD_CONF_DIR_D}/ipxe.conf"
+    # No fallback to /tftpboot here: TLS iPXE requires custom firmware with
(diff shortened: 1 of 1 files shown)


Rozzii (commented):
I believe this is intentional, even if the pre-built IPXE supports TLS protocol, one would need to bake in the certs into the iPXE firmware. I don't think it is a good idea to fall back to the default firmware if the user explicitly selected IPXE_TLS_SETUP. Also our TLS enabled IPXE relies on an embedded script that directs the firmware to download the boot.ipxe from a specific https endpoint.

/hold

elfosardo:
> I believe this is intentional, even if the pre-built IPXE supports TLS protocol, one would need to bake in the certs into the iPXE firmware. I don't think it is a good idea to fall back to the default firmware if the user explicitly selected IPXE_TLS_SETUP. Also our TLS enabled IPXE relies on an embedded script that directs the firmware to download the boot.ipxe from a specific https endpoint.

@Rozzii this should probable be documented inline

Rozzii (commented):
/approve

Rozzii:
Only a comment was added, no need to run integration tests.

PR #1129, opened 2026-07-28
:bug: Fix minor startup bugs in ironic-common.sh and httpd template

elfosardo:
- Fix error message in wait_for_interface_or_ip referencing PROVISIONING_IP instead of IRONIC_IP in the IRONIC_IP branch.
- Fix iPXE firmware validation using brace expansion that false-fails when only one firmware type (.kpxe or .efi) is present; test each glob independently so a valid partial set is accepted.
- Guard IRONIC_HTTPD_LOGLEVEL against exported empty values that would render a bare LogLevel directive and prevent Apache from starting.


elfosardo:
cc @Rozzii @diconico07

PR #1136, opened 2026-07-31
🌱 Add spec.images instead of patching spec.images/ironic

### LINKED ISSUE #1135: BMO tests broken due to changes kustomizations
The BMO tests are broken here since https://github.com/metal3-io/baremetal-operator/pull/3458, because we here try to apply a patch to the image overrides that were removed. See https://github.com/metal3-io/ironic-image/blob/1f9a4ce34ed7f33bcade5b3fb8c0a37e9e44517d/hack/prepare-bmo-tests.sh#L38
I think only the `op` need to be changed from `replace` to `add`.

UncivilizedGhost:
**What this PR does / why we need it**: BMO removed spec.images from its e2e Ironic overlay, so the old path no longer exists.

Fixes https://github.com/metal3-io/ironic-image/issues/1135

**Checklist:**

### DIFF
--- a/hack/prepare-bmo-tests.sh
+++ b/hack/prepare-bmo-tests.sh
@@ -33,9 +33,10 @@ patches:
 - target:
     kind: Ironic
   patch: |-
-    - op: replace
-      path: /spec/images/ironic
-      value: '${IRONIC_CUSTOM_IMAGE}'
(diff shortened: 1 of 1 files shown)


Rozzii (commented):
/lgtm

PR #1143, opened 2026-08-03
✨ Add no-inspect and idrac-redfish to inspect interfaces

dtantsur:
The no-inspect interface makes it clear that the node does not support inspection.
It will be eventually used for hosts that have inspection disabled to
avoid the hidden dependency on agent inspection.

The idrac-redfish interface is an enhanced version of redfish for Dell machines.

Required for: metal3-io/baremetal-operator#3283


dtantsur:
/test metal3-centos-e2e-integration-test-main

Artifactory timeout

PR #1146, opened 2026-08-03
:seedling: Work around re-usable workflow limitation

Rozzii:
This commit:
  - Adds explicit "--build-arg" option for every build argument after the first in the merge build workflow.

Reason:
 - The workflow inherited from metal3-io/project-infra "re-defines" the build-args parameter of the original mr-smithers-excellent/docker-build-push from list to string. Because of the "type conversion" the original action only appends one "--build-args" not one per key=value pair as it supposed to.
 - Given the situation, this seems to be the fix with the lowest footprint.

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -28,7 +28,7 @@ jobs:
       pushImage: true
       image-build-args: >-
         IRONIC_VERSION_LABEL=${{ needs.generate-core-component-source-versions.outputs.ironic-version }}
-        NGS_VERSION_LABEL=${{ needs.generate-core-component-source-versions.outputs.ngs-version }}
+        --build-arg NGS_VERSION_LABEL=${{ needs.generate-core-component-source-versions.outputs.ngs-version }}
       generate-sbom: true
       sign-image: true
     secrets:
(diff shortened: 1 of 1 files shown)


Rozzii:

Rozzii:

.github/workflows/build-images-action.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       pushImage: true
       image-build-args: >-
         IRONIC_VERSION_LABEL=${{ needs.generate-core-component-source-versions.outputs.ironic-version }}
-        NGS_VERSION_LABEL=${{ needs.generate-core-component-source-versions.outputs.ngs-version }}
+        --build-arg NGS_VERSION_LABEL=${{ needs.generate-core-component-source-versions.outputs.ngs-version }}
         BASE_IMAGE=quay.io/centos/centos:stream10-minimal
```
lentzi90: Wouldn't this also need `--build-arg`?

lentzi90 (commented):
Looks like list type inputs is not a thing. 🙁 
I found some workarounds in this discussion: https://github.com/orgs/community/discussions/11692
But for now I think we can try this first.

(I also got a suggestion from copilot that we could solve it by changing `image-build-args: >-` to `image-build-args: |-`. I am skeptical it would work. I know we already tried with `|`, the only difference here is that the `-` removes the newline at the end.)

.github/workflows/build-images-action.yml
Rozzii: yes indeed, thanks

Rozzii:

lentzi90 (commented):
/approve

lentzi90 (commented):
Err... I mean
/lgtm

PR #1151, opened 2026-08-04
[DNM] Test newer version of ironic

elfosardo:


elfosardo:
new code seems to break BMO e2e tests
a fix is being worked here https://review.opendev.org/c/openstack/ironic/+/999699

PR #1153, opened 2026-08-06
🐛 Bracket IPv6 IRONIC_EXTERNAL_IP when composing external URLs

### LINKED ISSUE #1058: IRONIC_EXTERNAL_IP does not support IPv6
### Summary

external_http_url/external_callback_url are built from IRONIC_EXTERNAL_IP without bracketing IPv6, producing invalid URLs

### What steps did you take and what happened?

  When `IRONIC_EXTERNAL_IP` is set to an IPv6 address, `scripts/configure-ironic.sh`
  composes the external URLs by concatenating `<ip>:<port>` directly:

  https://github.com/metal3-io/ironic-image/blob/main/scripts/configure-ironic.sh#L59-L64

  ```sh
  if [[ -n "$IRONIC_EXTERNAL_IP" ]]; then
      export IRONIC_EXTERNAL_CALLBACK_URL=${IRONIC_EXTERNAL_CALLBACK_URL:-"${IRONIC_SCHEME}://${IRONIC_EXTERNAL_IP}:${IRONIC_ACCESS_PORT}"}
      if [[ "$IRONIC_VMEDIA_TLS_SETUP" == "true" ]]; then
          export IRONIC_EXTERNAL_HTTP_URL=${IRONIC_EXTERNAL_HTTP_URL:-"https://${IRONIC_EXTERNAL_IP}:${VMEDIA_TLS_PORT}"}
      else
          export IRONIC_EXTERNAL_HTTP_URL=${IRONIC_EXTERNAL_HTTP_URL:-"http://${IRONIC_EXTERNAL_IP}:${HTTP_PORT}"}
      fi
  fi
 ```

  For an IPv6 address (e.g. 2001:db8::139) this yields
  http://2001:db8::139:6180, where the host and port are ambiguous. Ironic then
  fails to start while validating external_callback_url:

   ```CRITICAL ironic Unhandled error: oslo_config.cfg.ConfigFileValueError: Value for
  option external_callback_url ... is not valid: ('host was required but missing',
  URIReference(scheme='http', authority='2001:db8::139:6385', ...), ['host'])
  ```
  The IPv6 literal needs to be bracketed: http://[2001:db8::139]:6385.

 ###  Environment

  - ironic-image: main (the lines above are unchanged on main; also reproduces
  on the image used by Ironic version 32.0)
  - Driven via ironic-standalone-operator (networking.externalIP), but the
  operator only forwards the bare IP as IRONIC_EXTERNAL_IP — the operator's
  webhook intentionally rejects a bracketed value, so the bracketing has to
  happen here when the URL is composed.
  - Use case: Redfish/iDRAC virtual-media boot where the BMC reaches the image
  server over an IPv6 management network.

  ###  What did you expect to happen?

  external_http_url and external_callback_url should be valid when
  IRONIC_EXTERNAL_IP is IPv6, i.e. the address bracketed before appending :port.

  ###  Anything else you would like to add?

  The bug only affects the auto-composed defaults; passing
  IRONIC_EXTERNAL_HTTP_URL/IRONIC_EXTERNAL_CALLBACK_URL explicitly (already
  bracketed) works, which is the current workaround.

  Suggested fix: bracket the host when it parses as an IPv6 literal before building
  the URLs (a small [[ "$IRONIC_EXTERNAL_IP" == *:* ]] && host="[$IRONIC_EXTERNAL_IP]"
  guard, or a shared helper). 

I'd be happy to open a PR if this approach sounds like the right way to solve this issue

mehrdadbn9:
## What this fixes

`IRONIC_EXTERNAL_IP` was concatenated straight onto the port when composing the
external URLs, so an IPv6 address produced `http://2001:db8::139:6385`. That is
ambiguous, and Ironic refuses to start while validating it:

```
CRITICAL ironic Unhandled error: oslo_config.cfg.ConfigFileValueError: Value for
option external_callback_url ... is not valid: ('host was required but missing',
URIReference(scheme='http', authority='2001:db8::139:6385', ...), ['host'])
```

The address needs bracketing first. `ironic-common.sh` already does exactly this
when it builds `IRONIC_URL_HOST`, so this follows the same pattern rather than
introducing a new one.

## Behaviour

| `IRONIC_EXTERNAL_IP` | callback URL | http URL (TLS off) |
|---|---|---|
| `192.168.0.10` | `http://192.168.0.10:6385` | `http://192.168.0.10:6180` |
| `2001:db8::139` | `http://[2001:db8::139]:6385` | `http://[2001:db8::139]:6180` |
| `[2001:db8::139]` | `http://[2001:db8::139]:6385` | `http://[2001:db8::139]:6180` |

IPv4 output is unchanged, and an already bracketed value is passed through so it
is not wrapped twice. Only the auto-composed defaults are affected — setting
`IRONIC_EXTERNAL_CALLBACK_URL` or `IRONIC_EXTERNAL_HTTP_URL` explicitly still
overrides them, which is the workaround people use today.

Bracketing has to happen here rather than upstream: IrSO forwards only the bare
IP as `IRONIC_EXTERNAL_IP`, and its webhook rejects a bracketed value.

## Testing

- `./hack/shellcheck.sh` clean.
- Replayed the block against IPv4, IPv6, already-bracketed and link-local input
  and confirmed the composed URLs match the table above.
- Confirmed the old form is what breaks parsing and the new one is not:
  `urlparse("http://2001:db8::139:6385")` raises
  `Port could not be cast to integer value as 'db8::1'`, while
  `urlparse("http://[2001:db8::139]:6385")` gives host `2001:db8::139`, port `6385`.

Fixes #1058

### DIFF
--- a/scripts/configure-ironic.sh
+++ b/scripts/configure-ironic.sh
@@ -54,11 +54,23 @@ wait_for_interface_or_ip
 export IRONIC_CONDUCTOR_HOST=${IRONIC_CONDUCTOR_HOST:-${IRONIC_URL_HOST}}
 
 if [[ -n "$IRONIC_EXTERNAL_IP" ]]; then
-    export IRONIC_EXTERNAL_CALLBACK_URL=${IRONIC_EXTERNAL_CALLBACK_URL:-"${IRONIC_SCHEME}://${IRONIC_EXTERNAL_IP}:${IRONIC_ACCESS_PORT}"}
-    if [[ "$IRONIC_VMEDIA_TLS_SETUP" == "true" ]]; then
-        export IRONIC_EXTERNAL_HTTP_URL=${IRONIC_EXTERNAL_HTTP_URL:-"https://${IRONIC_EXTERNAL_IP}:${VMEDIA_TLS_PORT}"}
+    # An IPv6 literal has to be bracketed before a port is appended, the same
+    # way IRONIC_URL_HOST is built in ironic-common.sh. Without it the URL is
+    # composed as http://2001:db8::139:6385, where host and port are
+    # ambiguous, and Ironic refuses to start while validating
+    # external_callback_url with "host was required but missing".
+    # An already bracketed value is left alone so it is not double wrapped.
+    if [[ "$IRONIC_EXTERNAL_IP" =~ .*:.* ]] && [[ "$IRONIC_EXTERNAL_IP" != \[*\] ]]; then
+        IRONIC_EXTERNAL_URL_HOST="[$IRONIC_EXTERNAL_IP]"
     else
-        export IRONIC_EXTERNAL_HTTP_URL=${IRONIC_EXTERNAL_HTTP_URL:-"http://${IRONIC_EXTERNAL_IP}:${HTTP_PORT}"}
+        IRONIC_EXTERNAL_URL_HOST="$IRONIC_EXTERNAL_IP"
+    fi
+
+    export IRONIC_EXTERNAL_CALLBACK_URL=${IRONIC_EXTERNAL_CALLBACK_URL:-"${IRONIC_SCHEME}://${IRONIC_EXTERNAL_URL_HOST}:${IRONIC_ACCESS_PORT}"}
+    if [[ "$IRONIC_VMEDIA_TLS_SETUP" == "true" ]]; then
+        export IRONIC_EXTERNAL_HTTP_URL=${IRONIC_EXTERNAL_HTTP_URL:-"https://${IRONIC_EXTERNAL_URL_HOST}:${VMEDIA_TLS_PORT}"}
+    else
+        export IRONIC_EXTERNAL_HTTP_URL=${IRONIC_EXTERNAL_HTTP_URL:-"http://${IRONIC_EXTERNAL_URL_HOST}:${HTTP_PORT}"}
     fi
 fi
 


mehrdadbn9:
The `test` (BMO E2E) job failed, but this change is a 15-line shell-script IPv6 URL-bracketing fix in `ironic-common.sh` — it cannot affect a Bare Metal Operator end-to-end integration test (kind cluster + BMO). Note the `metal3-centos-e2e-integration-test-main` (Jenkins) passed and an earlier `test` run also passed, which points to an environmental flake in the ubuntu E2E run rather than this diff.

Could a maintainer please re-run the failed `test` job? Happy to provide more info if it fails again after a re-run.

mehrdadbn9:
## CI note: the `test` (e2e) failure is an infra flake, not a regression from this change.

The failing test is `Inspection should inspect a BMH using fast (out-of-band) inspection`, which **timed out after 600s** waiting for BMH registration. The surrounding log shows the e2e environment never came up:
- `dial tcp 127.0.0.1:35317: connect: connection refused` (the in-test Kubernetes API proxy)
- `namespaces "automated-cleaning" / "inspection" / ... not found`

That's the classic "e2e/Docker-in-Docker stack didn't initialize" failure. Result: **30 Passed | 1 Failed | 9 Skipped** — only one e2e test failed, and on an environment timeout, not an assertion.

This PR only brackets an IPv6 `IRONIC_EXTERNAL_IP` when composing external URLs — a pure string-formatting change. `unit-tests (3.12)` PASSED, and a BMH-registration timeout cannot be caused by it.

Could a maintainer please re-run the failed `test` job? Thanks!

PR #1202, opened 2026-09-04
:seedling: Pin default upper-constraint

Rozzii:
This PR:
 - Pins the upper-constraint default version
 - Introduces new Dockerfile build argument to facilitate upper-constraint pinning
 - Introduces a new code path in build-wheels.sh that downloads a pinned default version of the OpenStack Requirements upper-constraint file if there is no upper-constraint file provided in the root of the repo.
 - Introduces new renovate config to handle NGS,Ironic and OpenStack Requirements version pinning/update together.
 - Adds documentation explaining how does the upper constraint pinning and related CI automation works.

This change was needed because over the years there have been occasions where the whole Metal3 CI broke because the latest ironic-image was built with misaligned python dependencies.

- [x] Integration tests have been added, if necessary.


Rozzii:
/cc @elfosardo @terror96 @dtantsur @nuhakala @Sunnatillo 

I am ccing people who IMO should be in the loop either because of their ironic/ironic-image affiliations or because they work on the CI a lot.

tuminoid (commented):
Renovate config works, there is update even available so we get to test it out after this merges.

This needs to be linked/instructed in the release documentation, otherwise the requirements keep bumping for main and not the release supposedly pinned. There is already instructions how to do this pinning when cutting a branch.

terror96 (commented):
gopilot has some nit, but /lgtm

build-wheels.sh
```diff
@@ -11,7 +11,11 @@ UPPER_CONSTRAINTS_PATH="/tmp/${UPPER_CONSTRAINTS_FILE:-}"
 # we assume we're on the master branch
 if [[ ! -s "${UPPER_CONSTRAINTS_PATH}" ]]; then
     UPPER_CONSTRAINTS_PATH="/tmp/upper-constraints.txt"
-    curl -L https://releases.openstack.org/constraints/upper/master -o "${UPPER_CONSTRAINTS_PATH}"
+    if [[ -n "${OPENSTACK_REQUIREMENTS_SOURCE:-}" ]]; then
+        curl -fL "https://raw.githubusercontent.com/openstack/requirements/${OPENSTACK_REQUIREMENTS_SOURCE}/upper-constraints.txt" -o "${UPPER_CONSTRAINTS_PATH}"
```
tuminoid, line 15: We need to bring the `-f` to main too.

tuminoid (commented):
/lgtm

PR #1207, opened 2026-09-08
✨ Enable IPA pci-devices collector and accelerators inspection hook

fmuyassarov:
Enable the pci-devices collector on the IPA ramdisk and the accelerators processing hook on the inspector side so that discoveredc PCI accelerator devices (example GPUs) are populated into node.properties["accelerators"] during inspection.

This is required for BMO to expose accelerator inventory in hardwareData & BMH. xref: metal3-io/baremetal-operator#3601

**Checklist:**


Rozzii (commented):
/approve

PR #1219, opened 2026-09-18
🌱 Use release branch for BMO E2E

lentzi90:

Instead of a fixed tag, we use the release branch so we get fixes when needed.

Background: artifactory.nordix.org is down. We have changed BMO E2E to use other sources, but those changes are not propagating here since we check out a specific tag.

Fixes #

**Checklist:**


lentzi90:
/hold
I need to fix the kustomization

kashifest:
Although I understand the issue, shouldn't we do a new BMO patch release asap rather? Using release branch tip means we will use moving head and although there shouldn't be breaking/API changes now but theres a risk of drift coming from BMO code still here since we are not pinning a release tag. Do we want that? 

> **What this PR does / why we need it**:
> 
> Instead of a fixed tag, we use the release branch so we get fixes when needed.
> 
> Background: artifactory.nordix.org is down. We have changed BMO E2E to use other sources, but those changes are not propagating here since we check out a specific tag.

lentzi90:
That is a fair point. We can go with specific pinned versions instead. But, as you noted, we will need a new release first then.

Rozzii (commented):
Isn't 0.12.x a bit old for v37 Ironic? I mean it works but v37 came out much later.

lentzi90:
> Isn't 0.12.x a bit old for v37 Ironic? I mean it works but v37 came out much later.

The PR that added these tests came in February, and then we did not have v0.13 yet, so in that sense it is the latest we had.

lentzi90:
Discussed offline. We realized BMO v0.12 is no longer supported so we won't make a new tag for it. We should try to move to a supported release then. But let's take this first to unblock CI.

kashifest (commented):
I am changing my stance here since we wont have v0.12.x release for BMO anymore. Perhaps we missed a BMO bump? 
/lgtm

Rozzii (commented):
/lgtm

